PR #9, opened 2016-04-29
Enable gcc 4.4

lmussier:
Hi,

Here is part of the way to be able to use the SDK with gcc 4.4.
It adds a new macro to avoid multiple typedef with the same base type.

See https://github.com/Azure/azure-iot-sdks/issues/466

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,6 @@
 /build_all/windows/nuget.exe
 /cmake
 /devdoc/~$o_requirements.docm
+#### Manully added section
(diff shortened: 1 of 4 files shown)


olivierbloch:
We have already made a set of changes to support gcc 4.4 and are working at adding this in our gated build. Once it's gone will update this PR.

tameraw:
@lmussier - We have enabled 4.4 build in our gated system. Please reopen if you see more issues around 4.4 support.
Thanks.

PR #15, opened 2016-08-10
Support building on iOS

msjaosborn:
azure-c-shared-utility won't build on iOS due to a syntax error (missing ;) in condition_pthreads.c, as well as some issues with pthread_condattr_getclock & pthread_condattr_setclock.  They're not available on iOS.  Also, TCP_KEEPIDLE is not defined on iOS.

Using mach_gettime.c/.h from gist (https://gist.github.com/alfwatt/3588c5aa1f7a1ef7a3bb).

Not sure why condtion_pthread.c shows extensive changes below.  Suspect it's just a line break (CRLF) issue.

### DIFF
--- a/adapters/condition_pthreads.c
+++ b/adapters/condition_pthreads.c
@@ -10,6 +10,7 @@
 #include "azure_c_shared_utility/gballoc.h"
 #include "pthread.h"
 #include "time.h"
+#include "mach_gettime.h"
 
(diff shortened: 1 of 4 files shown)


dcristoloveanu:
2 requests:
Can you please get this PR done against the develop branch?
Also it would be very good to have a script to build this in the gated system we have. If you would create a build script for iOs in the Jenkins folder we can plug it in the gated build so we make sure it does not break.

Cheers,
/Dan

PR #16, opened 2016-08-11
Support building on iOS

msjaosborn:
azure-c-shared-utility won't build on iOS due to a syntax error (missing ;) in condition_pthreads.c, as well as some issues with pthread_condattr_getclock & pthread_condattr_setclock. They're not available on iOS. Also, TCP_KEEPIDLE is not defined on iOS.

Using mach_gettime.c/.h from gist (https://gist.github.com/alfwatt/3588c5aa1f7a1ef7a3bb).

Not sure why condtion_pthread.c shows extensive changes below. Suspect it's just a line break (CRLF) issue.

### DIFF
--- a/adapters/condition_pthreads.c
+++ b/adapters/condition_pthreads.c
@@ -10,6 +10,7 @@
 #include "azure_c_shared_utility/gballoc.h"
 #include "pthread.h"
 #include "time.h"
+#include "mach_gettime.h"
 
 DEFINE_ENUM_STRINGS(COND_RESULT, COND_RESULT_VALUES);
 
@@ -22,19 +23,23 @@ clockid_t time_basis = CLOCK_MONOTONIC;
 #ifdef CLOCK_REALTIME
 clockid_t time_basis = CLOCK_REALTIME;
 #else
-clockid_t time_basis = -1
+clockid_t time_basis = -1;
 #endif
 #endif
 
 COND_HANDLE Condition_Init(void)
 {
+#if TARGET_OS_IPHONE
+    //pthread_condattr_getclock and pthread_condattr_setclock not available on iOS 9.3
+    return NULL;
+#else
     // If we don't know our time basis, find it.
     if (time_basis == -1)
     {
         pthread_condattr_t cattr;
         pthread_condattr_init(&cattr);
         pthread_condattr_getclock(&cattr, &time_basis);
-        pthread_condattr_destroy(&cattr);
+        pthread_condattr_destroy(&cattr);
     }
 
     // Codes_SRS_CONDITION_18_002: [ Condition_Init shall create and return a CONDITION_HANDLE ]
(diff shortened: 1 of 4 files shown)


adapters/socketio_berkeley.c
```diff
@@ -638,8 +638,12 @@ int socketio_setoption(CONCRETE_IO_HANDLE socket_io, const char* optionName, con
         }
         else if (strcmp(optionName, "tcp_keepalive_time") == 0)
         {
+#if TARGET_OS_IPHONE
```
dcristoloveanu: >TARGET_OS_IPHONE [](start = 4, length = 16)

How does this get set? I'd expect this to be set through cmake.

adapters/mach_gettime.h
```diff
@@ -0,0 +1,26 @@
+#include <sys/types.h>
```
dcristoloveanu: Missing copyright header

adapters/mach_gettime.c
```diff
@@ -0,0 +1,43 @@
+#include "mach_gettime.h"
```
dcristoloveanu: Missing copyright header

adapters/socketio_berkeley.c
```diff
@@ -638,8 +638,12 @@ int socketio_setoption(CONCRETE_IO_HANDLE socket_io, const char* optionName, con
...
+            result = -1;
+#else
             result = setsockopt(socket_io_instance->socket, SOL_TCP, TCP_KEEPIDLE, value, sizeof(int));
```
dcristoloveanu: >SOL_TCP [](start = 60, length = 7)

Apparently this should be IPPROTO_TCP for the level. for both calls (here and in line 650).

dcristoloveanu: >TCP_KEEPIDLE [](start = 69, length = 12)

If TCP_KEEPIDLE is not used on iOs then so should be TCP_KEEPINTVL (and then SO_KEEPALIVE does not mean much either).

adapters/mach_gettime.c
```diff
@@ -0,0 +1,43 @@
...
+#include <mach/mach_time.h>
+
+#define MT_NANO (+1.0E-9)
+#define MT_GIGA UINT64_C(1000000000)
+
+// TODO create a list of timers,
+static double mt_timebase = 0.0;
+static uint64_t mt_timestart = 0;
```
dcristoloveanu: >uint64_t [](start = 7, length = 8)

This requires #include <stdint.h>

adapters/condition_pthreads.c
```diff
@@ -1,150 +1,157 @@
...
+clockid_t time_basis = -1;
+#endif
+#endif
+
+COND_HANDLE Condition_Init(void)
+{
+#if TARGET_OS_IPHONE
+    //pthread_condattr_getclock and pthread_condattr_setclock not available on iOS 9.3
+    return NULL;
+#else
```
dcristoloveanu: The condition code should be turned off by default.
In CMakeLists:

option(use_condition "set use_condition to ON if the condition module and its adapters should be enabled" OFF)

It is used in the IoT Gateway project only, so unless you would like to port that the changes in this file and the addition of mach_gettime should not be necessary.

dcristoloveanu:
:clock1:

Issue #18, opened 2016-08-16
malloc(0) can return NULL or valid pointer


yaweiw:
https://github.com/arduino-libraries/AzureIoTHub/pull/25

This could lead to incorrect/inconsistent behavior on different boards.

dcristoloveanu:
Hi @yaweiw 

I think what should happen here in BUFFER_Create is that we should succeed even if the size is 0. Then in cases where the code that uses BUFFER does not want to care what the size of the buffer being created is, the BUFFER_Create calls succeed even if the size is 0.

In order to streamline contributing the change:

- Let's have the PR against develop
- Please make sure to update the .md requirement doc for buffer in the devdoc folder in order to state that BUFFER_Create with 0 size shall succeed
- Please add a unit test in the BUFFER test suite under tests.

I'll look for your PR so I can review it in a timely fashion.

Thanks for contributing,
/Dan

yaweiw:
@dcristoloveanu , ok

dcristoloveanu:
This fix been merged.

PR #20, opened 2016-08-18
PR for malloc(0) can return NULL or valid pointer #18

yaweiw:
PR for malloc(0) can return NULL or valid pointer #18

### DIFF
--- a/devdoc/buffer_requirements.md
+++ b/devdoc/buffer_requirements.md
@@ -39,7 +39,7 @@ extern BUFFER_HANDLE BUFFER_create(const unsigned char* source, size_t size);
 
 BUFFER_create creates a new buffer from the memory at source, having size "size".
 **SRS_BUFFER_02_001: [**If source is NULL then BUFFER_create shall return NULL.**]** 
-**SRS_BUFFER_02_002: [**Otherwise, BUFFER_create shall allocate memory to hold size bytes and shall copy from source size bytes into the newly allocated memory.**]** 
+**SRS_BUFFER_02_002: [**Otherwise, BUFFER_create shall allocate memory to hold size bytes and shall copy from source size bytes into the newly allocated memory. If size parameter is 0 then buffer field of BUFFER_TAG shall still hold a valid pointer whose content is '\0' and size field shall be 0.**]** 
 **SRS_BUFFER_02_003: [**If allocating memory fails, then BUFFER_create shall return NULL.**]** 
 **SRS_BUFFER_02_004: [**Otherwise, BUFFER_create shall return a non-NULL handle.**]**
 
--- a/devdoc/buffer_requirements.md
+++ b/devdoc/buffer_requirements.md
@@ -39,10 +39,10 @@ extern BUFFER_HANDLE BUFFER_create(const unsigned char* source, size_t size);
 
 BUFFER_create creates a new buffer from the memory at source, having size "size".
 **SRS_BUFFER_02_001: [**If source is NULL then BUFFER_create shall return NULL.**]** 
-**SRS_BUFFER_02_002: [**Otherwise, BUFFER_create shall allocate memory to hold size bytes and shall copy from source size bytes into the newly allocated memory. If size parameter is 0 then buffer field of BUFFER_TAG shall still hold a valid pointer whose content is '\0' and size field shall be 0.**]** 
+**SRS_BUFFER_02_002: [**Otherwise, BUFFER_create shall allocate memory to hold size bytes and shall copy from source size bytes into the newly allocated memory.**]**
 **SRS_BUFFER_02_003: [**If allocating memory fails, then BUFFER_create shall return NULL.**]** 
 **SRS_BUFFER_02_004: [**Otherwise, BUFFER_create shall return a non-NULL handle.**]**
-
+**SRS_BUFFER_02_005: [**If size parameter is 0 then 1 byte of memory shall be allocated yet size of the buffer shall be set to 0.**]**
 ###BUFFER_delete
 ```c
 void BUFFER_delete(BUFFER_HANDLE handle)
@@ -122,6 +122,7 @@ unsigned char* BUFFER_u_char(BUFFER_HANDLE handle)
 
 **SRS_BUFFER_07_025: [**BUFFER_u_char shall return a pointer to the underlying unsigned char*.**]** 
 **SRS_BUFFER_07_026: [**BUFFER_u_char shall return NULL for any error that is encountered.**]**
+**SRS_BUFFER_07_029: [**BUFFER_u_char shall return NULL if underlying buffer size is zero.**]**
  
 ###BUFFER_length
 ```c
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -38,51 +38,27 @@ BUFFER_HANDLE BUFFER_new(void)
     return (BUFFER_HANDLE)temp;
 }
 
-// return 0 if safemalloc succeeds else return non-zero
-int BUFFER_safemalloc(BUFFER_HANDLE* handleptr, size_t size)
+static int BUFFER_safemalloc(BUFFER_HANDLE* handleptr, size_t size)
 {
     int result;
-    // if handleptr or handle is NULL, return immediately
-    if (!handleptr || !(*handleptr))
+    unsigned int sizetomalloc = size;
+    if (size == 0)
     {
+        sizetomalloc = 1;
+    }
+    (*handleptr)->buffer = (unsigned char*)malloc(sizetomalloc);
+    if ((*handleptr)->buffer == NULL)
+    {
+        /*Codes_SRS_BUFFER_02_003: [If allocating memory fails, then BUFFER_create shall return NULL.]*/
+        free(*handleptr);
+        *handleptr = NULL;
         result = __LINE__;
     }
     else
     {
-        if (size == 0)
-        {
-            (*handleptr)->buffer = (unsigned char*)malloc(1);
-            if ((*handleptr)->buffer == NULL)
-            {
-                /*Codes_SRS_BUFFER_02_003: [If allocating memory fails, then BUFFER_create shall return NULL.]*/
-                free(*handleptr);
-                *handleptr = NULL;
-                result = __LINE__;
-            }
-            else
-            {
-                (*handleptr)->buffer[0] = '\0';
-                // we still consider the real buffer size is 0
-                (*handleptr)->size = 0;
-                result = 0;
-            }
-        }
-        else
-        {
-            (*handleptr)->buffer = (unsigned char*)malloc(size);
-            if ((*handleptr)->buffer == NULL)
-            {
-                /*Codes_SRS_BUFFER_02_003: [If allocating memory fails, then BUFFER_create shall return NULL.]*/
-                free(*handleptr);
-                *handleptr = NULL;
-                result = __LINE__;
-            }
-            else
-            {
-                (*handleptr)->size = size;
-                result = 0;
-            }
-        }
+        // we still consider the real buffer size is 0
+        (*handleptr)->size = size;
+        result = 0;
     }
     return result;
 }
@@ -97,7 +73,7 @@ BUFFER_HANDLE BUFFER_create(const unsigned char* source, size_t size)
     }
     else
     {
-        /*Codes_SRS_BUFFER_02_002: [Otherwise, BUFFER_create shall allocate memory to hold size bytes and shall copy from source size bytes into the newly allocated memory. If size parameter is 0 then buffer field of BUFFER_TAG shall still hold a valid pointer whose content is '\0' and size field shall be 0.] */
+        /*Codes_SRS_BUFFER_02_002: [Otherwise, BUFFER_create shall allocate memory to hold size bytes and shall copy from source size bytes into the newly allocated memory.] */
         result = (BUFFER*)malloc(sizeof(BUFFER));
         if (result == NULL)
         {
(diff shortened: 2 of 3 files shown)


yaweiw:
@dcristoloveanu FYI

devdoc/buffer_requirements.md
```diff
@@ -39,7 +39,7 @@ extern BUFFER_HANDLE BUFFER_create(const unsigned char* source, size_t size);
 
 BUFFER_create creates a new buffer from the memory at source, having size "size".
 **SRS_BUFFER_02_001: [**If source is NULL then BUFFER_create shall return NULL.**]** 
-**SRS_BUFFER_02_002: [**Otherwise, BUFFER_create shall allocate memory to hold size bytes and shall copy from source size bytes into the newly allocated memory.**]** 
+**SRS_BUFFER_02_002: [**Otherwise, BUFFER_create shall allocate memory to hold size bytes and shall copy from source size bytes into the newly allocated memory. If size parameter is 0 then buffer field of BUFFER_TAG shall still hold a valid pointer whose content is '\0' and size field shall be 0.**]**
```
dcristoloveanu: >If size parameter is 0 then buffer field of BUFFER_TAG shall still hold a valid pointer whose content is '\0' and size field shall be 0. [](start = 161, length = 136)

I'd phrase this like:

"If size parameter is 0 then no memory shall be allocated and size for the buffer shall be set to 0."

That's because I think there is no reason to allocate anything, simply setting the internal buffer to NULL should be good.

dcristoloveanu: Can you please make this a separate requirement ID (SRS_BUFFER_99_001)?
The reason we try to keep these requirements slim is so we can trace that we implemented the tests for them.
Each test is tagged with at least a requirement ID and we also tag this in the code...

We have a tool that runs as part of the gated checkin we have that can check the coverage so we know requirements are covered.

---
In reply to: [75431413](https://github.com/Azure/azure-c-shared-utility/pull/20#discussion_r75431413) [](ancestors = 75431413)

dcristoloveanu:
Sorry for the delay, I just got to this.

---
In reply to: [240610414](https://github.com/Azure/azure-c-shared-utility/pull/20#issuecomment-240610414) [](ancestors = 240610414)

src/buffer.c
```diff
@@ -309,19 +350,44 @@ int BUFFER_append(BUFFER_HANDLE handle1, BUFFER_HANDLE handle2)
...
+                    result = __LINE__;
+                }
+                else
+                {
+                    b1->buffer = temp;
+                    memcpy(b1->buffer, b2->buffer, b2->size);
+                    b1->size = b2->size;
+                    result = 0;
+                }
             }
```
dcristoloveanu: This part should not be needed, as realloc should handle this gracefully.
If b1->size is 0, b1->buffer should be NULL. realloc will simply do an alloc in this case.

From C99, 7.20.3.4
...
3 If ptr is a null pointer, the realloc function behaves like the malloc function for the
specified size.

src/buffer.c
```diff
@@ -352,21 +418,48 @@ int BUFFER_prepend(BUFFER_HANDLE handle1, BUFFER_HANDLE handle2)
...
```
dcristoloveanu: Same comment as above, realloc should handle this gracefully.

src/buffer.c
```diff
@@ -378,7 +471,7 @@ int BUFFER_prepend(BUFFER_HANDLE handle1, BUFFER_HANDLE handle2)
 unsigned char* BUFFER_u_char(BUFFER_HANDLE handle)
 {
     unsigned char* result;
-    if (handle == NULL)
+    if (handle == NULL || handle->size == 0)
```
dcristoloveanu: >|| handle->size == 0) [](start = 23, length = 21)

Please add a requirement in the .md for this and a test.

dcristoloveanu:
@anporumb is added to the review. #Closed

devdoc/buffer_requirements.md
yaweiw: with regards to "That's because I think there is no reason to allocate anything, simply setting the internal buffer to NULL should be good", this is not going to work if internal buffer is NULL e.g. IoTHubMessage_CreateFromByteArray will fail. We want to make sure when size is 0 a valid pointer is returned and can be freed.

src/buffer.c
```diff
@@ -38,6 +38,55 @@ BUFFER_HANDLE BUFFER_new(void)
     return (BUFFER_HANDLE)temp;
 }
 
+// return 0 if safemalloc succeeds else return non-zero
+int BUFFER_safemalloc(BUFFER_HANDLE* handleptr, size_t size)
```
dcristoloveanu: >int [](start = 0, length = 3)

This should be static as it is only used internally.

src/buffer.c
```diff
@@ -38,6 +38,55 @@ BUFFER_HANDLE BUFFER_new(void)
...
+{
+    int result;
+    // if handleptr or handle is NULL, return immediately
+    if (!handleptr || !(*handleptr))
+    {
+        result = __LINE__;
+    }
+    else
+    {
```
dcristoloveanu: No need to check these as this should be internal and these should already be checked.

src/buffer.c
yaweiw: when b1->size = 0, b1->buffer is "\0". so szie of realloc should be b1->size + b2->size -1. That's why we need to consider b1->size =0 and !=0 respectively.

src/buffer.c
```diff
@@ -38,6 +38,55 @@ BUFFER_HANDLE BUFFER_new(void)
...
+        if (size == 0)
+        {
+            (*handleptr)->buffer = (unsigned char*)malloc(1);
```
dcristoloveanu: >  (*handleptr)->buffer = (unsigned char*)malloc(1); [](start = 10, length = 51)

I think you could compute the amount of memory needed (1 or size), have that in a variable and have the malloc in one place (no need for 2 branches allocating)

src/buffer.c
```diff
@@ -38,6 +38,55 @@ BUFFER_HANDLE BUFFER_new(void)
...
+            if ((*handleptr)->buffer == NULL)
+            {
+                /*Codes_SRS_BUFFER_02_003: [If allocating memory fails, then BUFFER_create shall return NULL.]*/
+                free(*handleptr);
+                *handleptr = NULL;
+                result = __LINE__;
+            }
+            else
+            {
+                (*handleptr)->buffer[0] = '\0';
```
dcristoloveanu: >buffer[0] = '\0'; [](start = 30, length = 17)

No need to initialize the memory to anything. This will allow you to have one code path for calling malloc.

devdoc/buffer_requirements.md
dcristoloveanu: I understand. Thinking about it again, yes, it is better to allocate 1 byte, but there is no need to set the memory to '\0', as noone should ever use that.
The other point about having this as a separate req id stands.

---
In reply to: [75432507](https://github.com/Azure/azure-c-shared-utility/pull/20#discussion_r75432507) [](ancestors = 75432507)

dcristoloveanu:
Done reviewing, I'll have a look tomorrow morning to merge in the new iteration.

---
In reply to: [240936319](https://github.com/Azure/azure-c-shared-utility/pull/20#issuecomment-240936319) [](ancestors = 240936319,240610414)

dcristoloveanu:
:clock1:

src/buffer.c
dcristoloveanu: I agree the size of the allocated memory for b1->buffer is 1 byte.
b1->size is 0.
b2->size is let's say 10.

The amount of memory to be allocated by realloc should be 10 bytes (b1->size + b2->size). Forget about the fact that the extra byte was allocated for b1 ...

Thus there is no need for special handling.

---
In reply to: [75432767](https://github.com/Azure/azure-c-shared-utility/pull/20#discussion_r75432767) [](ancestors = 75432767)

src/buffer.c
yaweiw: agree it's not actually needed.

src/buffer.c
yaweiw: agree

yaweiw:
@dcristoloveanu please have a look at the new iteration.

tests/buffer_ut/buffer_ut.c
```diff
@@ -782,6 +787,102 @@ BEGIN_TEST_SUITE(Buffer_UnitTests)
...
     }
 
+    /* Codes_SRS_BUFFER_07_024: [BUFFER_append concatenates b2 onto b1 without modifying b2 and shall return zero on success.] */
+    TEST_FUNCTION(BUFFER_append_HANDLE2_SIZE_ZERO_SUCCEED)
+    {
+        ///arrange
+        
+        BUFFER_HANDLE handle1 = BUFFER_create(BUFFER_Test1, BUFFER_TEST1_SIZE);
+        BUFFER_HANDLE handle2 = BUFFER_create(BUFFER_Test2, 0);
+        // umock_c_reset_all_calls();
```
dcristoloveanu: >// umock_c_reset_all_calls(); [](start = 8, length = 29)

This should not be commented. Same applies for Line 808.
Basically the test should also check that no extra allocation happens because b2->size is 0.

dcristoloveanu: >// umock_c_reset_all_calls(); [](start = 8, length = 29)

Same comment like above.

tests/buffer_ut/buffer_ut.c
```diff
@@ -782,6 +787,102 @@ BEGIN_TEST_SUITE(Buffer_UnitTests)
...
+        BUFFER_delete(handle2);
+    }
+
+    /* Codes_SRS_BUFFER_07_024: [BUFFER_append concatenates b2 onto b1 without modifying b2 and shall return zero on success.] */
+    TEST_FUNCTION(BUFFER_append_HANDLE1_SIZE_ZERO_SUCCEED)
+    {
+        ///arrange
+        BUFFER_HANDLE handle1 = BUFFER_create(BUFFER_Test1, 0);
+        BUFFER_HANDLE handle2 = BUFFER_create(BUFFER_Test2, BUFFER_TEST2_SIZE);
+        // umock_c_reset_all_calls();
```
dcristoloveanu: > // umock_c_reset_all_calls( [](start = 7, length = 28)

Same comment like above

tests/buffer_ut/buffer_ut.c
```diff
@@ -782,6 +787,102 @@ BEGIN_TEST_SUITE(Buffer_UnitTests)
...
+        BUFFER_delete(handle1);
+        BUFFER_delete(handle2);
+    }
+
+    TEST_FUNCTION(BUFFER_prepend_HANDLE1_SIZE_ZERO_SUCCEED)
+    {
+        ///arrange
+        BUFFER_HANDLE handle1 = BUFFER_create(BUFFER_Test1, 0);
+        BUFFER_HANDLE handle2 = BUFFER_create(BUFFER_Test2, BUFFER_TEST2_SIZE);
+        // umock_c_reset_all_calls();
```
dcristoloveanu: >// umock_c_reset_all_calls(); [](start = 8, length = 29)

Same like above

tests/buffer_ut/buffer_ut.c
```diff
@@ -782,6 +787,102 @@ BEGIN_TEST_SUITE(Buffer_UnitTests)
...
+
+        //ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+
+        ///cleanup
+        BUFFER_delete(handle1);
+        BUFFER_delete(handle2);
+    }
+
+    TEST_FUNCTION(BUFFER_prepend_HANDLE1_SIZE_ZERO_SUCCEED)
+    {
```
dcristoloveanu: Can you please add with /* Codes ... */ the tested requirement? Thanks.

tests/buffer_ut/buffer_ut.c
```diff
@@ -782,6 +787,102 @@ BEGIN_TEST_SUITE(Buffer_UnitTests)
...
+        ASSERT_ARE_EQUAL(size_t, BUFFER_TEST2_SIZE, howBig);
+
+        //ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+
+        ///cleanup
+        BUFFER_delete(handle1);
+        BUFFER_delete(handle2);
+    }
+
+    TEST_FUNCTION(BUFFER_prepend_HANDLE2_SIZE_ZERO_SUCCEED)
```
dcristoloveanu: >T_FUNCTION(BU [](start = 7, length = 13)

Same here, please add the tested requirement.

tests/buffer_ut/buffer_ut.c
```diff
@@ -782,6 +787,102 @@ BEGIN_TEST_SUITE(Buffer_UnitTests)
...
+        ASSERT_ARE_EQUAL(size_t, BUFFER_TEST2_SIZE, howBig);
+
+        //ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+
+        ///cleanup
+        BUFFER_delete(handle1);
+        BUFFER_delete(handle2);
+    }
+
+    TEST_FUNCTION(BUFFER_prepend_HANDLE1_SIZE_ZERO_SUCCEED)
```
dcristoloveanu: >NCTIO [](start = 11, length = 5)

Can you please add with /* Tests ... */ the tested requirement? Thanks.

dcristoloveanu:
Done, only some minor changes in tests needed and I will pull this in.

---
In reply to: [240952628](https://github.com/Azure/azure-c-shared-utility/pull/20#issuecomment-240952628) [](ancestors = 240952628)

dcristoloveanu:
If you don't get to adjust the tests according to the comments I can do that myself tomorrow morning.

---
In reply to: [241129386](https://github.com/Azure/azure-c-shared-utility/pull/20#issuecomment-241129386) [](ancestors = 241129386,240952628)

yaweiw:
@dcristoloveanu feel free to add comments if you like and merge it. thanks.

dcristoloveanu:
I have updated the code to have the appropriate /* Tests comments and merging this in now.

---
In reply to: [241302088](https://github.com/Azure/azure-c-shared-utility/pull/20#issuecomment-241302088) [](ancestors = 241302088)

yaweiw:
cool thanks @dcristoloveanu

PR #21, opened 2016-08-19
updates to get OSX build working

damonbarry:
This gets azure-c-shared-utility building on OSX using the default arguments. I still need to make sure other cmake options build. I know valgrind fails, but need to consider whether that would be a different pull request.


adapters/socketio_berkeley.c
```diff
@@ -650,7 +652,11 @@ int socketio_setoption(CONCRETE_IO_HANDLE socket_io, const char* optionName, con
         }
         else if (strcmp(optionName, "tcp_keepalive_time") == 0)
         {
+#ifdef __APPLE__
```
dcristoloveanu: The same applies to iOs
I know Jamie had the same #ifdef but with

#if TARGET_OS_IPHONE

If you add both of them here we can also close Jamie's PR.

dcristoloveanu: [https://github.com/Azure/azure-c-shared-utility/pull/16](https://github.com/Azure/azure-c-shared-utility/pull/16)

---
In reply to: [75419320](https://github.com/Azure/azure-c-shared-utility/pull/21#discussion_r75419320) [](ancestors = 75419320)

dcristoloveanu: Also one more point is that I think we should simply not support any KA options for these cases, right? Why would we allow setting only some of them?

---
In reply to: [75419327](https://github.com/Azure/azure-c-shared-utility/pull/21#discussion_r75419327) [](ancestors = 75419327,75419320)

src/constbuffer.c
```diff
@@ -98,7 +98,9 @@ CONSTBUFFER_HANDLE CONSTBUFFER_CreateFromBuffer(BUFFER_HANDLE buffer)
     }
     else
     {
-        result = (CONSTBUFFER_HANDLE_DATA*)CONSTBUFFER_Create_Internal(BUFFER_u_char(buffer), BUFFER_length(buffer));
+        size_t length = BUFFER_length(buffer);
+        unsigned char* rawBuffer = BUFFER_u_char(buffer);
+        result = (CONSTBUFFER_HANDLE_DATA*)CONSTBUFFER_Create_Internal(rawBuffer, length);
```
dcristoloveanu: I like this.

testtools/micromock/inc/timediscretemicromock.h
```diff
@@ -381,6 +381,11 @@ class stims_base
     template<typename resultType, typename arg1Type, typename arg2Type, typename arg3Type, typename arg4Type, typename arg5Type, typename arg6Type, typename arg7Type, typename arg8Type, class C> friend class call8Arg;
     template<typename resultType, typename arg1Type, typename arg2Type, typename arg3Type, typename arg4Type, typename arg5Type, typename arg6Type, typename arg7Type, typename arg8Type, typename arg9Type, class C> friend class call9Arg;
     template<typename resultType, typename arg1Type, typename arg2Type, typename arg3Type, typename arg4Type, typename arg5Type, typename arg6Type, typename arg7Type, typename arg8Type, typename arg9Type, typename arg10Type, class C> friend class call10Arg;
+    template<typename resultType, typename arg1Type, typename arg2Type, typename arg3Type, typename arg4Type, typename arg5Type, typename arg6Type, typename arg7Type, typename arg8Type, typename arg9Type, typename arg10Type, typename arg11Type, class C> friend class call11Arg;
```
dcristoloveanu: >template<typename resultType, typename arg1Type, typename arg2Type, typename arg3Type, typename arg4Type, typename arg5Type, typename arg6Type, typename arg7Type, typename arg8Type, typename arg9Type, typename arg10Type, typename arg11Type, class C> friend class call11Arg; [](start = 4, length = 273)

ewwww? Why do we need to do this?

dcristoloveanu:
:shipit:

adapters/socketio_berkeley.c
damonbarry: I believe OSX supports the same three keep-alive options, it just has a different name for one of them (`IPPROTO_TCP, TCP_KEEPALIVE` instead of `SOL_TCP, TCP_KEEPIDLE`. Of course, I still need to test this...

src/constbuffer.c
damonbarry: I had to do this because umock_c can't _not_ enforce call ordering.  :(

adapters/socketio_berkeley.c
dcristoloveanu: I think we should always use IPPROTO_TCP (per [http://pubs.opengroup.org/onlinepubs/9699919799/](http://pubs.opengroup.org/onlinepubs/9699919799/))

---
In reply to: [75430277](https://github.com/Azure/azure-c-shared-utility/pull/21#discussion_r75430277) [](ancestors = 75430277)

src/constbuffer.c
dcristoloveanu: I know. I suggest asking the author for a new feature .

---
In reply to: [75430488](https://github.com/Azure/azure-c-shared-utility/pull/21#discussion_r75430488) [](ancestors = 75430488)

testtools/micromock/inc/timediscretemicromock.h
damonbarry: The compiler complained that `call(11|12|13|14|15)Arg` didn't have the friend access it needed, so...

adapters/socketio_berkeley.c
damonbarry: UPDATE: I confirmed these keep-alive options work in OSX (El Capitan), but haven't tried `IPPROTO_TCP, TCP_KEEPALIVE` in other platforms yet

damonbarry: UPDATE 2: I tried to use IPPROTO_TCP always per your comment @dcristoloveanu, but this file doesn't compile for ubuntu/gcc if I try to use `IPPROTO_TCP, TCP_KEEPALIVE`. So I'll keep the ifdef for now...

dcristoloveanu: Ok, no problem.

---
In reply to: [75620983](https://github.com/Azure/azure-c-shared-utility/pull/21#discussion_r75620983) [](ancestors = 75620983)

Issue #13, opened 2016-08-04
Fails to build with clang


damienpontifex:
My project utilise the azure-iot-sdks which has a dependency on this c-shared-utility and using clang 3.8.0 compiler. Currently the shared utility library fails to build with clang.

Some initial errors found include:

``` c
azure-iot-sdks/c/azure-c-shared-utility/src/constbuffer.c:116:9: error: address argument to atomic
      operation must be a pointer to _Atomic type ('uint32_t *' (aka 'unsigned int *') invalid)
        INC_REF(CONSTBUFFER_HANDLE_DATA, constbufferHandle);
```

There are others, but still to investigate further the amount of changes that may be required.

Would this be amenable to enabling clang compilation for this project? Or too far out of scope?

tameraw:
@damienpontifex - Currently we don't have plans to add support to build with clang. But we'd definitely welcome contribution to support that.

PR #24, opened 2016-08-24
Fixing warnings creating errors with -Werror when compiling with clang.

damienpontifex:
Builds with clang but unit test build still fails due to continued warnings in Azure-ctest submodule. So cmake generate with `-Dskip_unittests=ON` to ensure building just this project.

Tested a build with gcc also and hasn't caused any failures by these additions.

Associated with #13


olivierbloch:
Thanks for your contribution, we are reviewing it.

PR #26, opened 2016-09-06
Fix lib install locations for multilib environments

srware:
When installing libs or header files CMake variables should be used rather than using hard-coded paths. This can cause issues with multilib environments for example.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,9 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
     set(LINUX TRUE)
 endif()
 
+# Set CMAKE_INSTALL_LIBDIR if not defined
+include(GNUInstallDirs)
+
 include (CTest)
 
 #these are the include folders
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,7 +53,11 @@ option(compileOption_C "passes a string to the command line of the C compiler" O
 option(compileOption_CXX "passes a string to the command line of the C++ compiler" OFF)
 
 # Start of variables used during install
-set (LIB_INSTALL_DIR lib CACHE PATH "Library object file directory")
+if (CMAKE_INSTALL_LIBDIR)
+  set (LIB_INSTALL_DIR "${CMAKE_INSTALL_LIBDIR}" CACHE PATH "Library object file directory")
+else ()
+  set (LIB_INSTALL_DIR "lib" CACHE PATH "Library object file directory")
+endif ()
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -807,7 +811,7 @@ endif()
 
 if(WIN32)
 else()
-    install (TARGETS aziotsharedutil DESTINATION lib)
-    install (FILES ${source_h_files} DESTINATION include/azureiot/azure_c_shared_utility)
+    install (TARGETS aziotsharedutil DESTINATION ${LIB_INSTALL_DIR})
+    install (FILES ${source_h_files} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/azureiot/azure_c_shared_utility)
```
andrew-buckley: I don't think CMAKE_INSTALL_INCLUDEDIR is a default CMake variable. Or is it?

andrew-buckley:
This looks like the advantage is that it lets the user set where they want to install the libs and include files separately. Alternatively, a user can override the CMAKE_INSTALL_PREFIX variable to place the libs and include files somewhere other than the default. Would this accomplish what you're thinking?

srware:
These variables are standard CMake variables which if not already populated will be populated by including GNUInstallDirs (https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html). I forgot to add the include but have updated the pull request.

I have put in a catch as a precaution for older CMake versions in case CMAKE_INSTALL_LIBDIR is not populated but I don't think this is necessary for 2.8.11 and above. CMAKE_INSTALL_PREFIX can still be used with this implementation.

CMakeLists.txt
```diff
@@ -53,7 +56,11 @@ option(compileOption_C "passes a string to the command line of the C compiler" O
 option(compileOption_CXX "passes a string to the command line of the C++ compiler" OFF)
 
 # Start of variables used during install
-set (LIB_INSTALL_DIR lib CACHE PATH "Library object file directory")
+if (CMAKE_INSTALL_LIBDIR)
```
andrew-buckley: This variable will always be set, right? If you include GNUInstallDirs. And the default is already "lib" I think.

andrew-buckley:
@srware can you make this PR against develop instead of master?

CMakeLists.txt
andrew-buckley: I'm seeing you mentioned your reasoning in the previous comment.

srware:
Going to close this one and open a new PR against develop rather than creating a merge or rebase mess :)

PR #27, opened 2016-09-07
Add missing link library UUID

srware:
When building with Yocto I get multiple 'undefined reference' errors relating to uuid functions when linking against the shared utility library unless 'uuid' is included as a target link library.


andrew-buckley:
Hi @srware. I actually ran into this problem yesterday while working on making our gateway sdk installable/finable in cmake. This looks good to me, maybe @jebrando would like to have a look?

PR #30, opened 2016-09-09
apple openssl include directories

damienpontifex:
Errors when building on OS X for openssl header files not found. I have openssl installed with homebrew and adding this enabled a successful build with `OpenSSLDir='/usr/local/opt/openssl'`


olivierbloch:
Thanks for your contribution. We will pull this one in real soon.

PR #31, opened 2016-09-19
Fix OpenSSL memory leak after deinit().

SparhawkSoftware:
Minor issue.

### DIFF
--- a/src/tlsio_openssl.c
+++ b/src/tlsio_openssl.c
@@ -927,13 +927,13 @@ void tlsio_openssl_deinit(void)
 {
     openssl_dynamic_locks_uninstall();
     openssl_static_locks_uninstall();
-
 	FIPS_mode_set(0);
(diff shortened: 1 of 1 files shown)


src/tlsio_openssl.c
```diff
@@ -927,13 +927,13 @@ void tlsio_openssl_deinit(void)
...
     openssl_dynamic_locks_uninstall();
     openssl_static_locks_uninstall();
-
 	FIPS_mode_set(0);
 	CRYPTO_set_locking_callback(NULL);
 	CRYPTO_set_id_callback(NULL);
 	ERR_remove_state(0);
     ERR_free_strings();
 	EVP_cleanup();
+	SSL_COMP_free_compression_methods();
```
tameraw: In the latest OpenSSL version this is deprecated. Could you please add an ifdef to guard this.

SparhawkSoftware:
Added #ifdef as requested.

PR #36, opened 2016-10-25
Port for ESP8266

ritazh:
@olivierbloch @tameraw @dcristoloveanu 

This PR is part of the ESP8266 port of Azure IoT C SDK. It contains updates to C utility to enable ESP8266. Please take a 👀 . Thanks! Will squash commits after review is done.

Future PRs will contain deployment scripts for copying necessary files needed to run Azure IoT C sdk and mqtt samples as part of the ESP8266 RTOS SDK.


ritazh:
@mamokarz Thanks for taking a look at this PR. I think you reviewed the earliest commit instead of all my changes as a whole under the `File changed` tab. I just squashed all the commits to make it easier for review. Thanks! :bowing_woman:

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,960 @@
...
+    TLS_IO_INSTANCE* result;
+
+    if (tls_io_config == NULL)
+    {
+        result = NULL;
+        LogError("NULL tls_io_config.");
+    }
+    else
+    {
+        result = malloc(sizeof(TLS_IO_INSTANCE));
```
mamokarz: Please cast the result of malloc to TLS_IO_INSTANCE

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,960 @@
...
+            result->on_io_error_context = NULL;
+
+            result->tlsio_state = TLSIO_STATE_NOT_OPEN;
+
+            result->x509certificate = NULL;
+            result->x509privatekey = NULL;
+        }
+    }
+
+    return result;
```
mamokarz: Please cast the result to CONCRETE_IO_HANDLE

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,960 @@
...
+    failed3:
+        //close(socket);
+        return ;
+}
+
+static int send_handshake_bytes(TLS_IO_INSTANCE* tls_io_instance)
+{
+    system_print_meminfo();
+    LogError("free heap size %d", system_get_free_heap_size());
+    openssl_thread_LWIP_CONNECTION(tls_io_instance);
```
mamokarz: It looks like the openssl_thread_LWIP_CONNECTION is, in fact, the xio_open. If I'm right...
1. Please move it together with the other xio functions.
2. Consider to make it async replacing the for(;;) by the xio_dowork.
3. Change the tlsio_state to TLSIO_STATE_OPEN and call the indicate_open_complete only when the xio call its own open_complete callback (on_underlying_io_open_complete).

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,960 @@
...
+            goto failed3;
+        } else{
+            os_printf("correct\n");
+        }
+    }
+    os_printf("OK\n");
+
+    char recv_buf[128];
+    int recv_bytes = sizeof(recv_buf);
+    for(;;){
```
mamokarz: Consider to have a timeout on this loop, and a delay between retries.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,960 @@
...
+    int sock_errno = 0;
+    u32_t optlen = sizeof(sock_errno);
+    getsockopt(fd, SOL_SOCKET, SO_ERROR, &sock_errno, &optlen);
+    return sock_errno;
+}
+
+static void httpconn_set_non_block(int fd) 
+{
+  int flags = -1;
+  int error = 0;
```
ritazh: Can you please explain why initialization is not required here? I thought it's always good to initialize any variable defined.

adapters/tlsio_ssl_esp8266.c
ritazh: For item 1,  `tlsio_openssl_open` calls `send_handshake_bytes` which calls `openssl_thread_LWIP_CONNECTION`,  so this function is actually indirectly called by `tlsio_openssl_open`. Is this what you mean?

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,960 @@
...
+    return 0;
+}
+
+
+static int create_openssl_instance(TLS_IO_INSTANCE* tlsInstance)
+{
+    return 0;
+}
+
+int tlsio_openssl_init(void)
```
ritazh: I am following the Linux implementation [here](https://github.com/Azure/azure-c-shared-utility/blob/develop/src/tlsio_openssl.c#L908). That's why you see this function here.  Even though I'm not doing anything here, I want to make a stub function here in case we need to add anything for initialization in the future and to be consistent with other platforms. What do you think?

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,960 @@
...
+            tls_io_instance->on_io_open_complete_context = on_io_open_complete_context;
+
+            tls_io_instance->on_bytes_received = on_bytes_received;
+            tls_io_instance->on_bytes_received_context = on_bytes_received_context;
+
+            tls_io_instance->on_io_error = on_io_error;
+            tls_io_instance->on_io_error_context = on_io_error_context;
+
+            tls_io_instance->tlsio_state = TLSIO_STATE_OPENING_UNDERLYING_IO;
+
```
ritazh: Same explanation provided regarding underlying_io.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,1085 @@
...
+    {
+        LogError("invalid parameter detected: const char* name=%p, const void* value=%p", name, value);
+    }
+    else
+    {
+        if (
+            (strcmp(name, "TrustedCerts") == 0) ||
+            (strcmp(name, "x509certificate") == 0) ||
+            (strcmp(name, "x509privatekey") == 0)
+        )
```
ritazh: I'm following the current [implementation of tlsio_openssl.c](https://github.com/Azure/azure-c-shared-utility/blob/develop/src/tlsio_openssl.c#L147). If this needs to be different, can you provide an example? Thanks!

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,960 @@
...
+            os_printf("lwip_select %d\n", ret);
+        }
+    }
+
+    int s = socket;
+    ctx = SSL_CTX_new(TLSv1_client_method());
+    if (!ctx) {
+        os_printf("failed\n");
+
+    }
```
ritazh: I have added a `return;` to the failed case.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,960 @@
...
+{
+    return 0;
+}
+
+int tlsio_openssl_init(void)
+{
+    return 0;
+}
+
+void tlsio_openssl_deinit(void)
```
ritazh: These are common functions defined in the `tlsio_openssl.h` file. The `tlsio_openssl.h` file is used by `tlsio_ssl_esp8266.c`. Just to be clear, `tlsio_ssl_esp8266.c` is using the openssl interface. But the difference is that it does not use BIO. If the name is changed, I will need to create a new header file for this, which seems unnecessary.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,960 @@
...
+    {
+        LogError("NULL tls_io.");
+    }
+    else
+    {
+        TLS_IO_INSTANCE* tls_io_instance = (TLS_IO_INSTANCE*)tls_io;
+        if (tls_io_instance->certificate != NULL)
+        {
+            free(tls_io_instance->certificate);
+            tls_io_instance->certificate = NULL;
```
ritazh: This is the same as [tlsio_openssl.c](https://github.com/Azure/azure-c-shared-utility/blob/develop/src/tlsio_openssl.c#L1010). I think it's always a good idea to initialize variables. I have experienced many bugs because of uninitialized variables.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,960 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include <stdlib.h>
+#ifdef _CRTDBG_MAP_ALLOC
+#include <crtdbg.h>
+#endif
+
+#include "openssl/ssl_compat-1.0.h"
```
ritazh: Thanks @mamokarz! Do you have a sample unit test I can use to get started?

ritazh:
@mamokarz Just pushed a new commit adding updates based on your feedback. Please take another look. Thanks! :bowing_woman:

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,960 @@
...
+            result->on_io_close_complete = NULL;
+            result->on_io_close_complete_context = NULL;
+
+            result->on_io_error = NULL;
+            result->on_io_error_context = NULL;
+
+            result->tlsio_state = TLSIO_STATE_NOT_OPEN;
+
+            result->x509certificate = NULL;
+            result->x509privatekey = NULL;
```
mamokarz: In this case, it is not clear to me, where the tls_io_instance->underlying_io came from. You are using it as the handler for xio_destry, or xio_close, but you are not attributing any value to it.

You can call the ssl in 2 ways:

1. Use the xio as the interface between the tlsio and the ssl, something like
        [ MQTT ] ----<xio>---> [ tlsio ] ----<xio>---> [ underlySSL ] ---->[ ssl ]
2. Or, call the ssl directly from the tlsio
        [ MQTT ] ----<xio>---> [ tlsio ] ----> [ underlySSL ] ---->[ ssl ]

For instance, Arduino uses the sequence 2, the sslClient is equivalent to the openssl. 
        [ httpapi_compact ] ----<xio>---> [ tlsio_arduino ] ----> [ sslClient_arduino ] ---->[ sslClient ]

So, sounds me that you are implementing the sequence 2, but you have some calls to xio as the sequence 1. Is that correct?

adapters/tlsio_ssl_esp8266.c
mamokarz: Agree, but in this case, you are destroying the whole tls_io_instance.

adapters/tlsio_ssl_esp8266.c
mamokarz: I see, ok.

adapters/tlsio_ssl_esp8266.c
mamokarz: I took a look how the sdk use it, and it is not necessary to establish relation between the name and value. The implementation is correct as is.

mamokarz: I took a look how the sdk use it, and it is not necessary to establish relation between the name and value. The implementation is correct as is.

adapters/tlsio_ssl_esp8266.c
mamokarz: I have concerns about a function that do not belongs to the standard interface. It can clearly became a problem in the future. I prefer to joint the init functions to the create and the deinit to destroy. If it is not possible, we should add this function to the xio interface.

mamokarz: I have concerns about a function that do not belongs to the standard interface. It can clearly became a problem in the future. I prefer to joint the init functions to the create and the deinit to destroy. If it is not possible, we should add this function to the xio interface.

mamokarz: I have concerns about a function that do not belongs to the standard interface. It can clearly became a problem in the future. I prefer to joint the init functions to the create and the deinit to destroy. If it is not possible, we should add this function to the xio interface.

mamokarz: I have concerns about a function that do not belongs to the standard interface. It can clearly became a problem in the future. I prefer to joint the init functions to the create and the deinit to destroy. If it is not possible, we should add this function to the xio interface.

adapters/tlsio_ssl_esp8266.c
ritazh: The current implementation is actually: 
[ MQTT ] ----< xio >---> [ tlsio ] ----> [ ESP8266 SSL ] . I have updated the code to reflect this.

adapters/tlsio_ssl_esp8266.c
ritazh: I see your point here. However, I would like to be consistent with the Linux implementation in tlsio_openssl.c (https://github.com/Azure/azure-c-shared-utility/blob/develop/src/tlsio_openssl.c#L1010). Also I think it really does not hurt to zero the memory before freeing it. In some situations, this is safer.

ritazh:
@mamokarz Just pushed a new commit removing underlying_io, init, and deinit based on your feedback. Please take another look. Thanks!

ritazh:
@mamokarz Unit tests have been added. Please take another look. Thanks!

inc/azure_c_shared_utility/xlogging.h
```diff
@@ -39,19 +39,32 @@ typedef void(*LOGGER_LOG)(LOG_CATEGORY log_category, const char* file, const cha
...
+#elif defined(ESP8266_RTOS)
+#include "c_types.h"
+#define LogInfo(FORMAT, ...) do {    \
+        static const char flash_str[] ICACHE_RODATA_ATTR STORE_ATTR = FORMAT;  \
+        printf(flash_str, ##__VA_ARGS__);   \
+        printf("\n");\
+    } while(0)
+
+#define LogError LogInfo
+#define LOG(...)
```
mamokarz: Please consider to add the code to print logs from the LOG call. It can ignore the log_category and log_options, and act as the LogInfo.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+    tlsio_openssl_open,
+    tlsio_openssl_close,
+    tlsio_openssl_send,
+    tlsio_openssl_dowork,
+    tlsio_openssl_setoption
+};
+
+static LOCK_HANDLE * openssl_locks = NULL;
+
+static struct CRYPTO_dynlock_value* openssl_dynamic_locks_create_cb(const char* file, int line)
```
mamokarz: Some compilers consider that parameter are variables that you declare, and you must use it in someway. If you do not use this variables, the compiler will throw a compilation error. You can avoid this problem adding something like:

(void)(file, line);

on the beginning of the function implementation.

You need to fix it here and in all the other functions that you do not use some of the parameters.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+    fd_set writeset;
+    fd_set errset;
+
+    ip_addr_t target_ip;
+    SSL_CTX *ctx;
+    SSL *ssl;
+
+    TLS_IO_INSTANCE* tls_io_instance = (TLS_IO_INSTANCE*)p;
+
+    LogInfo("OpenSSL thread start...\n");
```
mamokarz: LogInfo already add a \n at the end of the text. Please consider remove it, from here and the other LogInfos.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+    // LogInfo("get target IP is %d.%d.%d.%d\n", (unsigned char)((target_ip.addr & 0x000000ff) >> 0),
+    //                                             (unsigned char)((target_ip.addr & 0x0000ff00) >> 8),
+    //                                             (unsigned char)((target_ip.addr & 0x00ff0000) >> 16),
+    //                                             (unsigned char)((target_ip.addr & 0xff000000) >> 24));
+
+    LogInfo("create socket ......");
+    sock = socket(AF_INET, SOCK_STREAM, 0);
+    if (sock < 0) {
+        LogError("create socket failed\n");
+        return;
```
mamokarz: Our code guidelines do not allow returns in the middle of the function. Please use only one return at the end of the function.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+    }
+    LogInfo("create socket OK\n");
+
+    LogInfo("bind socket ......");
```
mamokarz: Please consider reduce the number of LogInfo in this function.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+            LogError("socket connect failed %s\n", tls_io_instance->hostname);
+            return;
+        } else{
+            LogInfo("socket connect in progress\n");
+        }
+    }
+    LogInfo("socket connect OK\n");
+
+    char recv_buf[128];
+    int recv_bytes = (int)sizeof(recv_buf);
```
mamokarz: Please define recv_bytes as size_t besides int.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+          }
+      } else{
+          break;
+      }
+  }
+
+}
+
+
+LOCAL void openssl_thread_LWIP_CONNECTION(void *p)
```
mamokarz: Please consider use TLS_IO_INSTANCE* tls_io_instance as the parameter for these function besides void* p.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+{
+    int ret = -1;
+
+    int sock = -1;
```
mamokarz: Please do not initialize these variables here with a 'predict' value. It is better to have a compilation error if you try to use it without initialization than introduce a bug in the execution time because you are using it with a wrong value.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+            LogInfo("socket connect in progress\n");
+        }
+    }
+    LogInfo("socket connect OK\n");
+
+    char recv_buf[128];
+    int recv_bytes = (int)sizeof(recv_buf);
+    int retry = 0;
+
+    while (retry < 10){
```
mamokarz: Please consider create a constant for the number of retries besides use directly the number 10.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+    
+            if (FD_ISSET(sock, &readset)){
+                memset(recv_buf, 0, recv_bytes);
+                break;
+            }
+        }else{
+            LogInfo("lwip_select returned: %d\n", ret);
+        }
+
+        retry++;
```
mamokarz: Don't it needs a small delay between retries?

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+    }
+
+    int s = sock;
```
mamokarz: It is not clear for me why do you need to create this 's' variable here. Please consider to replace s by sock in the rest of the function and avoid this copy.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+    ctx = SSL_CTX_new(TLSv1_client_method());
+    if (!ctx) {
+        LogError("create new SSL CTX failed\n");
+        return;
+    }
+    LogInfo("create new SSL CTX OK\n");
+    
+    LogInfo("SSL set fragment\n");
+    ret = SSL_set_fragment(ctx, OPENSSL_FRAGMENT_SIZE);
```
mamokarz: There is no propose to store the result of the SSL_set_fragment if you will not use it. Please consider to judge this result.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+    while (-1 == SSL_connect(ssl))
+    {  
+        FD_ZERO(&readset);
+        FD_SET(s, &readset);
+        FD_ZERO(&writeset);
+        FD_SET(s, &writeset);
+        FD_ZERO(&errset);
+        FD_SET(s, &errset);
+
+        ret = lwip_select(s + 1, &readset, &writeset, &errset, NULL);
```
mamokarz: There is no propose to store the result of the lwip_select if you will not use it. Please consider to judge this result.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+
+            if (create_openssl_instance(tls_io_instance) != 0)
+            {
+
+                tls_io_instance->tlsio_state = TLSIO_STATE_ERROR;
+                result = __LINE__;
+            }
+            else
+            {
+                send_handshake_bytes(tls_io_instance);
```
mamokarz: Please create the mechanism to propagate errors in the handshake to this point and check it to decide if this function shall return success of fail.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+}
+
+static int send_handshake_bytes(TLS_IO_INSTANCE* tls_io_instance)
+{
+    //system_print_meminfo(); // This is useful for debugging purpose.
+    //LogInfo("free heap size %d", system_get_free_heap_size()); // This is useful for debugging purpose.
+    openssl_thread_LWIP_CONNECTION(tls_io_instance);
+    tls_io_instance->tlsio_state = TLSIO_STATE_OPEN;
+    indicate_open_complete(tls_io_instance, IO_OPEN_OK);    
+    int result = 0;
```
mamokarz: Please check if the calls on this function had success before return success or fail.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+
+    return result;
+}
+
+
+static int decode_ssl_received_bytes(TLS_IO_INSTANCE* tls_io_instance)
+{
+    int result = 0;
```
mamokarz: Please do not initialize this result predicting success.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+    unsigned char buffer[64];
+
+    int rcv_bytes = 1;
```
mamokarz: Please do not initialize rcv_bytes with a predicted value.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+    rcv_bytes = SSL_read(tls_io_instance->ssl, buffer, sizeof(buffer));
+    LogInfo("decode ssl recv bytes: %d\n", rcv_bytes);
```
mamokarz: I agree that this log is useful for debug, but it will probably generate lots of line. Please consider remove it.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+    }
+
+    return result;
+}
+
+static void destroy_openssl_instance(TLS_IO_INSTANCE* tls_io_instance)
+{
+    if (tls_io_instance != NULL)
+    {
+        SSL_free(tls_io_instance->ssl);
```
mamokarz: Please check it ssl is not NULL before call SSL_free.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+        SSL_CTX_free(tls_io_instance->ssl_context);
```
mamokarz: Please check it ssl_context is not NULL before call SSL_CTX_free.

mamokarz: Please make ssl and ssl_context equals NULL.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+static int create_openssl_instance(TLS_IO_INSTANCE* tlsInstance)
+{
+    return 0;
+}
+
+
+/* Codes_SRS_TLSIO_SSL_ESP8266_99_005: [ The tlsio_openssl_create succeed. ]*/
+CONCRETE_IO_HANDLE tlsio_openssl_create(void* io_create_parameters)
+{
+    TLSIO_CONFIG* tls_io_config = io_create_parameters;
```
mamokarz: Please cast io_create_parameters to TLSIO_CONFIG*

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+    {
+        LogError("NULL tls_io.");
+    }
+    else
+    {
+        TLS_IO_INSTANCE* tls_io_instance = (TLS_IO_INSTANCE*)tls_io;
+        if (tls_io_instance->certificate != NULL)
+        {
+            free(tls_io_instance->certificate);
+            tls_io_instance->certificate = NULL;
```
mamokarz: Because you are destroying the tls_io_instance, there is no reason to make certificate equals NULL here. Please consider remove this attribution.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+    }
+    else
+    {
+
+        TLS_IO_INSTANCE* tls_io_instance = (TLS_IO_INSTANCE*)tls_io;
+
+        /* Codes_SRS_TLSIO_SSL_ESP8266_99_007: [ The tlsio_openssl_open invalid state. ]*/
+        if (tls_io_instance->tlsio_state != TLSIO_STATE_NOT_OPEN)
+        {
+            result = __LINE__;
```
mamokarz: You probably need to change the state here
tls_io_instance->tlsio_state = TLSIO_STATE_ERROR;

mamokarz: If callbacks was provided, you probably need to call on_io_error and on_io_open_complete with IO_OPEN_ERROR.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+
+            tls_io_instance->on_io_error = on_io_error;
+            tls_io_instance->on_io_error_context = on_io_error_context;
+
+            tls_io_instance->tlsio_state = TLSIO_STATE_OPENING;
+
+            if (create_openssl_instance(tls_io_instance) != 0)
+            {
+
+                tls_io_instance->tlsio_state = TLSIO_STATE_ERROR;
```
mamokarz: If callbacks was provided, you probably need to call on_io_error and on_io_open_complete with IO_OPEN_ERROR.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+#include "azure_c_shared_utility/crt_abstractions.h"
+
+#define OPENSSL_FRAGMENT_SIZE 5120
+#define OPENSSL_LOCAL_TCP_PORT 1000
+
+typedef enum TLSIO_STATE_TAG
+{
+    TLSIO_STATE_NOT_OPEN,
+    TLSIO_STATE_OPENING,
+    TLSIO_STATE_IN_HANDSHAKE,
```
mamokarz: Please remove TLSIO_STATE_IN_HANDSHAKE form this enum

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+    {
+        result = __LINE__;
+        LogError("NULL tls_io.");
+    }
+    else
+    {
+        TLS_IO_INSTANCE* tls_io_instance = (TLS_IO_INSTANCE*)tls_io;
+
+        if ((tls_io_instance->tlsio_state == TLSIO_STATE_NOT_OPEN) ||
+            (tls_io_instance->tlsio_state == TLSIO_STATE_CLOSING))
```
mamokarz: The best behavior here is probably

        if ((tls_io_instance->tlsio_state == TLSIO_STATE_NOT_OPEN) ||
            (tls_io_instance->tlsio_state == TLSIO_STATE_ERROR))
        {
            result = 0;
            tls_io_instance->tlsio_state = TLSIO_STATE_NOT_OPEN;
            LogInfo("Wrong tlsio_state. Expected state is TLSIO_STATE_OPEN.");
        }
        else if ((tls_io_instance->tlsio_state == TLSIO_STATE_OPENING) ||
            (tls_io_instance->tlsio_state == TLSIO_STATE_CLOSING))
        {
            result = __LINE__;
            tls_io_instance->tlsio_state = TLSIO_STATE_ERROR;
            LogError("Invalid tlsio_state. Expected state is TLSIO_STATE_OPEN.");
        }
        else
        {
            tls_io_instance->tlsio_state = TLSIO_STATE_CLOSING;
            tls_io_instance->on_io_close_complete = on_io_close_complete;
            tls_io_instance->on_io_close_complete_context = callback_context;

            destroy_openssl_instance(tls_io_instance);
            result = 0;
        }

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+            result = __LINE__;
+            LogError("Invalid tlsio_state. Expected state is TLSIO_STATE_OPEN.");
+        }
+        else
+        {
+            int total_write = 0;
+            int res = 0;
+            int retry = 0;
+
+            while(size > 0 && retry < 500){
```
mamokarz: Sounds me that 500 is too much. Please define a constant for the maximum send retries, and consider a smaller value, probably near 10. Consider that if SSL_write returns 0 is because something is wrong.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+                }
+                else
+                {
+                    retry++;
+                }
+            }
+
+            if (retry >= 500)
+            {
+                result = 1;
```
mamokarz: Please call on_send_complete with IO_SEND_ERROR.

As a standard result should receive __LINE__.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+            }
+            else
+            {
+                result = 0;
```
mamokarz: Please call on_send_complete with IO_SEND_OK.

adapters/tlsio_ssl_esp8266.c
mamokarz: If it failed, and if callbacks was provided, you probably need to call on_io_error and on_io_open_complete with IO_OPEN_ERROR.

mamokarz: If it failed, you probably need to change the state:

tls_io_instance->tlsio_state = TLSIO_STATE_ERROR;

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+        if ((tls_io_instance->tlsio_state == TLSIO_STATE_NOT_OPEN) ||
+            (tls_io_instance->tlsio_state == TLSIO_STATE_CLOSING))
+        {
+            /* Codes_SRS_TLSIO_SSL_ESP8266_99_012: [ The tlsio_openssl_close wrong state.]*/
+            result = __LINE__;
+            LogError("Invalid tlsio_state. Expected state is TLSIO_STATE_NOT_OPEN or TLSIO_STATE_CLOSING.");
+        }
+        else
+        {
+            tls_io_instance->tlsio_state = TLSIO_STATE_CLOSING;
```
mamokarz: Nothing is changing the state from TLSIO_STATE_CLOSING to TLSIO_STATE_NOT_OPEN. You probably need to do that in someway after call destroy_openssl_instance.

tests/tlsio_esp8266_ut/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+#Copyright (c) Microsoft. All rights reserved.
+#Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#this is CMakeLists.txt for <myTestName>_ut
```
mamokarz: Please exchange <myTestName>_ut by tlsio_esp8266_ut

tests/tlsio_esp8266_ut/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+# ../../src/<myTestOnSRC>.c
+##
+../../adapters/tlsio_ssl_esp8266.c
+real_crt_abstractions.c
+)
+
+set(${theseTestsName}_h_files
+##
+# Include all headers that you need to execute the test. Normally we don't need any.
+##
```
mamokarz: You can remove this comment.

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
```diff
@@ -0,0 +1,733 @@
...
+        ///arrange
+        int result = 0;
+        TLS_IO_INSTANCE instance;
+        instance.on_bytes_received = on_bytes_received;
+
+        const IO_INTERFACE_DESCRIPTION* tlsioInterfaces = tlsio_openssl_get_interface_description();
+        ASSERT_IS_NOT_NULL(tlsioInterfaces);
+
+        umock_c_reset_all_calls();
+        EXPECTED_CALL(SSL_read(IGNORED_PTR_ARG, IGNORED_PTR_ARG, IGNORED_NUM_ARG));
```
mamokarz: Please add an unit test to check the callback when dowork read new valid data. And another if no new data is available.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+    if (tls_io == NULL)
+    {
+        /* Codes_SRS_TLSIO_SSL_ESP8266_99_018: [ The tlsio_openssl_dowork NULL parameter. No crash when passing NULL]*/
+        LogError("NULL tls_io.");
+    }
+    else
+    {
+        TLS_IO_INSTANCE* tls_io_instance = (TLS_IO_INSTANCE*)tls_io;
+
+        decode_ssl_received_bytes(tls_io_instance);
```
mamokarz: Only ready new data if it is connected (state=OPEN).

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
```diff
@@ -0,0 +1,733 @@
...
+
+        ///act
+        tlsioInterfaces->concrete_io_dowork(&instance);
```
mamokarz: Please add unit test for dowork in different tlsio states.

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
```diff
@@ -0,0 +1,733 @@
...
+     * The test suite will call this function to cleanup your machine for the next test.
+     * It is called after execute each test.
+     */
+    TEST_FUNCTION_CLEANUP(cleans)
+    {
+        TEST_MUTEX_RELEASE(g_testByTest);
+    }
+
+    /* Tests_SRS_TLSIO_SSL_ESP8266_99_019: [ The tlsio_openssl_dowrok succeed]*/
+    TEST_FUNCTION(tlsio_openssl_dowork__succeed)
```
mamokarz: This test must be failed, you are reading data from a TLS without connection.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+        else
+        {
+            int total_write = 0;
+            int res = 0;
+            int retry = 0;
+
+            while(size > 0 && retry < 500){
+                /* Codes_SRS_TLSIO_SSL_ESP8266_99_016: [ The tlsio_openssl_send SSL_write success]*/
+                /* Codes_SRS_TLSIO_SSL_ESP8266_99_017: [ The tlsio_openssl_send SSL_write failure]*/
+                res = SSL_write(tls_io_instance->ssl, ((uint8*)buffer)+total_write, size);
```
mamokarz: Please test if the buffer is not NULL before use it.

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
```diff
@@ -0,0 +1,733 @@
...
+    {
+        ///arrange
+        int result;
+        const IO_INTERFACE_DESCRIPTION* tlsioInterfaces = tlsio_openssl_get_interface_description();
+        ASSERT_IS_NOT_NULL(tlsioInterfaces);
+
+        TLS_IO_INSTANCE instance;
+        instance.tlsio_state = TLSIO_STATE_NOT_OPEN;
+        ///act
+        result = tlsioInterfaces->concrete_io_send(&instance, NULL, 0,  NULL, NULL);
```
mamokarz: Please add unit test for the send callback.

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
```diff
@@ -0,0 +1,733 @@
...
+
+        const IO_INTERFACE_DESCRIPTION* tlsioInterfaces = tlsio_openssl_get_interface_description();
+        ASSERT_IS_NOT_NULL(tlsioInterfaces);
+
+        umock_c_reset_all_calls();
+        STRICT_EXPECTED_CALL(SSL_free(IGNORED_PTR_ARG)).IgnoreArgument(1);
+        STRICT_EXPECTED_CALL(SSL_CTX_free(IGNORED_PTR_ARG)).IgnoreArgument(1);
+
+        ///act
+        tlsioInterfaces->concrete_io_close(&instance, NULL, NULL);
```
mamokarz: Add unit test for close callback.

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
```diff
@@ -0,0 +1,733 @@
...
+        ///act
+        result = tlsioInterfaces->concrete_io_close(&instance, NULL, NULL);
+
+        ///assert
+        ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+        ASSERT_ARE_NOT_EQUAL(int, 0, result);
+
+        instance.tlsio_state = TLSIO_STATE_CLOSING;
+        result = tlsioInterfaces->concrete_io_close(&instance, NULL, NULL);
+        ///act
```
mamokarz: Please move the comment before call concrete_io_close.

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
```diff
@@ -0,0 +1,733 @@
...
+        STRICT_EXPECTED_CALL(TLSv1_client_method());
+        STRICT_EXPECTED_CALL(SSL_CTX_new(IGNORED_PTR_ARG)).IgnoreArgument(1);
+        STRICT_EXPECTED_CALL(SSL_set_fragment(IGNORED_PTR_ARG,IGNORED_NUM_ARG)).IgnoreArgument(1).IgnoreArgument(2);
+        STRICT_EXPECTED_CALL(SSL_new(IGNORED_PTR_ARG)).IgnoreArgument(1);
+        STRICT_EXPECTED_CALL(SSL_set_fd(IGNORED_PTR_ARG, IGNORED_NUM_ARG)).IgnoreArgument(1).IgnoreArgument(2);
+        STRICT_EXPECTED_CALL(SSL_connect(IGNORED_PTR_ARG)).IgnoreArgument(1);
+        
+
+        ///act
+        result = tlsioInterfaces->concrete_io_open(&tls_io_instance, test_on_io_open_complete, NULL, NULL, NULL, NULL, NULL);
```
mamokarz: Please check the open_complete callback, and add test for the other callbacks.

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
```diff
@@ -0,0 +1,733 @@
...
+    {
+        ///arrange
+        int result;
+        const IO_INTERFACE_DESCRIPTION* tlsioInterfaces = tlsio_openssl_get_interface_description();
+        ASSERT_IS_NOT_NULL(tlsioInterfaces);
+
+        TLS_IO_INSTANCE instance;
+        instance.tlsio_state = TLSIO_STATE_NOT_OPEN;
+        ///act
+        result = tlsioInterfaces->concrete_io_close(&instance, NULL, NULL);
```
mamokarz: Most of these unit tests change the tlsio state. Please include a assert on all of then to check if the final state is correct.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+    return;
+}
+
+void ICACHE_FLASH_ATTR ERR_error_string(uint32 error, char* out)
+{
+    return;
+}
+
+/*this function will clone an option given by name and value*/
+static void* tlsio_openssl_CloneOption(const char* name, const void* value)
```
mamokarz: I cannot find any call to these function.

mamokarz:
:-1:

mamokarz:
Hi Rita,

I apologize for the wrong feedback at the end of your review, I removed it. 

Marcos

damonbarry:
@ritazh Any update?

ritazh:
@damonbarry Will pick this back up again. Had to work on few other projects before. Sorry for the major delay. :bowing_woman:

adapters/tlsio_ssl_esp8266.c
ritazh: This is derived from tlsio_openssl.h, which declares the function with `void* io_create_parameters`
```
MOCKABLE_FUNCTION(, CONCRETE_IO_HANDLE, tlsio_openssl_create, void*, io_create_parameters);
```

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,712 @@
...
+    int sock_errno = 0;
+    u32_t optlen = sizeof(sock_errno);
+    getsockopt(fd, SOL_SOCKET, SO_ERROR, &sock_errno, &optlen);
+    return sock_errno;
+}
+
+static void lwip_set_non_block(int fd) 
+{
+  int flags = -1;
+  int error = 0;
```
ritazh: These are global variables shared by both of the while loops. They need to be outside.

adapters/tlsio_ssl_esp8266.c
ritazh: This value is based on lots of testing to ensure data is written properly.

adapters/threadapi_esp8266.c
```diff
@@ -0,0 +1,31 @@
...
+
+void ThreadAPI_Exit(int res)
+{
+	vTaskDelete(NULL);
+}
+
+void ThreadAPI_Sleep(unsigned int milliseconds)
+{
+	vTaskDelay(milliseconds);
+}
```
ritazh: done

adapters/tickcounter_esp8266.c
```diff
@@ -0,0 +1,87 @@
...
+typedef struct TICK_COUNTER_INSTANCE_TAG
+{
+    time_t init_time_value;
+    uint64_t current_ms;
+} TICK_COUNTER_INSTANCE;
+
+TICK_COUNTER_HANDLE tickcounter_create(void)
+{
+    TICK_COUNTER_INSTANCE* result = (TICK_COUNTER_INSTANCE*)malloc(sizeof(TICK_COUNTER_INSTANCE));
+    if (result != NULL)
```
ritazh: done

adapters/tlsio_ssl_esp8266.c
ritazh: I have only added `tls_io_instance->on_io_error(tls_io_instance->on_io_error_context);` to minimize memory footprint.

adapters/esp8266_mock.h
```diff
@@ -0,0 +1,430 @@
...
+//                struct timeval *timeout);
+
+
+#define htons(x) (x)
+#define ntohs(x) (x)
+#define htonl(x) (x)
+#define ntohl(x) (x)
+
+
+#endif
```
mamokarz: Please add a new line at the end of the file.

adapters/platform_esp8266.c
```diff
@@ -0,0 +1,35 @@
...
+#include "lwip/apps/sntp_time.h"
+
+int platform_init(void)
+{
+    sntp_init();
+    u32_t ts = 0;
+    while(ts == 0){
+        vTaskDelay(5000 / portTICK_RATE_MS);
+        ts = sntp_get_current_timestamp();
+        printf("%s\n", sntp_get_real_time(ts));
```
mamokarz: Please replace the printf by LogInfo

adapters/esp8266_mock.h
mamokarz: Please add a new line at the end of the file.

adapters/tlsio_ssl_esp8266.c
mamokarz: I just cannot see reason for that, once the result of one loop will not interfere in the other.

adapters/tlsio_ssl_esp8266.c
mamokarz: I see your point, but we cannot guarantee each one the upper layer will use, so, we need to call both.

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
ritazh: This is no result path for `tlsio_openssl_dowork` so it is not easy to test different states to cause it to fail. I have added `tlsio_openssl_dowork__failed`, to make sure that when it fails, it does not crash. But other than that, nothing interesting to test.

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
ritazh: Done. Added `tlsio_openssl_dowork__succeed_without_data` and `tlsio_openssl_dowork__succeed_with_data`

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
```diff
@@ -0,0 +1,733 @@
...
+        int result = 0;
+        TLS_IO_INSTANCE instance;
+        instance.tlsio_state = TLSIO_STATE_OPEN;
+        g_ssl_write_success = 1;
+
+        const IO_INTERFACE_DESCRIPTION* tlsioInterfaces = tlsio_openssl_get_interface_description();
+        ASSERT_IS_NOT_NULL(tlsioInterfaces);
+
+        ///act
+        result = tlsioInterfaces->concrete_io_send(&instance, NULL, 10, NULL, NULL);
```
ritazh: Done. Added `tlsio_openssl_send__SSL_write_null_buffer__failed`

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
ritazh: Done. Added `test_on_send_complete`

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
ritazh: Done. Added `test_on_io_close_complete`

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
ritazh: Done. Added `test_on_io_open_complete` and `test_on_io_error`

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
ritazh: Done. Added checks for `TLSIO_STATE_ERROR` and `TLSIO_STATE_NOT_OPEN` for final state.

adapters/esp8266_mock.h
ritazh: Done.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,693 @@
...
+    // LogInfo("get target IP is %d.%d.%d.%d", (unsigned char)((target_ip.addr & 0x000000ff) >> 0),
+    //                                             (unsigned char)((target_ip.addr & 0x0000ff00) >> 8),
+    //                                             (unsigned char)((target_ip.addr & 0x00ff0000) >> 16),
+    //                                             (unsigned char)((target_ip.addr & 0xff000000) >> 24));
+
+    //LogInfo("create socket ......");
+    sock = socket(AF_INET, SOCK_STREAM, 0);
+    if (sock < 0) {
+        LogError("create socket failed");
+        return;
```
ritazh: Done.

ritazh:
@mamokarz Thank you for your feedback. I have updated all the implementation as well as added unit tests per your request. Please take another look. Thanks!

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
```diff
@@ -0,0 +1,1028 @@
...
+    if (g_ssl_set_fd_success == 1){
+        return 1;
+    }else{
+        return 0;
+    }
+}
+
+SSL* my_SSL_new(SSL_CTX *ssl_ctx){
+    if (g_ssl_new_success == 1){
+         SSL* ssl = (SSL*)malloc(sizeof(SSL));
```
mamokarz: SSL is defined as 
```c
typedef void SSL;
```
So `sizeof(SSL)` is not allowed. It is the same for SSL_CTX.

adapters/tlsio_ssl_esp8266.c
ritazh: At the beginning of the second loop, you will see `flags` from the previous loop is used: `flags = fcntl(fd, F_SETFL, flags | O_NONBLOCK);`

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
```diff
@@ -0,0 +1,733 @@
...
+        STRICT_EXPECTED_CALL(bind(IGNORED_NUM_ARG,IGNORED_PTR_ARG, IGNORED_NUM_ARG)).IgnoreArgument(1).IgnoreArgument(2).IgnoreArgument(3);   
+        STRICT_EXPECTED_CALL(connect(IGNORED_NUM_ARG,IGNORED_PTR_ARG, IGNORED_NUM_ARG)).IgnoreArgument(1).IgnoreArgument(2).IgnoreArgument(3);
+        STRICT_EXPECTED_CALL(lwip_select(IGNORED_NUM_ARG, IGNORED_PTR_ARG, 
+             IGNORED_PTR_ARG, IGNORED_PTR_ARG, IGNORED_PTR_ARG)).IgnoreArgument(1).IgnoreArgument(2).IgnoreArgument(3).IgnoreArgument(4).IgnoreArgument(5);
+        STRICT_EXPECTED_CALL(TLSv1_client_method());
+        STRICT_EXPECTED_CALL(SSL_CTX_new(IGNORED_PTR_ARG)).IgnoreArgument(1);
+        STRICT_EXPECTED_CALL(SSL_set_fragment(IGNORED_PTR_ARG,IGNORED_NUM_ARG)).IgnoreArgument(1).IgnoreArgument(2);
+        STRICT_EXPECTED_CALL(SSL_new(IGNORED_PTR_ARG)).IgnoreArgument(1);
+        STRICT_EXPECTED_CALL(SSL_set_fd(IGNORED_PTR_ARG, IGNORED_NUM_ARG)).IgnoreArgument(1).IgnoreArgument(2);
+        STRICT_EXPECTED_CALL(SSL_connect(IGNORED_PTR_ARG)).IgnoreArgument(1);
```
ritazh: Done. Added `my_TLSv1_client_method` to unit test.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,724 @@
...
+#include "openssl/ssl_compat-1.0.h"
+#include "lwip/opt.h"
+#include "lwip/sockets.h"
+#include "lwip/sys.h"
+#include "espressif/esp8266/ets_sys.h"
+#include "espressif/espconn.h"
+
+#else
+//mock header
+#include "esp8266_mock.h"
```
ritazh: Okay.

ritazh:
@mamokarz Thanks for the feedback. Ready for another round of review. Will squash all the commits into one when ready. Thanks!

adapters/tlsio_ssl_esp8266.c
mamokarz: You are right.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,853 @@
...
+                                        //(void*)printf("SSL_set_fd ret:%d \n", ret);
+                                        if (ret != 1){
+                                            result = __LINE__;
+                                            LogError("SSL_set_fd failed");
+                                        }
+                                        else{
+                                            // LogInfo("SSL connect... ");
+                                            /* Codes_SRS_TLSIO_SSL_ESP8266_99_027: [ The tlsio_openssl_open shall set the tlsio to try to open the connection for 20 times before assuming that connection failed. ]*/
+                                            /* Codes_SRS_TLSIO_SSL_ESP8266_99_089: [ If SSL_connect failed, the tlsio_openssl_open shall return __LINE__. ] */
+                                            int retry = 0;
```
mamokarz: Please change the `retry` name for something else, you already have another variable with the same name in this context.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,853 @@
...
+        if (result == NULL)
+        {
+            /* Codes_SRS_TLSIO_SSL_ESP8266_99_012: [ If there is not enough memory to create the tlsio, the tlsio_openssl_create shall return NULL as the handle. ]*/
+            LogError("Failed allocating TLSIO instance.");
+        }
+        else
+        {
+            memset(result, 0, sizeof(TLS_IO_INSTANCE));
+            mallocAndStrcpy_s(&result->hostname, tls_io_config->hostname);
+            result->port = tls_io_config->port;
```
mamokarz: Please cast `tls_io_config->port` to `u16_t`

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,853 @@
...
+        {
+            /* Codes_SRS_TLSIO_SSL_ESP8266_99_059: [ If the tlsio state is not TLSIO_STATE_OPEN, the tlsio_openssl_send shall return _LINE_. ]*/
+            result = __LINE__;
+            LogError("TLS is not ready to send data. Expected state is TLSIO_STATE_OPEN.");
+            //(void)printf("TLS is not ready to send data. Expected state is TLSIO_STATE_OPEN.\n");
+        }
+        else
+        {
+            int total_write = 0;
+            int res = 0;
```
mamokarz: Please change `res` type to `size_t`

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
```diff
@@ -0,0 +1,1695 @@
...
+}
+#define  EINPROGRESS    115  /* Operation now in progress */
+int my_getsockopt(int s, int level, int optname, void *optval, socklen_t *optlen){
+    if (g_getsockopt_success == 1){
+        return EINPROGRESS;
+    }else{
+        return -1;
+    }
+}
+
```
mamokarz: Please create a implementation of the `my_fcntl`, something like:

```c
int my_fcntl(int parameterA, int parameterB, int parameterC)
{
    (void)(parameterA);
    (void)(parameterB);
    (void)(parameterC);
    return 0;
}
```

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
```diff
@@ -0,0 +1,1695 @@
...
+        REGISTER_GLOBAL_MOCK_HOOK(SSL_shutdown, my_SSL_shutdown);
+        REGISTER_GLOBAL_MOCK_HOOK(SSL_set_fd, my_SSL_set_fd);
+        REGISTER_GLOBAL_MOCK_HOOK(SSL_new, my_SSL_new);
+        REGISTER_GLOBAL_MOCK_HOOK(SSL_set_fragment, my_SSL_set_fragment);
+        REGISTER_GLOBAL_MOCK_HOOK(SSL_CTX_new, my_SSL_CTX_new);
+        REGISTER_GLOBAL_MOCK_HOOK(netconn_gethostbyname, my_netconn_gethostbyname);
+        REGISTER_GLOBAL_MOCK_HOOK(TLSv1_client_method, my_TLSv1_client_method);
+        REGISTER_GLOBAL_MOCK_HOOK(bind, my_bind);
+        REGISTER_GLOBAL_MOCK_HOOK(getsockopt, my_getsockopt);
+        REGISTER_GLOBAL_MOCK_HOOK(socket, my_socket);
```
mamokarz: Please hook the `fcntl` function, like

```c
REGISTER_GLOBAL_MOCK_HOOK(fcntl, my_fcntl);
```

ritazh: Done.

tests/tlsio_esp8266_ut/tlsio_esp8266_ut.c
ritazh: Done.

adapters/esp8266_mock.h
```diff
@@ -0,0 +1,446 @@
...
+
+//int bind(int s, const struct sockaddr* name, socklen_t namelen);
+MOCKABLE_FUNCTION(, int, bind, int, s, const struct sockaddr*, name, socklen_t, namelen);
+
+//int connect(int s, const struct sockaddr *name, socklen_t namelen);
+MOCKABLE_FUNCTION(, int, connect, int, s, const struct sockaddr*, name, socklen_t, namelen);
+
+//int lwip_getsockopt(int s, int level, int optname, void *optval, socklen_t *optlen)
+MOCKABLE_FUNCTION(, int, getsockopt, int, s, int, level, int, optname, void*, optval, socklen_t*, optlen);
+
```
ritazh: Done.

adapters/esp8266_mock.h
```diff
@@ -0,0 +1,446 @@
...
+/*
+ * SSL_CTX_new - create a SSL context
+ *
+ * @param method - the SSL context configuration file
+ *
+ * @return the context point, if create failed return NULL
+ */
+//SSL_CTX* SSL_CTX_new(SSL_METHOD *method);
+MOCKABLE_FUNCTION(, SSL_CTX*, SSL_CTX_new, SSL_METHOD*, method);
```
ritazh: Adding wrapper calls means more calling stack usage. Waiting to hear back from LCA. Unless LCA says we cannot mock these apis, we should not add wrapper calls.

ritazh:
@mamokarz @tameraw ready for another round of review.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,888 @@
...
+                            {
+                                if (!FD_ISSET(sock, &writeset) || FD_ISSET(sock, &errset)) {
+                                    result = __LINE__;
+                                    LogError("socket Error: %d", lwip_net_errno(sock));
+                                }else
+                                {
+                                    {
+                                        LogInfo("Socket Connect OK");
+                                        /* Codes_SRS_TLSIO_SSL_ESP8266_99_087: [ If SSL_new failed, the tlsio_openssl_open shall return __LINE__. ] */
+                                        ssl = SSL_new(ctx);
```
JetstreamRoySprowl: The ssl that is allocated here will leak if there are errors later on and the openssl_thread_LWIP_CONNECTIONcall fails.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,888 @@
...
+    SSL *ssl;
+
+    TLS_IO_INSTANCE* tls_io_instance = p;
+
+    LogInfo("OpenSSL thread start...");
+
+    {
+        LogInfo("create SSL context");
+        /* Codes_SRS_TLSIO_SSL_ESP8266_99_085: [ If SSL_CTX_new failed, the tlsio_openssl_open shall return __LINE__. ] */
+        ctx = SSL_CTX_new(TLSv1_client_method());
```
JetstreamRoySprowl: The ctx that is allocated here will leak if there are error later and openssl_thread_LWIP_CONNECTION fails.

adapters/tlsio_ssl_esp8266.c
ritazh: Thanks for the feedback @JetstreamRoySprowl 

In `destroy_openssl_instance`, I always free this if it is not NULL.

```
if (tls_io_instance->ssl != NULL)
        {
            SSL_free(tls_io_instance->ssl);
            tls_io_instance->ssl = NULL;
            LogInfo("SSL_free");
        }
```

adapters/tlsio_ssl_esp8266.c
ritazh: Same comment as `ssl`.

adapters/tlsio_ssl_esp8266.c
```diff
@@ -0,0 +1,888 @@
...
+          if (error != EINTR){
+              break;
+          }
+      } else{
+          break;
+      }
+  }
+
+  while(1){
+      flags = fcntl(fd, F_SETFL, flags | O_NONBLOCK);
```
JetstreamRoySprowl: If this fcntl call were to fail it would overwrite the value of flags with -1. Then when it retries it would try to set the flags to -1 instead of the value derived in the first loop. This bug is invisible for two reasons. One is that O_NONBLOCK is the only bit in flags that lwIP respects, so -1 happens to work.

The other reason is that when fcntl is used to set flags it is not capable of returning EINTR. If you look into the implementation for both lwIP and Linux you'll see that F_GETFL and F_SETFL are case statment labels that do simple bit masks and flips. So if a correct socket and flag value are supplied (which can be guaranteed before making the call) then there is  no way for the fcntl calls to fail, so this whole function is not needed.

adapters/tlsio_ssl_esp8266.c
JetstreamRoySprowl: I'm sorry, you are correct; I was looking at an older version of the file.

adapters/tlsio_ssl_esp8266.c
JetstreamRoySprowl: My bad; I was looking at an older version of the file.

adapters/tlsio_ssl_esp8266.c
ritazh: @JetstreamRoySprowl Thanks for the feedback. I have updated this block to only call fcntl F_SETFL once. `fcntl(fd, F_SETFL, O_NONBLOCK);`

PR #43, opened 2017-01-04
Required for clang on OSX

cwrenniks:


damonbarry:
The only change needed was to base this work on `develop` instead `master` so I just did a cherry-pick and ran it through Jenkins. Commited as 5dc2e0b. Closing this PR...

cwrenniks:
Ah, good call. I made the same mistake with the other PRs. Thanks Damon!

PR #44, opened 2017-01-12
Integrate with Azure IOT sdk with MICO board

lirenhe:
This is the init work that we make Azure IOT sdk work in MICO platform.
MICO is a company located in China who product MXChip board.
Here is the site http://mico.io/ (In Chinese)

Some note about this change:
1. MICO OS has customized makefile for those OS, so we could not use folder name in the 'including'
2. That system contains a header file of platform.h, so in order to avoid conflict, we renamed platform.h to platform_mico.h

### DIFF
--- a/adapters/agenttime_mico.c
+++ b/adapters/agenttime_mico.c
@@ -1,0 +1,36 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include <stdlib.h>
+#ifdef _CRTDBG_MAP_ALLOC
+#include <crtdbg.h>
+#endif
+#include <time.h>
+#include "mico.h"
+#include "gballoc.h"
+#include "agenttime.h"
+
+time_t get_time(time_t* currentTime)
+{
+    mico_utc_time_t utime;
+
+    mico_time_get_utc_time( &utime );
+    return (time_t)utime;
+}
+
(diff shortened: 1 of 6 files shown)


anporumb:
We are discussing internally whether we will take this in.

Best Regards,
Andrei Porumb

dcristoloveanu:
Hi @lirenhe,

We have reviewed this pull request. Generally the code looks OK. As a general comment the socket APIs seem not to have the return codes checked and that should definitely be addressed. In order to pull this into mainline we also require unittests and a requirements document to accompany the checked in modules.

We typically have one .md file with the requirements for the unit/module being developed and we also typically have a unit test suite covering those requirements.

Examples for the requirements docs are in the devdoc folder and examples of unit test suites are in the tests folder.
For testing we use umock_c in order to mock calls and allow writing true unit tests. There are plenty of examples in the repo in the tests folder.

We will also need a build script that will allow us to build this either on a Linux machine or on a Windows one. We prefer Linux since it allows for an easier build of a Docker container at the moment. This is needed in order that we do not break the code when we check in new bits.

Until these are resolved, in order to have this available for your customers I would like to suggest keeping this code in a fork that you make public to your customers.

Thanks,
/Dan

dcristoloveanu:
@lirenhe 

Closing, since we assume that you are keeping the code in a fork anyway.
If you would like to contribute this with tests and the build script feel free to reopen.

Thanks,
/Dan

PR #52, opened 2017-02-07
Enable CMake policy to use @rpath on Mac

damonbarry:
I hit a CMake warning while trying to compile the gateway SDK, which depends on azure-c-shared-utility. The warning recommended that I explicitly enable a Mac-specific policy that directs CMake to use `@rpath`.


anporumb:
LGTM

PR #53, opened 2017-02-07
Use CMake find_package() for curl

damonbarry:
When cross-compiling (and potentially in other scenarios too), the linker can't find curl because it isn't in the default location. This fix uses CMake's find_package() machinery to locate curl.


anporumb:
LGTM.

PR #57, opened 2017-02-24
Call on_send_complete in case errors occur

vjrantal:
Fixes Azure/azure-iot-sdk-c#73.

### DIFF
--- a/adapters/socketio_berkeley.c
+++ b/adapters/socketio_berkeley.c
@@ -524,6 +524,10 @@ int socketio_send(CONCRETE_IO_HANDLE socket_io, const void* buffer, size_t size,
         }
     }
 
+    if (on_send_complete != NULL && result != 0) {
+        on_send_complete(callback_context, IO_SEND_ERROR);
+    }
+
     return result;
 }
 


dcristoloveanu:
Hi @vjrantal 

Firstly I am sorry it took so long for me to get to review this change.
The _send functions of the XIO interface do not need to call the on_send_complete in case of errors. The error gets returned immediately and the layer that called send should deal with it right away.

There most likely is something else wrong for Azure/azure-iot-sdk-c#73. One of our devs is taking a look at that issue and we will have a fix for it soon.

Let me know if I missed anything.

vjrantal:
@dcristoloveanu I was basically looking at https://github.com/Azure/azure-c-shared-utility/blob/abf2d4b85fc4c5c9cec7aeb177163ab96c5fd129/adapters/tlsio_arduino.c#L377-L378 as reference. I wonder should the Arduino implementation be updated since that seems to be calling `on_send_complete` in case of errors? If I got it right, the right requirement would be to call `on_send_complete` only in case of success, but never in case of error?

vjrantal:
@dcristoloveanu Can you give your thoughts on the comment above? This relates to the ESP8266 port where currently, the `on_send_complete` is called in case of errors similarly than in the Arduino code ( see this line https://github.com/Azure/azure-c-shared-utility/pull/36/files#diff-8c46433401166533c0791c5eb3f9ac77R798 ). However, based on https://github.com/Azure/azure-c-shared-utility/pull/57#issuecomment-284505208 there is no need to call it. Which way would you recommend and should all adapters have the same behavior?

dcristoloveanu:
@vjrantal I am extremely sorry about the delay, this got unnoticed for quite a while becaus ethe PR was closed. Lesson learned.

The _send functions should return immediately with an error if the error is in the context of the _send call. In that case they should not trigger the callback as it is unnecessary. The caller knows by the result of send that there is no way the send will proceed so it should take any action needed when the send fails.

So any XIO adapters that do call send_complete in _send() are wrong and have to be fixed.

Also for quite some time now we wanted to put up a spec for XIO so that it is clear what the intent is. Hope to get to it soon.

Let me know if this makes sense or not.

Thanks,
/Dan

PR #65, opened 2017-03-23
Don't add VS unit test include path to target in VS2017

damonbarry:
This fixes a problem that prevents the gateway from building in VS 2017. The gateway has several unit test projects that use the older cpp-based micromock framework. In VS 2017, the directory being added to the `micromock_cpp_unittest` target doesn't exist, but the environment already has the *right* include path, so we can simply skip this CMake command.


damonbarry:
@andrew-buckley @Azure/iot-gateway-devs

PR #75, opened 2017-05-08
Fix error for building production code on VS2008

dcristoloveanu:
- Fix error for building production code on VS2008.
- Added stdint.h and inttypes.h with minimal code needed to satisfy building the libs.


tameraw:
:shipit:

PR #78, opened 2017-05-25
Build the SHARED lib as part of the regular build

dcristoloveanu:


CMakeLists.txt
```diff
@@ -481,11 +481,10 @@ ELSE()
...
 #this is the product (a library)
-if(${build_as_dynamic})
-    add_library(aziotsharedutil SHARED ./src/aziotsharedutil.def ${source_c_files} ${source_h_files})
-else()
-    add_library(aziotsharedutil ${source_c_files} ${source_h_files})
-endif()
+add_library(aziotsharedutil ${source_c_files} ${source_h_files})
+
+#make sure we can link as dll/so
+add_library(aziotsharedutil_dll SHARED ./src/aziotsharedutil.def ${source_c_files} ${source_h_files})
```
darobs: So, could we experiment with:
```
set_target_properties(aziotsharedutil_dll PROPERTIES OUTPUT_NAME "aziotsharedutil" )
```
to see if we get a `libaziotsharedutil.a` and `libaziotsharedutil.so`?

CMakeLists.txt
```diff
@@ -481,11 +481,10 @@ ELSE()
...
```
darobs: Also, could you add this new target to the INSTALL command below?

CMakeLists.txt
darobs: I made a simple test project, and this worked, although for this cmake file, if a user sets "BUILD_SHARED_LIBS=ON", it would build the same file twice, which doesn't seem to be a problem.

dcristoloveanu: Sure, done.

---
In reply to: [118510420](https://github.com/Azure/azure-c-shared-utility/pull/78#discussion_r118510420) [](ancestors = 118510420)

dcristoloveanu: Cool, thanks!

---
In reply to: [118518878](https://github.com/Azure/azure-c-shared-utility/pull/78#discussion_r118518878) [](ancestors = 118518878)

CMakeLists.txt
dcristoloveanu: Actually I am hesitant to do that.
I'd like to understand first if this dll is to be used or it should be built only as validation of the fact that it can be built.
If it is to be used, yes, I can add it to the install targets.

---
In reply to: [118512431](https://github.com/Azure/azure-c-shared-utility/pull/78#discussion_r118512431) [](ancestors = 118512431)

dcristoloveanu:
@darobs Can you have a look at the comments and let me know?
Thanks!
/Dan

CMakeLists.txt
dcristoloveanu: After in person discussion I added these to the install TARGETS

---
In reply to: [118546936](https://github.com/Azure/azure-c-shared-utility/pull/78#discussion_r118546936) [](ancestors = 118546936,118512431)

PR #89, opened 2017-07-07
have ETW logger

anporumb:
with -Duse_etw:BOOL=ON azure c shared utility will log to ETW instead of the console screen.

Additional steps needed, that can only happen after the application is build:
1) link in the application the resources of the azure-c-shared-utility (in the "res" folder)
1) register the provider with wevtutil utility


tameraw:
:shipit:

configs/azure_c_shared_utilityFunctions.cmake
```diff
@@ -484,6 +484,16 @@ function(set_platform_files c_shared_dir)
         if(${use_condition})
             set(CONDITION_C_FILE ${c_shared_dir}/adapters/condition_win32.c PARENT_SCOPE)
         endif()
+        if(NOT ${use_etw})
+            set(XLOGGING_C_FILE ${c_shared_dir}/src/xlogging.c PARENT_SCOPE)
+            set(LOGGING_C_FILE ${c_shared_dir}/src/consolelogger.c PARENT_SCOPE)
+            set(LOGGING_H_FILE ${c_shared_dir}/inc/azure_c_shared_utility/consolelogger.h PARENT_SCOPE)
```
jebrando: Need to be set in the Linux else block

src/etwlogger.man
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0"?>
```
jebrando: Should this file be under src?  Should it be under a new config.

src/etwlogger_driver.c
```diff
@@ -0,0 +1,262 @@
...
+            printf("failure in EventRegisterMicrosoft_ServiceBus");
+        }
+        else
+        {
+            /*should check if the provider is registered with the system... maybe at a later time*/
+            isETWLoggerInit = 1; /*and stays 1 until the process exits*/ /*sorry, no graceful exit with EventUnregisterMicrosoft_ServiceBus*/
+            LogInfo("EventRegisterMicrosoft_ServiceBus success"); /*selflogging that the log service has started*/
+        }
+    }
+
```
jebrando: Need to be in initialization.

src/etwlogger_driver.c
```diff
@@ -0,0 +1,262 @@
...
+void etwlogger_log(LOG_CATEGORY log_category, const char* file, const char* func, int line, unsigned int options, const char* format, ...)
+{
+    (void)options;
+    /*lazily init the logger*/
+    if (isETWLoggerInit == 0)
+    {
+        if (EventRegisterMicrosoft_ServiceBus() != ERROR_SUCCESS)
+        {
+            /*go back to printf, maybe somebody notices*/
+            printf("failure in EventRegisterMicrosoft_ServiceBus");
```
jebrando: void

src/etw_provider_generate.cmd
```diff
@@ -0,0 +1,11 @@
...
+
+REM this file generates the code needed for the ETW provider that acts as a logger
+
+REM following are hints about modifying the ETW provider
+REM 1) to change the provider, type ecmangen etwlogger.man in a Visual Studio developer command prompt.
+REM 2) to add the provider to the system, administrative privileges are needed. type "wevtutil im block_storage_etw.man /rf:"d:\etw01_folder\ETW01.exe" /mf:"d:\etw01_folder\ETW01.exe". It will not work without full path name.
+REM     a) errors might be due to user rights on the .exe, make sure that LocalService account has access right to the file.
+REM 3) Event Viewer will list the ETW provider under "Application and Service Logs -> Microsoft-ServiceBus"
+
+mc -um -h ../inc/azure_c_shared_utility -r ../res -v etwlogger.man
```
jebrando: >../res [](start = 43, length = 6)

Should the res files be checked in at all.  We can have a discussion about it at a later date.

jebrando:
:shipit:

jebrando:
After thinking about it the res files should not be checked in.  These are transient files and will change for each implementation and should only be stored locally.

configs/azure_c_shared_utilityFunctions.cmake
anporumb: done.

---
In reply to: [126213933](https://github.com/Azure/azure-c-shared-utility/pull/89#discussion_r126213933) [](ancestors = 126213933)

anporumb:
Thy will be done

---
In reply to: [313778465](https://github.com/Azure/azure-c-shared-utility/pull/89#issuecomment-313778465) [](ancestors = 313778465)

src/etwlogger.man
anporumb: I have obviously thought that it needs to stay under "src" since it gets compiled by "mc"... Not sure what is asked of me to do here... 

---
In reply to: [126215361](https://github.com/Azure/azure-c-shared-utility/pull/89#discussion_r126215361) [](ancestors = 126215361)

src/etwlogger_driver.c
anporumb: Awesome catch! I have put it in a static function.

---
In reply to: [126215732](https://github.com/Azure/azure-c-shared-utility/pull/89#discussion_r126215732) [](ancestors = 126215732)

src/etwlogger_driver.c
anporumb: ok :)

---
In reply to: [126215792](https://github.com/Azure/azure-c-shared-utility/pull/89#discussion_r126215792) [](ancestors = 126215792)

src/etw_provider_generate.cmd
anporumb: Turns out I can do without having the ../res folder checked in

---
In reply to: [126219049](https://github.com/Azure/azure-c-shared-utility/pull/89#discussion_r126219049) [](ancestors = 126219049)

PR #90, opened 2017-07-11
Add additional proxy options

markrad:
Add support for socks4, socks5 and socks5h web proxy protocols.

### DIFF
--- a/adapters/httpapi_curl.c
+++ b/adapters/httpapi_curl.c
@@ -732,59 +732,71 @@ HTTPAPI_RESULT HTTPAPI_SetOption(HTTP_HANDLE handle, const char* optionName, con
             char* proxy_auth;
 
             HTTP_PROXY_OPTIONS* proxy_data = (HTTP_PROXY_OPTIONS*)value;
-
-            if (sprintf_s(proxy, MAX_HOSTNAME_LEN, "%s:%d", proxy_data->host_address, proxy_data->port) <= 0)
-            {
-                LogError("failure constructing proxy address");
-                result = HTTPAPI_ERROR;
-            }
-            else
-            {
-                if (curl_easy_setopt(httpHandleData->curl, CURLOPT_PROXY, proxy) != CURLE_OK)
-                {
-                    LogError("failure setting curl proxy address");
-                    result = HTTPAPI_ERROR;
-                }
-                else
-                {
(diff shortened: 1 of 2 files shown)


adapters/httpapi_curl.c
```diff
@@ -732,59 +732,71 @@ HTTPAPI_RESULT HTTPAPI_SetOption(HTTP_HANDLE handle, const char* optionName, con
...
-                    else
-                    {
-                        result = HTTPAPI_OK;
-                    }
-                }
-            }
+			char *index = proxy;
+			index[0] = '\0';
+			
+			if (proxy_data->protocol != NULL && strcpy_s(proxy, MAX_HOSTNAME_LEN, proxy_data->protocol) != 0)
```
jebrando: >proxy_data->protocol [](start = 7, length = 20)

Be sure to check in upper layers of the code for non initialization of this variable.

jebrando:
:shipit:

adapters/httpapi_curl.c
markrad: This variable is initialized by the client code utilizing our SDK.

---
In reply to: [126739446](https://github.com/Azure/azure-c-shared-utility/pull/90#discussion_r126739446) [](ancestors = 126739446)

tameraw:
:shipit:

inc/azure_c_shared_utility/shared_util_options.h
```diff
@@ -15,6 +15,7 @@ extern "C"
         int port;
         const char* username;
         const char* password;
+		const char* protocol;
```
dcristoloveanu: >protocol [](start = 14, length = 8)

What values can be used here? It would be nice to have at least a comment with the possible values somewhere.

adapters/httpapi_curl.c
dcristoloveanu: >if (proxy_data [](start = 3, length = 14)

Since you are using anyhow sprintf_s, why not paste an empty string in case proxy_data->protocol is NULL and then do it all in one call rather than multiple copy and strlen calls?

dcristoloveanu: Also in that case index is not needed at all.

---
In reply to: [130945898](https://github.com/Azure/azure-c-shared-utility/pull/90#discussion_r130945898) [](ancestors = 130945898)

dcristoloveanu:
:shipit:

PR #96, opened 2017-08-15
Fix linux cross compile

JuergenKosel:
Hello,

this pull request fixes some small issues to cross build https://github.com/Azure/azure-iot-sdk-c .

Greetings
  Juergen

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -542,10 +542,16 @@ if(${use_http})
             set(aziotsharedutil_target_libs ${aziotsharedutil_target_libs} crypt32 winhttp)
         endif()
     else()
-        # try pkg-config first
-        find_package(PkgConfig)
-        if(PKG_CONFIG_FOUND)
+        if (CMAKE_CROSSCOMPILING)
+          # As mentioned at https://cmake.org/Wiki/CMake_Cross_Compiling the
+          # pkg-config tool can not be used by cmake while cross compiling.
+          message(STATUS "Cross compiling not using pkg-config")
+        else()
+          # try pkg-config first
+          find_package(PkgConfig)
+          if(PKG_CONFIG_FOUND)
             pkg_check_modules(CURL libcurl)
+          endif()
         endif()
 
         # if that didn't work, try CMake find_package


CMakeLists.txt
```diff
@@ -37,12 +37,17 @@ set(SHARED_UTIL_SRC_FOLDER ${CMAKE_CURRENT_LIST_DIR}/src CACHE INTERNAL "this is
...
-
+get_property(dirs DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES)
+foreach(dir ${dirs})
+  message(STATUS "dir='${dir}'")
+endforeach()
 CHECK_INCLUDE_FILES(stdint.h HAVE_STDINT_H)
 CHECK_INCLUDE_FILES(stdbool.h HAVE_STDBOOL_H)
 
-if ((NOT HAVE_STDINT_H) OR (NOT HAVE_STDBOOL_H))
+if (WINDOWS)
```
markrad: I'm interested to understand in what circumstances you have encountered issues with this code. I have cross compiled the SDK for many different targets and, other than one exception, cmake has always been able to find these two headers and hence did not execute the Windows CE circumvention. What platform are you seeing this as an issue?

JuergenKosel: I was cross compiling for a linux-armhf (Debian 8) device.
Due to bad linker flags, cmake claimed that it can not find stdint.h nor stdbool.h.
This had caused, that cmake accidentally adds the windowsce directory to the include path.
Which made it even more difficult, to see what is really going wrong.

markrad: Do you mean a bad include path? A linker flag shouldn't have any affect on the preprocessor's search for headers.

My issue with this is that you are getting blinkered by the fact that the directory happens to be called windowsce. Regardless of the OS that the build is targeting, if these two files are missing then they need to be provided by the SDK. Your modification will prevent that from happening except on a Windows platform.

By the way, your pull request still contains debug code unless you intended adding the list of include directories to the output. I prefer to use -v passed to gcc for that information.

JuergenKosel: There is a problem how cmake checks header files:
It first tries to compile with the `-c` option of gcc. If this succeeds, it tries to link.
And this second step failed, because cmake did not use the linker flags as expected.
(autoconf does it better: It first checks, that the compiler and provided flags works).

Maybe there is something wrong with the CMakeList.txt files of the super project, which uses this project as a sub-module: The linker flags are only set into the Makefiles, if cmake is called twice.

I will modify this pull request, by reducing to commit e661836.

markrad: This has caused me issues too. I wonder if using find-file would be a better option. I have a cross compile where the compiler checks always fail so I override them. The problem is that then cmake still tries to compile to verify header files and that always fails too even though the files are present.

tameraw:
@JuergenKosel - Thanks for the PR.

PR #117, opened 2017-10-04
Optimize singlylinkedlist to not have perf degradation when adding in an increasingly bigger list

dcristoloveanu:
Optimize singlylinkedlist to not have perf degradation when adding in an increasingly bigger list


dcristoloveanu:
The functionality has not changed, this is all internal optimization. However I took a look if any areas do not have coverage and added 1 test to fill that.

Thanks for the review.

PR #120, opened 2017-10-10
Add ECC support for curl in HTTP API

alexmrtn:
Hi,

We need ECC support in HTTP API for Linux but the adapter for curl does not implement the option to set ECC credentials.

We added the support of `OPTION_X509_ECC_KEY` and `OPTION_X509_ECC_CERT` options in  in `HTTPAPI_SetOption` method.

We also added a boolean flag `isECC` that let us know if we need to use the `x509_openssl_add_credentials` of `x509_openssl_add_ecc_credentials` function.

We checked that the unit tests pass.

### DIFF
--- a/adapters/httpapi_curl.c
+++ b/adapters/httpapi_curl.c
@@ -854,7 +854,7 @@ HTTPAPI_RESULT HTTPAPI_CloneOption(const char* optionName, const void* value, co
                 result = HTTPAPI_OK;
             }
         }
-        else if (strcmp(SU_OPTION_X509_CERT, optionName) == 0)
+        else if (strcmp(SU_OPTION_X509_CERT, optionName) == 0 || strcmp(OPTION_X509_ECC_CERT, optionName) == 0)
         {
             /*this is getting the x509 certificate. In this case, value is a pointer to a const char* that contains the certificate as a null terminated string*/
             if (mallocAndStrcpy_s((char**)savedValue, value) != 0)
@@ -868,7 +868,7 @@ HTTPAPI_RESULT HTTPAPI_CloneOption(const char* optionName, const void* value, co
                 result = HTTPAPI_OK;
             }
         }
-        else if (strcmp(SU_OPTION_X509_PRIVATE_KEY, optionName) == 0)
+        else if (strcmp(SU_OPTION_X509_PRIVATE_KEY, optionName) == 0 || strcmp(OPTION_X509_ECC_KEY, optionName) == 0)
         {
             /*this is getting the x509 private key. In this case, value is a pointer to a const char* that contains the private key as a null terminated string*/
             if (mallocAndStrcpy_s((char**)savedValue, value) != 0)
--- a/adapters/httpapi_curl.c
+++ b/adapters/httpapi_curl.c
@@ -36,6 +36,7 @@ typedef struct HTTP_HANDLE_DATA_TAG
     long verbose;
     const char* x509privatekey;
     const char* x509certificate;
+    bool isECC;
     const char* certificates; /*a list of CA certificates*/
 } HTTP_HANDLE_DATA;
 
@@ -218,7 +219,7 @@ static size_t ContentWriteFunction(void *ptr, size_t size, size_t nmemb, void *u
 static CURLcode ssl_ctx_callback(CURL *curl, void *ssl_ctx, void *userptr)
 {
     CURLcode result;
-    
+
     if (
         (curl == NULL) ||
         (ssl_ctx == NULL) ||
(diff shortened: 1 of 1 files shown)


jspaith:
The change here looks good and we'll be happy to merge this back to our master.

However when i pulled this down and ran an e2e test, it failed.  At minimum you'll need to modify httpapi_curl.c's HTTPAPI_CloneOption() since we make copies of this data but it doesn't know about the ECC options.

Good the UT's are passing, but for integration like this we still rely on e2e.  We don't have e2e client SDK for ECC (yet) but you can do some basics here:
* In IOT C client you'll need to force it to use ECC.  I've created https://github.com/Azure/azure-iot-sdk-c/tree/http_ecc_branch that has needful, it'll show where e2e tests are.
* The environment variables you'll need to set are 'IOTHUB_E2E_X509_CERT_BASE64' and 'IOTHUB_E2E_X509_PRIVATE_KEY_BASE64' which will be base 64 encoded versions of a self-signed ECC cert.  

Thanks @alexmrtn for taking this on.

alexmrtn:
I modified the HTTPAPI_CloneOption functions.

Unfortunately I was not able to run the e2e tests, even though I compiled the SDK without any modification:

- I followed the https://github.com/Azure/azure-iot-sdk-c/blob/master/doc/run_end_to_end_tests.md.
- I populated the correct variable with the correct values in the "iot_device_params.txt" file.
- I double checked my connection string and eventhub connection string.
- I used RSA certificates.

On all e2e tests I obtained the following error:

```
 1/10 Test  #1: iothubclient_uploadtoblob_e2e .........***Failed    2.04 sec
 === Executing test suite iothubclient_uploadtoblob_e2e ===
Info: Created Device csdk_e2eDevice_sas_j_please_delete_c5008aa8-1061-4c5a-af50-8ea41e004427.
Info: Created Device csdk_e2eDevice_x509_j_please_delete_830e0950-1963-4243-b820-4bd1176bb289.
Error: Time:Fri Oct 27 14:37:38 2017 File:/tmp/azure-iot-sdk-c/c-utility/adapters/httpapi_curl.c Func:HTTPAPI_ExecuteRequest Line:594 Failure in HTTP communication: server reply code is 400
Info: HTTP Response:{"Message":"ErrorCode:ArgumentInvalid;BadRequest","ExceptionMessage":"Tracking ID:01a4f3f5851f4426be09b127dc08e14b-G:4-TimeStamp:10/27/2017 12:37:38"}
Error: Time:Fri Oct 27 14:37:38 2017 File:/tmp/azure-iot-sdk-c/iothub_service_client/src/iothub_registrymanager.c Func:sendHttpRequestCRUD Line:1000 Http Failure status code 400.
Error: Time:Fri Oct 27 14:37:38 2017 File:/tmp/azure-iot-sdk-c/testtools/iothub_test/src/iothub_account.c Func:provisionDevice Line:348 IoTHubRegistryManager_CreateDevice failed
Error: Time:Fri Oct 27 14:37:38 2017 File:/tmp/azure-iot-sdk-c/testtools/iothub_test/src/iothub_account.c Func:IoTHubAccount_Init_With_Config Line:628 Failed to create the X509 device
  Assert failed in line 64: non-NULL expected. Could not initialize IoTHubAccount
TEST_SUITE_INITIALIZE failed - suite ending
```

Any idea where does this come from?
Can you check from your side if now the e2e tests pass?

Thanks @jspaith!

jspaith:
1 - I've incorporated your changes and run e2e and everything works.  There is a bug in attached clone_options change, namely there's two "||" which cause compile break.  I've fixed that bit up locally to unblock.

2 - Could you please fix the "||" issue and then squash the changes into 1?  (I'm assuming you can do that on a pull request...?)  That makes it easier to track history.  Once you do that, let me know and...

3 - I'lll get on checking it through our proper gates.

4 - For ECC instructions specifically I forgot a step, which is you need to do set environment variable 'IOTHUB_E2E_X509_THUMBPRINT' to the thumbprint of the certs that you're using.  This is the thumbprint that you set when creating the device in IoT Hub.

5 - For E2E not working, is it only upload blob or is it *everything*?  Uploadblob needs some extra server setup.  In any case, if it's universal I'd recommend opening a new github issue (against https://github.com/Azure/azure-iot-sdk-c/, not this repo) for tracking since we want customers to be able to run e2e obviously.

6 - Thanks and welcome to the (extended) Azure IoT team!

alexmrtn:
Hi @jspaith

1,2 - If fixed it, sorry for this stupid error, I squashed it into the previous commit.
3 - Let me know when it's ready.
4,5 - I previously set the `IOTHUB_E2E_X509_THUMBPRINT` variable, this error was for all e2e tests not only the upload blob. I'll check again everything and if needed make an issue on the right repo. 
6 Thanks!

jspaith:
I've just submitted this to master of cutility.  Sorry for delay and thanks again for this change.

Note that I have not yet updated the C SDK submodule to point to C-Utility, so a simple clone of CSDK won't pickup this change.  We update submodules fairly frequently though so this will be on the next train up.

PR #124, opened 2017-10-25
Extend DEC and INC macros to reach up to 2048

mredd:
The current limit of the `INC(x)` and `DEC(x)` macros was a little bit too low for our data model, so I had to extend them. Now, they reach up to `INC(2048)` and `DEC(2048)`.

The code was created with these two lines of bash code and then pasted into the header file:
```bash
for i in {2048..1024}; do echo "#define INC${i} $((i+1))"; done
for i in {2048..1025}; do echo "#define DEC${i} $((i-1))"; done
```


mredd:
Don't merge this pull request now. I just figured out that there is a whole lot more to do to allow for bigger models. I'm currently dealing with a model with 89 WITH_DATA statements, all of them are of struct types, and counting. Seriously Microsoft, when you go with all that preprocessor vodoo, why are you restricting yourself (or us) with so small models? Only 61 attributes allowed? What type of machine should that be to monitor?

I'll close this pull request and provide a new one once I've found all the macros to extend.

Issue #103, opened 2017-08-23
HTTPAPI_SetOption missing proxy option on Windows implementation


alexmrtn:
Hi,

I'm facing an issue using the HTTP API.
I need to use the SetOption method to configure a proxy :

`httpapi_res = HTTPAPI_SetOption(http_handle, OPTION_HTTP_PROXY, (void*)&proxy);`

This code works on Linux but not on Windows, the option seems not to be handled in the httpapi_winhttp.c implementation:

`unknown option proxy_data`

Thanks.

tameraw:
@alexmrtn - That's correct. Currently this option is not supported in the SDK on Windows. However, you should be able to specify the proxy settings in the OS directly and should get picked up by the SDK.

alexmrtn:
Hi @tameraw,

When I  set the proxy directly on windows, it is not working either. I get a:
`WinHttpSendRequest: (result = HTTPAPI_SEND_REQUEST_FAILED).`

Do I need to configure something to let the SDK get the OS proxy settings?
Is this option will be supported eventually? I need a portable way to set proxy for the HTTP API on both Linux and Windows.

Thanks!

tameraw:
@alexmrtn - One way to set it up would be thru the "Internet Options" control panel applet.
If you need this done programatically, would you be able to file a PR for this? Thanks.

alexmrtn:
Hi,

I created a PR (https://github.com/Azure/azure-c-shared-utility/pull/126) as the default behavior of the SDK is not to fetch the OS configuration. With this fix, the SDK actually pick up the OS proxy settings.

I'll create later a new PR to set the proxy option in a programatically way.

tameraw:
@alexmrtn thank you for your contribution to our open-sourced project!  Please help us improve by filling out this 2-minute [customer satisfaction survey](https://www.surveymonkey.com/r/Q5PLZ3V).

PR #126, opened 2017-10-27
change Proxy to default access in httpapi_winhttp.c

alexmrtn:
This PR is to change the default proxy configuration so that the SDK can retreive the system configuration on Windows for the HTTP API.

As I understood this should have been the normal behavior as @tameraw told me in a previous issue ( see https://github.com/Azure/azure-c-shared-utility/issues/103#issuecomment-325265256 )

I also aded in the SetOption the OPTION_HTTP_PROXY case, that has a similar behavior as the "TrustedCerts" option.

### DIFF
--- a/adapters/httpapi_winhttp.c
+++ b/adapters/httpapi_winhttp.c
@@ -130,7 +130,7 @@ HTTPAPI_RESULT HTTPAPI_Init(void)
     {
         if ((g_SessionHandle = WinHttpOpen(
             NULL,
-            WINHTTP_ACCESS_TYPE_DEFAULT_PROXY,
+            WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY,
             WINHTTP_NO_PROXY_NAME,
             WINHTTP_NO_PROXY_BYPASS,
             0)) == NULL)


adapters/httpapi_winhttp.c
```diff
@@ -736,6 +736,11 @@ HTTPAPI_RESULT HTTPAPI_SetOption(HTTP_HANDLE handle, const char* optionName, con
             /*winhttp accepts all certificates, because it actually relies on the system ones*/
             result = HTTPAPI_OK;
         }
+        else if (strcmp(OPTION_HTTP_PROXY, optionName) == 0)
+        {
+            /*winhttp accepts all proxy, because it actually relies on the system ones*/
+            result = HTTPAPI_OK;
+        }
```
tameraw: Please change this to error out and produce a log message stating that comment you have. The reason we need this to error out so that it is not misleading for users of the SDK thinking that the options they provided thru the SDK are in effect, while the options are set in OS.

tameraw:
@alexmrtn - Thanks for the PR. Looks good. Just please address the one comment in the review. Regards.

adapters/httpapi_winhttp.c
tameraw: Or simply remove the check and let it fall thru the next block.

---
In reply to: [150671536](https://github.com/Azure/azure-c-shared-utility/pull/126#discussion_r150671536) [](ancestors = 150671536)

alexmrtn:
Hi @tameraw  
I removed the error return, I squashed the two commit to keep a nice history.
Regards

ewertons:
I had to make a small change to your code, because that option you are using only exists on VS2010 and later.
https://msdn.microsoft.com/en-us/library/windows/desktop/aa384098(v=vs.85).aspx
https://stackoverflow.com/questions/70013/how-to-detect-if-im-compiling-code-with-visual-studio-2008

The code will be checked in tomorrow, there is a small build issue we need to fix before we proceed.
More updates to come once done.

ewertons:
This code is now checked in.

commit edab51e7d3ae6357c635104f7ce14a51e7256a66
Author: Ewerton Scaboro da Silva <ewertons@microsoft.com>
Date:   Tue Nov 14 17:05:05 2017 -0800

    Update to change 04a39360de7ba8d (automatic proxy selection on Windows)

commit 04a39360de7ba8d6fa545099c79bf534971cec76
Author: Alexis Martin <alexis.martin@schneider-electric.com>
Date:   Fri Oct 27 18:01:49 2017 +0200

    change Proxy to default access in httpapi_winhttp.c

ewertons:
It still needs to be made available at the Azure IoT SDK level, which will be done today.

I'll close this PR, please feel free to reopen it if needed.
Thanks for contributing to the Azure IoT SDK.

PR #129, opened 2017-12-06
Add ECC option support for winhttp

alexmrtn:
Hi,

This PR is designed to add ECC support for winhttp. It is similar to the one I made for curl (https://github.com/Azure/azure-c-shared-utility/pull/120)

The underlying function in `X509_schannel.c` handle the RSA and ECC certificates but the options were missing in the `SetOption` and `CloneOption` functions of winhttp implementation of HTTPAPI.

### DIFF
--- a/adapters/httpapi_winhttp.c
+++ b/adapters/httpapi_winhttp.c
@@ -691,7 +691,7 @@ HTTPAPI_RESULT HTTPAPI_SetOption(HTTP_HANDLE handle, const char* optionName, con
             httpHandleData->timeout = timeout;
             result = HTTPAPI_OK;
         }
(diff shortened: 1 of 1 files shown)


jspaith:
Thanks for another good change, @alexmrtn!  I've verified manually that works on Windows and have submitted to our C-Utility 4d729fb4cb41e933f64eb1efdb054f37646bb1b0.

I haven't updated C SDK submodule here yet as we typically have an update every few days free anyway.  

If there's any urgency here let us know and we should be able to get the submodule pointed to sooner.

alexmrtn:
Hi @jspaith !
There is no urgency (yet!), the sooner the better though :) 
Thanks

PR #132, opened 2018-01-12
add proxy support for httpapi_winhttp

alexmrtn:
Hi,

The goal of this PR is to add the support of proxy via SetOption to the winhttp implementation of httpapi.

I had to rollback on the changed I suggested in https://github.com/Azure/azure-c-shared-utility/pull/126, otherwise it is not possible to use the `WinHttpSetOption` function on the `requestHandle` (we get an `ERROR_WINHTTP_OPTION_NOT_SETTABLE` error).

I added the support of the `OPTION_HTTP_PROXY` in both `HTTPAPI_SetOption` and `HTTPAPI_CloneOption`.
Since proxy cannot be added on a connection, I added proxy settings in the `HTTP_HANDLE_DATA` and use it when creating a request handler in the `HTTPAPI_ExecuteRequest`

### DIFF
--- a/adapters/httpapi_winhttp.c
+++ b/adapters/httpapi_winhttp.c
@@ -33,6 +33,9 @@ typedef struct HTTP_HANDLE_DATA_TAG
     unsigned int timeout;
     const char* x509certificate;
     const char* x509privatekey;
+    const char* proxy_host;
+    const char* proxy_username;
+    const char* proxy_password;
 } HTTP_HANDLE_DATA;
 
 static HTTPAPI_STATE g_HTTPAPIState = HTTPAPI_NOT_INITIALIZED;
@@ -130,11 +133,7 @@ HTTPAPI_RESULT HTTPAPI_Init(void)
     {
         if ((g_SessionHandle = WinHttpOpen(
             NULL,
-#if defined _MSC_VER && _MSC_VER < 1600
             WINHTTP_ACCESS_TYPE_DEFAULT_PROXY,
-#else
-            WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY,
-#endif
             WINHTTP_NO_PROXY_NAME,
             WINHTTP_NO_PROXY_BYPASS,
             0)) == NULL)
@@ -239,6 +238,9 @@ HTTP_HANDLE HTTPAPI_CreateConnection(const char* hostName)
                             result->x509certificate = NULL;
                             result->x509privatekey = NULL;
                             result->x509SchannelHandle = NULL;
+                            result->proxy_host = NULL;
+                            result->proxy_username = NULL;
+                            result->proxy_password = NULL;
                         }
                     }
                     free(hostNameTemp);
@@ -267,6 +269,18 @@ void HTTPAPI_CloseConnection(HTTP_HANDLE handle)
                 (void)WinHttpCloseHandle(handleData->ConnectionHandle);
                 /*no x509 free because the options are owned by httpapiex.*/
                 handleData->ConnectionHandle = NULL;
+            }
+            if (handleData->proxy_host != NULL)
+            {
+                free((void*)handleData->proxy_host);
+            }
+            if (handleData->proxy_username != NULL)
+            {
+                free((void*)handleData->proxy_username);
+            }
+            if (handleData->proxy_password != NULL)
+            {
+                free((void*)handleData->proxy_password);
             }
             x509_schannel_destroy(handleData->x509SchannelHandle);
             free(handleData);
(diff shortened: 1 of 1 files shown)


adapters/httpapi_winhttp.c
```diff
@@ -740,6 +818,52 @@ HTTPAPI_RESULT HTTPAPI_SetOption(HTTP_HANDLE handle, const char* optionName, con
             /*winhttp accepts all certificates, because it actually relies on the system ones*/
             result = HTTPAPI_OK;
         }
+        else if (strcmp(OPTION_HTTP_PROXY, optionName) == 0)
+        {
+            char proxy[MAX_HOSTNAME_LEN];
+            HTTP_PROXY_OPTIONS* proxy_data = (HTTP_PROXY_OPTIONS*)value;
```
ewertons: You must check the fields within proxy_data for NULL values.

ewertons: Also since ->port is int, check for value greater than 0

adapters/httpapi_winhttp.c
```diff
@@ -740,6 +818,52 @@ HTTPAPI_RESULT HTTPAPI_SetOption(HTTP_HANDLE handle, const char* optionName, con
...
+                {
+                    LogError("failure allocate proxy host");
+                    result = HTTPAPI_ERROR;
+                }
+                else
+                {
+                    (void)strcpy(httpHandleData->proxy_host, proxy);
+                    if (proxy_data->username != NULL && proxy_data->password != NULL)
+                    {
+                        httpHandleData->proxy_username = (char*)malloc(strlen(proxy_data->username)+1);
```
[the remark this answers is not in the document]
ewertons: And check for the result of each operation step by step (regarding doing both malloc in sequence before checking the results)

adapters/httpapi_winhttp.c
```diff
@@ -740,6 +818,52 @@ HTTPAPI_RESULT HTTPAPI_SetOption(HTTP_HANDLE handle, const char* optionName, con
...
+                        httpHandleData->proxy_password = (char*)malloc(strlen(proxy_data->password)+1);
+                        if (httpHandleData->proxy_username == NULL || httpHandleData->proxy_password == NULL)
+                        {
+                            LogError("failure allocate proxy username or password");
```
ewertons: I won't leave more comments on this block, waiting for a revised code using mallocAndStrcpy_s

adapters/httpapi_winhttp.c
```diff
@@ -820,6 +944,49 @@ HTTPAPI_RESULT HTTPAPI_CloneOption(const char* optionName, const void* value, co
...
+                result = HTTPAPI_ERROR;
+            }
+            else
+            {
+                HTTP_PROXY_OPTIONS *savedOption = (HTTP_PROXY_OPTIONS*) *savedValue;
+                HTTP_PROXY_OPTIONS *option = (HTTP_PROXY_OPTIONS*) value;
+                if (mallocAndStrcpy_s((char**)(savedOption->host_address),
+                                      (const char*)option->host_address) != 0)
+                {
+                  LogError("unable to clone the proxy host adress content");
```
ewertons: If any failures occur you need to release any memory allocated within this function and NULL the *savedValue variable

ewertons: This applies to each block bellow as well.

adapters/httpapi_winhttp.c
```diff
@@ -820,6 +944,49 @@ HTTPAPI_RESULT HTTPAPI_CloneOption(const char* optionName, const void* value, co
...
+                  result = HTTPAPI_ERROR;
+                }
+                else
+                {
+                    savedOption->port = option->port;
+                    if (option->username != NULL || mallocAndStrcpy_s((char**)savedOption->username,
```
ewertons: Did you mean to use && instead of || ?

adapters/httpapi_winhttp.c
```diff
@@ -820,6 +944,49 @@ HTTPAPI_RESULT HTTPAPI_CloneOption(const char* optionName, const void* value, co
...
+                                                                      (const char*)option->username) != 0)
+                    {
+                      LogError("unable to clone the proxy username content");
+                      result = HTTPAPI_ERROR;
+                    }
+                    else
+                    {
+                        if (option->password != NULL || mallocAndStrcpy_s((char**)savedOption->password,
```
ewertons: Did you mean to use && instead of || ?

alexmrtn:
Hi @ewertons 

Thanks for the quick-and-well-detailed review!

I revert all changes related to spaces (my text editor was automatically cleaning spaces on empty lines).
I also corrected my mistakes, added test for mbstowcs, added the free where they need to be (basically I followed all your review ;) )

I added the `#if defined _MSC_VER && _MSC_VER >= 1600` check to include `#include <VersionHelpers.h>`. And I used `#ifndef WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY` as you suggested.

So here the implemented behavior depending platform and compiler version: 

| Compiler Version | Runtime | includes versionhelper.h ? | uses automatic proxy ? |
|:----------------:|:-------:|:--------------------------:|:----------------------:|
|       <1600      | <Win8.1 |             NO             |           NO           |
|       1600+      | <Win8.1 |             YES            |           NO           |
|       <1600      | Win8.1+ |             NO             |           NO           |
|       1600+      | Win8.1+ |             YES            |           YES          |

Can you confirm this is what you expect?

Thanks,
Alexis.

ewertons:
Hi @alexmrtn ,
thank you so much for the changes, I will review them again. Let's get this in as soon as possible.
I have one concern about the checks above (thanks for the table, that's pretty good). _MSC_VER refers to the version of the compiler (so it's related to the VS version installed) while VersionHelpers.h is part of the Windows API (related to the operating system version), so I understand you should not use _MSC_VER in this case. In one of my comments I mention another macro, MSVER, which would be the correct one. Take a look on the macro definitions and let us know what you think.

alexmrtn:
Hi @ewertons 

I'm a bit confused about this version checking so let's recap all of this:
- **Proxy option**, must be checked at **runtime** ([see](https://msdn.microsoft.com/en-us/library/windows/desktop/aa384098(v=vs.85).aspx)):
  - We use `WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY` option for Win 8.1+
  - We use `WINHTTP_ACCESS_TYPE_DEFAULT_PROXY` option for <Win8.1
  - We use `IsWindows8Point1OrGreater()` function defined in `versionhelpers.h` to determine which option to use.
  - I think this part is OK.
- **versionhelpers.h** is only available on **Win8.1 SDK** ([see](https://msdn.microsoft.com/en-us/library/windows/desktop/dn424972(v=vs.85).aspx))
  - From my understanding, we need to check the **SDK version** and neither Windows version on which we compile, nor the version of the compiler we use. 
  - The `#if` preprocessor for the include  `versionhelpers.h` and the one that defines a block that uses `IsWindows8Point1OrGreater()` must be the same, otherwise `IsWindows8Point1OrGreater()` is undefined.
  - We use `WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY` macro to determine if SDK is Win 8.1+.

Questions:
- You suggested to use the macro `MSVER` instead of `_MSC_VER`, but did you mean `WINVER` as you mentioned in your previous comment?
- So I guess we need to use `#if defined WINVER && WINVER < 0x0602` for both `#if`?
- Does `WINVER` really reflect the SDK version or does it reflect the windows version on which we compile? It seems that we can use SDK Win8.1 on Win7 ([see](https://stackoverflow.com/questions/30085672/using-windows-sdk-8-1-for-development-on-windows-7)) I just want to make sure we implement the correct behavior :)
- Where are those macros defined `WINVER`/`MSVER` ?
- Otherwise, can't  we use `WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY` for both `"if`?

Thanks!

alexmrtn:
Hi @ewertons 

I did few tests and is seems that our support to the automatic proxy is more complex than what I proposed:
If we use the `WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY` when we create the *session handler*, we cannot use the function `WinHttpSetOption` with the option `WINHTTP_OPTION_PROXY` on the *request handler* that depend on this session (we obtain an error code  `12011       ERROR_INTERNET_OPTION_NOT_SETTABLE: The request option cannot be set, only queried`).

As **both** *set proxy option* and *automatic proxy* are exclusive, it would require a more complex mechanism involving the elements we discussed previously (windows version, compiler version) **plus** this a new parameter to support all cases. Basically it would require to erase and recreate the `g_SessionHandle` that is initialliy created in the `HTTPAPI_Init()` function. However, this *session handle* is common for all the application that uses the `HTTPAPI` and we do not want to beak all other connections already created. 

As I do not know how you want to support the automatic proxy regarding these problems,
I changed my PR to use only `WINHTTP_ACCESS_TYPE_NO_PROXY` instead of `WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY` and yet it allows a user to manually set the proxy, thus leveraging the issue on windows 7 currently present. 

I would suggest we do the automatic proxy support in another PR, as this one is already complex and these are two different features.

What do you thinks?

Thanks,
Alexis.

adapters/httpapi_winhttp.c
```diff
@@ -740,6 +826,65 @@ HTTPAPI_RESULT HTTPAPI_SetOption(HTTP_HANDLE handle, const char* optionName, con
...
+            else
+            {
+                if (sprintf_s(proxyAddressAndPort, MAX_HOSTNAME_LEN, "%s:%d", proxy_data->host_address, proxy_data->port) <= 0)
+                {
+                    LogError("failure constructing proxy address");
+                    result = HTTPAPI_ERROR;
+                }
+                else
+                {
+                    if(mallocAndStrcpy_s((char**)&(httpHandleData->proxy_host), (const char*)proxyAddressAndPort) != 0)
```
ewertons: This leads to a possible memory leak if the option was previously set.
Check if httpHandleData->proxy_host is not NULL, then free and NULL it before calling mallocAndStrcpy_s.

adapters/httpapi_winhttp.c
```diff
@@ -740,6 +826,65 @@ HTTPAPI_RESULT HTTPAPI_SetOption(HTTP_HANDLE handle, const char* optionName, con
...
+                    {
+                        LogError("failure allocate proxy host");
+                        result = HTTPAPI_ERROR;
+                    }
+                    else
+                    {
+                        if (proxy_data->username != NULL && proxy_data->password != NULL)
+                        {
+                            if(mallocAndStrcpy_s((char**)&(httpHandleData->proxy_username), (const char*)proxy_data->username) != 0)
```
ewertons: This leads to a possible memory leak if the option was previously set.
Check if httpHandleData->proxy_username is not NULL, then free and NULL it before calling mallocAndStrcpy_s.

adapters/httpapi_winhttp.c
```diff
@@ -740,6 +826,65 @@ HTTPAPI_RESULT HTTPAPI_SetOption(HTTP_HANDLE handle, const char* optionName, con
...
+                            {
+                                LogError("failure allocate proxy username");
+                                free((void*)httpHandleData->proxy_host);
+                                httpHandleData->proxy_host = NULL;
+                                result = HTTPAPI_ERROR;
+                            }
+                            else
+                            {
+                                if(mallocAndStrcpy_s((char**)&(httpHandleData->proxy_password), (const char*)proxy_data->password) != 0)
```
ewertons: This leads to a possible memory leak if the option was previously set.
Check if httpHandleData->proxy_password is not NULL, then free and NULL it before calling mallocAndStrcpy_s.

adapters/httpapi_winhttp.c
```diff
@@ -740,6 +826,65 @@ HTTPAPI_RESULT HTTPAPI_SetOption(HTTP_HANDLE handle, const char* optionName, con
...
+                                {
+                                    LogError("failure allocate proxy username");
```
ewertons: Typo. 
Change "username" to "password"

adapters/httpapi_winhttp.c
```diff
@@ -268,6 +270,21 @@ void HTTPAPI_CloseConnection(HTTP_HANDLE handle)
                 /*no x509 free because the options are owned by httpapiex.*/
                 handleData->ConnectionHandle = NULL;
             }
+            if (handleData->proxy_host != NULL)
+            {
+                free((void*)handleData->proxy_host);
+                handleData->proxy_host = NULL;
```
ewertons: No need to NULL if it's freeing "handleData" bellow.

adapters/httpapi_winhttp.c
```diff
@@ -268,6 +270,21 @@ void HTTPAPI_CloseConnection(HTTP_HANDLE handle)
...
+            }
+            if (handleData->proxy_username != NULL)
+            {
+                free((void*)handleData->proxy_username);
+                handleData->proxy_username = NULL;
```
ewertons: No need to NULL if it's freeing "handleData" bellow.

adapters/httpapi_winhttp.c
```diff
@@ -268,6 +270,21 @@ void HTTPAPI_CloseConnection(HTTP_HANDLE handle)
...
+            }
+            if (handleData->proxy_password != NULL)
+            {
+                free((void*)handleData->proxy_password);
+                handleData->proxy_password = NULL;
```
ewertons: No need to NULL if it's freeing "handleData" bellow.

adapters/httpapi_winhttp.c
```diff
@@ -130,11 +133,7 @@ HTTPAPI_RESULT HTTPAPI_Init(void)
     {
         if ((g_SessionHandle = WinHttpOpen(
             NULL,
-#if defined _MSC_VER && _MSC_VER < 1600
-            WINHTTP_ACCESS_TYPE_DEFAULT_PROXY,
-#else
-            WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY,
-#endif
+            WINHTTP_ACCESS_TYPE_NO_PROXY,
```
ewertons: This concerns me a little. 
We are going from an automatic mode to no proxy, I need to check if that will break anyone.
I'll post more briefly about it

ewertons:
In reply to "_I did few tests and is seems that our support to the automatic proxy is more complex than what I proposed..._":

I agree, we should separate the issues. Thank you so much for noting that.

However changing the proxy mode to no proxy seems risky to me at this point, let me get more details and I'll reply back.

adapters/httpapi_winhttp.c
ewertons: Our original value here (when the API was written) was WINHTTP_ACCESS_TYPE_DEFAULT_PROXY.
Alexis changed the previous value from *DEFAULT* to WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY.

```
commit 04a39360de7ba8d6fa545099c79bf534971cec76
Author: Alexis Martin <alexis.martin@schneider-electric.com>
Date:   Fri Oct 27 18:01:49 2017 +0200

    change Proxy to default access in httpapi_winhttp.c
```

1. Will your change work for Alexis? We don't want to disrupt his project.
2. Instead of *NO_PROXY*, could we revert it back to *DEFAULT* until we figure the solution for the multi-OS/SDK-version issue? Would that work for you?

ewertons:
So upon discussing here, we believe you should not use the NO_PROXY option as the default value. 
This would be the block to be used for now (until we change it again for the version discussion):

```
#if defined WINVER && WINVER < 0x0602
             WINHTTP_ACCESS_TYPE_DEFAULT_PROXY,
#else
             WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY,
#endif
``` 

Please adjust the indentation as appropriate.

ewertons:
@alexmrtn, 
we are planning on releasing a new version this Wednesday. 
If you get the changes addressed before that we can take them in on this release. Would that work for you?

alexmrtn:
Hi @ewertons 

Thanks for the review, I corrected the problems related to memory leaks and typo. However I still have a problem reintroducing the the block:
```
#if defined WINVER && WINVER < 0x0602
             WINHTTP_ACCESS_TYPE_DEFAULT_PROXY,
#else
             WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY,
#endif
```

On Windows 8.1+ the `WinHttpOpen` function takes the `WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY` mode, and when we use the set option we get the error:
```
[09:46:09.698][ERROR][AZURE_IOTHUB]                     failure setting proxy address (12011)
```

Indeed, using the `WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY` prevent us to set a manual proxy (err 12011 is "The request option cannot be set, only queried.").
Using the `WINHTTP_ACCESS_TYPE_DEFAULT_PROXY` option does not trigger the error. What I suggest is to revert to this original value (when the API was written) before my PR adding this automatic proxy mode. 

The if/else block you suggested can be then added back in a next PR that will solve the automatic proxy problem. I also think we should add the same mechanism in transport protocols for the connection to the iotHub in order to have some consistency in the SDK. But this will be addressed in another thread.

It would be awesome if this changes are merged into the next release! If you still have comments I will do the last modifications during my CET-Wednesday, this will be available for your PST-Wednesday!

Thanks again,
Alexis.

adapters/httpapi_winhttp.c
```diff
@@ -820,6 +977,61 @@ HTTPAPI_RESULT HTTPAPI_CloneOption(const char* optionName, const void* value, co
...
+                {
+                  LogError("unable to clone the proxy host adress content");
+                  free((void*)*savedValue);
+                  *savedValue = NULL;
+                  result = HTTPAPI_ERROR;
+                }
+                else
+                {
+                    savedOption->port = option->port;
+                    if (option->username != NULL && mallocAndStrcpy_s((char**)&(savedOption->username),
```
ewertons: Oops, found a bug here. Sorry.

At lines 1003 and 1015 savedOption->username and savedOption->password are only set if option->username or option->password are not NULL, respectively.
So if any is NULL the respective value in savedOption is not set.
However, savedOption values as never NULLed initially, so they will have invalid pointer addresses (gargabe) set in this case, which later can cause a memory access violation.

Could you please add this call between lines 991 and 992?

memset(savedValue, 0, sizeof(HTTP_PROXY_OPTIONS));

adapters/httpapi_winhttp.c
```diff
@@ -130,11 +133,7 @@ HTTPAPI_RESULT HTTPAPI_Init(void)
     {
         if ((g_SessionHandle = WinHttpOpen(
             NULL,
-#if defined _MSC_VER && _MSC_VER < 1600
             WINHTTP_ACCESS_TYPE_DEFAULT_PROXY,
```
ewertons: Agreed. We will revisit it later.

adapters/httpapi_winhttp.c
alexmrtn: Oops, missed that too! Fixed.

ewertons: Just an update, I had a typo on the command above.
I fixed it myself to be:
memset((void*)*savedValue, 0, sizeof(HTTP_PROXY_OPTIONS));

ewertons:
@alexmrtn , I sent you an email with some questions, please let me know if you got it. Thanks.

PR #135, opened 2018-01-25
Proxy issues with Windows 7

PeterSmithson:
Hi,

I work for Schneider-electric in the UK as a developer.  We have submitted something similar to this via e-mail but nothing came of it.  I'm now using github for the first time so I hope I'm doing this right.

We had some issues with our EcoStuxureware product in an office running Windows 7 with a PAC file for the proxy settings.  These changes resolved that issue.  I've also checked it works on Windows 10 with a PAC file and also without any proxy (in a 3rd office).

I noticed a branch called "EcoStruxure-SE/automatic_proxy_windows" which seems related by @alexmrtn 

There are two problems. 
1) If the code is compiled with a Windows 10 SDK on a recent compiler, the code will use WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY.  When the compiled code is ran on a Windows 7 PC, the WinHttpOpen will fail as that option is not available on that OS. The fix will allow one compiled binary to be used on Windows 7 or later.  First (assuming you are compiling on up to date system), it will try the AUTOMATIC option.  If that fails, it falls back to DEFAULT.
2) In the DEFAULT mode, the proxy settings are required. We found this on a site where a PAC script was used - we don't know if it would apply to other situations.  A call to WinHttpSetOption has been added.

If the binary is ran on Windows 10 (and presumably 8.1), the AUTOMATIC option is used and everything works fine (we have found).

### DIFF
--- a/adapters/httpapi_winhttp.c
+++ b/adapters/httpapi_winhttp.c
@@ -756,8 +756,8 @@ HTTPAPI_RESULT HTTPAPI_SetOption(HTTP_HANDLE handle, const char* optionName, con
         else if (strcmp(OPTION_HTTP_PROXY, optionName) == 0)
         {
             /*Can't set proxy if using automatic*/
-	    if (g_ProxyAutomatic == 0)
-	    {
+            if (g_ProxyAutomatic == 0)
+            {
                 char proxy_address[MAX_HOSTNAME_LEN];
                 wchar_t wproxy_address[MAX_HOSTNAME_LEN];
 
@@ -768,9 +768,9 @@ HTTPAPI_RESULT HTTPAPI_SetOption(HTTP_HANDLE handle, const char* optionName, con
                     LogError("failure constructing proxy address");
                     result = HTTPAPI_ERROR;
                 }
-		else
-		{
-                    MultiByteToWideChar( CP_UTF8 , 0 , proxy_address , -1, wproxy_address , MAX_HOSTNAME_LEN);
+                else
+                {
+                    MultiByteToWideChar(CP_UTF8, 0, proxy_address, -1, wproxy_address, MAX_HOSTNAME_LEN);
 
                     WINHTTP_PROXY_INFO proxy = { 0 };
                     proxy.dwAccessType = WINHTTP_ACCESS_TYPE_NAMED_PROXY;
@@ -782,13 +782,13 @@ HTTPAPI_RESULT HTTPAPI_SetOption(HTTP_HANDLE handle, const char* optionName, con
                         result = HTTPAPI_ERROR;
                     }
                     else
-                      result = HTTPAPI_OK;
-		}
-	    }
-	    else
-	    {
-              result = HTTPAPI_OK;
-	    }
+                        result = HTTPAPI_OK;
+                }
+            }
+            else
+            {
+                result = HTTPAPI_OK;
+            }
         }
         else
         {
--- a/adapters/httpapi_winhttp.c
+++ b/adapters/httpapi_winhttp.c
@@ -36,6 +36,7 @@ typedef struct HTTP_HANDLE_DATA_TAG
 } HTTP_HANDLE_DATA;
 
 static HTTPAPI_STATE g_HTTPAPIState = HTTPAPI_NOT_INITIALIZED;
+static int g_ProxyAutomatic = 1;
 
 /*There's a global SessionHandle for all the connections*/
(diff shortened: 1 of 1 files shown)


PeterSmithson:
I just realised that this conflicts with https://github.com/Azure/azure-c-shared-utility/pull/132 which also deals with the proxy.  It doesn't use the AUTOMATIC option at all though.

ewertons:
Hi @PeterSmithson ,
thanks for the PR, however based on recent verification of the code the issue was root caused and the possible solution will need to be different than the proposed above.
Here is the ticket tracking this issue: https://github.com/Azure/azure-c-shared-utility/issues/136

adapters/httpapi_winhttp.c
```diff
@@ -128,26 +129,37 @@ HTTPAPI_RESULT HTTPAPI_Init(void)
 
     if (nUsersOfHTTPAPI == 0)
     {
-        if ((g_SessionHandle = WinHttpOpen(
-            NULL,
-#if defined _MSC_VER && _MSC_VER < 1600
-            WINHTTP_ACCESS_TYPE_DEFAULT_PROXY,
-#else
-            WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY,
+        g_SessionHandle = NULL;
```
ewertons: No pre-assignment of variables. Please add it after a #else

adapters/httpapi_winhttp.c
```diff
@@ -128,26 +129,37 @@ HTTPAPI_RESULT HTTPAPI_Init(void)
...
-            WINHTTP_NO_PROXY_NAME,
-            WINHTTP_NO_PROXY_BYPASS,
-            0)) == NULL)
+        if (g_SessionHandle == NULL)
         {
-            LogErrorWinHTTPWithGetLastErrorAsString("WinHttpOpen failed.");
-            result = HTTPAPI_INIT_FAILED;
+	    g_ProxyAutomatic = 0;
+            /*Automatic option is not supported on Windows 7 or earlier so try older option*/
+            g_SessionHandle = WinHttpOpen(
```
ewertons: Indeed please move this whole block to within the #else

adapters/httpapi_winhttp.c
```diff
@@ -128,26 +129,37 @@ HTTPAPI_RESULT HTTPAPI_Init(void)
...
+                NULL,
+                WINHTTP_ACCESS_TYPE_DEFAULT_PROXY,
+                WINHTTP_NO_PROXY_NAME,
+                WINHTTP_NO_PROXY_BYPASS,
+                0);
         }
-        else
+        if (g_SessionHandle)
```
ewertons: Please revert this if block order change to minimize the code changes.

adapters/httpapi_winhttp.c
```diff
@@ -740,6 +753,43 @@ HTTPAPI_RESULT HTTPAPI_SetOption(HTTP_HANDLE handle, const char* optionName, con
...
+                wchar_t wproxy_address[MAX_HOSTNAME_LEN];
+
+                HTTP_PROXY_OPTIONS* proxy_data = (HTTP_PROXY_OPTIONS*)value;
+
+                if (sprintf_s(proxy_address, MAX_HOSTNAME_LEN, "%s:%d", proxy_data->host_address, proxy_data->port) <= 0)
+                {
+                    LogError("failure constructing proxy address");
+                    result = HTTPAPI_ERROR;
+                }
+		else
```
ewertons: Please fix the identation

adapters/httpapi_winhttp.c
PeterSmithson: If this was in the #else, that would cause a problem.  I want this code to be executed even if this code is compiled on a system where AUTOMATIC is available because it might be ran on a system (like Win 7) where automatic is not available.  It's confusing as I'm trying to allow for both the system it's compiled on and also the system it runs on.  Perhaps I need to do a more major change to make it clearer?

adapters/httpapi_winhttp.c
PeterSmithson: I don't think I can do that without breaking my fix.  I have tried to do minimal changes.  If I can re-write this code, perhaps I can make it clearer.

PeterSmithson:
This is covered by PR #132 so I'm closing it.

ewertons:
Alright, we will focus on #132. Thanks @PeterSmithson

PR #137, opened 2018-02-13
Adds the shared library to the install and versions the .so file

markrad:
Now the .so or .dll file is built correctly when build_as_dynamic is passed to cmake, this fix ensures that the .so file is installed and versioned and the .dll file uses a cmake facility to remove the requirement to keep the .def file up to date. It also uses a obj build so that the code is only compiled once and then used to build the .so/.dll and .lib files.


markrad:
@ewertons is added to the review. #Closed

ewertons:
If possible please run a azure-iot-sdk-c buddy build on the gate using this before checking it into c-shared.

jspaith:
Per offline discussion with Mark, closing for now until code is completely ready.

PR #141, opened 2018-03-08
Copy hostname and port from socketio to tlsio config

### LINKED ISSUE #139: uws_client_create throws Access violation reading location.
Repro (win32):
```
static const WS_PROTOCOL protocols[] = { { "test_protocol" } };
uws_client_create("test_host", 444, "/aaa", true, protocols, sizeof(protocols) / sizeof(protocols[0]));
```

looks like the hostname is not copied into tlsio_config.

The exception is raised in tlsio_schannel.c:995.

raaaar:
Fixes #139

### DIFF
--- a/src/uws_client.c
+++ b/src/uws_client.c
@@ -202,6 +202,8 @@ UWS_CLIENT_HANDLE uws_client_create(const char* hostname, unsigned int port, con
                                     socketio_config.port = port;
                                     socketio_config.accepted_socket = NULL;
 
+                                    tlsio_config.hostname = hostname;
+                                    tlsio_config.port = port;
                                     tlsio_config.underlying_io_interface = socketio_get_interface_description();
                                     tlsio_config.underlying_io_parameters = &socketio_config;
 


src/uws_client.c
```diff
@@ -202,6 +202,8 @@ UWS_CLIENT_HANDLE uws_client_create(const char* hostname, unsigned int port, con
                                     socketio_config.port = port;
                                     socketio_config.accepted_socket = NULL;
 
+                                    tlsio_config.hostname = hostname;
```
ewertons: Although it's understandable that the char pointer saved here would prevent more memory allocation, there is a risk here of bad memory management happening. 
I would recommending following the same practice within this function of copying the value (see mallocAndStrcpy_s above) instead of passing the pointer.

raaaar: I'm not entirely convinced this is actually the practice in this function.
A copy of the hostname is created on L148 and stored in the result structure.
However, whenever hostname is used as a part of a config to invoke xio_create, it's passed as is without creating an extra copy. See lines 201 and 233.

ewertons: Fine, we will fix it later.

raaaar:
ping

PR #142, opened 2018-03-09
Fix HTTP_API proxy credentials

alexmrtn:
Hi,

It seems that the current implementation of the proxy support for HTTP_API contains a bug. When sending the request to a proxy server that requests user name and password, the expected http header `Proxy-Authorization` ([see](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Proxy-Authorization)) is missing:

**Problem**: Using `WinHttpSetOption` with the option flag `WINHTTP_OPTION_PROXY_USERNAME` or `WINHTTP_OPTION_PROXY_PASSWORD` seems not  to have any effect on the corresponding `HINTERNET` request.

**Solution**: Using `WinHttpSetCredentials` function instead of `WinHttpSetOption` fixes the missing header in the http request and is recommended ([see](https://msdn.microsoft.com/en-us/library/windows/desktop/aa384112(v=vs.85).aspx)).

I also changed the previous use of `mbstowcs` and used the secure `mbstowcs_s`.

Thanks,
Alexis.

### DIFF
--- a/adapters/httpapi_winhttp.c
+++ b/adapters/httpapi_winhttp.c
@@ -58,7 +58,7 @@ static const char* ConstructHeadersString(HTTP_HEADERS_HANDLE httpHeadersHandle)
     else
     {
         size_t i;
-        
+
         /*the total size of all the headers is given by sumof(lengthof(everyheader)+2)*/
         size_t toAlloc = 0;
         for (i = 0; i < headersCount; i++)
@@ -84,7 +84,7 @@ static const char* ConstructHeadersString(HTTP_HEADERS_HANDLE httpHeadersHandle)
         else
         {
             result = (char*)malloc(toAlloc*sizeof(char) + 1 );
-            
+
             if (result == NULL)
             {
                 LogError("unable to malloc");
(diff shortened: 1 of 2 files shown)


adapters/httpapi_winhttp.c
```diff
@@ -444,27 +442,36 @@ HTTPAPI_RESULT HTTPAPI_ExecuteRequest(HTTP_HANDLE handle, HTTPAPI_REQUEST_TYPE r
...
                                                                 {
-                                                                    result = HTTPAPI_OK;
+                                                                    if (WinHttpSetCredentials(requestHandle,
+                                                                        WINHTTP_AUTH_TARGET_PROXY,
+                                                                        WINHTTP_AUTH_SCHEME_BASIC,
+                                                                        wusername,
+                                                                        wpassword,
+                                                                        NULL) != TRUE)
+                                                                    {
+                                                                        printf("Failure setting proxy credentials (%i)\n", GetLastError());
```
jspaith: Please use  LogErrorWinHTTPWithGetLastErrorAsString.

jspaith:
I have one small change that needs addressing & squashing but looks good otherwise.  Once you get this fixed, we can take this in.  Thanks.

alexmrtn:
Hi, I made the modification and the squashing, thanks @jspaith!

PR #152, opened 2018-04-24
Use declspec macro instead of .def file

markrad:
Putting this out here for review. I know there are a couple of items I need to address, one being the DLLEXPORT macro defined in two places but I wanted to get some general feedback on the changes before I go any further. Passes e2e and unit tests. Also installs the C shared so file properly with version numbers etc.


inc/azure_c_shared_utility/agenttime.h
```diff
@@ -19,42 +19,46 @@
...
-MOCKABLE_FUNCTION(, time_t, get_mktime, struct tm*, cal_time);
+DLLEXPORT MOCKABLE_FUNCTION(, time_t, get_mktime, struct tm*, cal_time);
 
 /** @brief Gets a C-string representation of the given time.
 *
 *	@details This function provides the same functionality as the
 *	standard C @c ctime() function.
 */
-MOCKABLE_FUNCTION(, char*, get_ctime, time_t*, timeToGet);
+DLLEXPORT MOCKABLE_FUNCTION(, char*, get_ctime, time_t*, timeToGet);
```
dcristoloveanu: >DLLEXPORT [](start = 0, length = 9)

What does this buy us over .def?

jspaith: UploadToblob flag currently requires 2 .def files -- one with uploadtoblob stuff and the other without.  With modules (sadly) looking like they'll need a separate cmake flag and in turn need custom Dllexports based on whether its built, the combinitorics of managing this with .def files start blowing up.  

We could try to filter out a master .def as we did in WinCE/SYSGEN days but those were bespoke CE tools IIRC.  Building this up based on declspec's seems most natural.

Thoughts?

CMakeLists.txt
```diff
@@ -490,6 +506,13 @@ endif()
 # Set CMAKE_INSTALL_* if not defined
 include(GNUInstallDirs)
 
+    
+if(${build_as_dynamic})
+    set(targets aziotsharedutil aziotsharedutil_dll)
+else(${build_as_dynamic})
```
jspaith: Could you please replace with else() which means same thing and is more readable to non-cmake experts.

CMakeLists.txt
```diff
@@ -353,8 +365,12 @@ endif()
 
 if(${build_as_dynamic})
     #make sure we can link as dll/so
-    add_library(aziotsharedutil_dll SHARED ./src/aziotsharedutil.def ${source_c_files} ${source_h_files})
```
jspaith: >./src/aziotsharedutil.def  [](start = 43, length = 26)

Could we please remove this .def file from the repo?

inc/azure_c_shared_utility/umock_c_prod.h
```diff
@@ -28,6 +34,6 @@
 /* Codes_SRS_UMOCK_C_LIB_01_005: [**If ENABLE_MOCKS is not defined, MOCKABLE_FUNCTION shall only generate a declaration for the function.] */
 /* Codes_SRS_UMOCK_C_LIB_01_001: [MOCKABLE_FUNCTION shall be used to wrap function definition allowing the user to declare a function that can be mocked.]*/
 #define MOCKABLE_FUNCTION(modifiers, result, function, ...) \
-    result modifiers function(IF(COUNT_ARG(__VA_ARGS__),,void) FOR_EACH_2_COUNTED(UMOCK_C_PROD_ARG_IN_SIGNATURE, __VA_ARGS__));
+	result modifiers function(IF(COUNT_ARG(__VA_ARGS__), , void) FOR_EACH_2_COUNTED(UMOCK_C_PROD_ARG_IN_SIGNATURE, __VA_ARGS__));
```
jspaith: >result modifiers function(IF(COUNT_ARG(__VA_ARGS__), , void) FOR_EACH_2_COUNTED(UMOCK_C_PROD_ARG_IN_SIGNATURE, __VA_ARGS__)); [](start = 1, length = 125)

Do we need to change the space here in final product?  If not, could we please revert just to limit # of changes.

inc/azure_c_shared_utility/umock_c_prod.h
```diff
@@ -10,6 +10,12 @@
     by adjusting the include paths).
 */
 
+#if defined(_DLL) && (defined(_WIN32) || defined(_WIN64))
```
jspaith: >_DLL [](start = 12, length = 4)

Where does the _DLL get defined?

inc/azure_c_shared_utility/umock_c_prod.h
```diff
@@ -10,6 +10,12 @@
...
+#define DLLEXPORT __declspec(dllexport)
+#else
+#define DLLEXPORT
+#endif
```
dcristoloveanu: This is the wrong place to do this. umock should not care about DLLEXPORT.

dcristoloveanu: As a matter of fact any changes to this file should be reverted. Normally the C shared and IoT SDK should use the umock_c_prod header from umock. But because MBED this shortcut exists and it never got cleaned up. The point is you dont want to have this file diverging from the one in umock_c.

---
In reply to: [184153082](https://github.com/Azure/azure-c-shared-utility/pull/152#discussion_r184153082) [](ancestors = 184153082)

inc/azure_c_shared_utility/agenttime.h
dcristoloveanu: I see.

---
In reply to: [184147428](https://github.com/Azure/azure-c-shared-utility/pull/152#discussion_r184147428) [](ancestors = 184147428)

inc/azure_c_shared_utility/umock_c_prod.h
dcristoloveanu: I could suggest having a common header called "dll_export.h" :-) which should contain the define et all

---
In reply to: [184153426](https://github.com/Azure/azure-c-shared-utility/pull/152#discussion_r184153426) [](ancestors = 184153426,184153082)

inc/azure_c_shared_utility/threadapi.h
```diff
@@ -45,7 +45,7 @@ typedef void* THREAD_HANDLE;
  * @return	@c THREADAPI_OK if the API call is successful or an error
  * 			code in case it fails.
  */
-MOCKABLE_FUNCTION(, THREADAPI_RESULT, ThreadAPI_Create, THREAD_HANDLE*, threadHandle, THREAD_START_FUNC, func, void*, arg);
+DLLEXPORT MOCKABLE_FUNCTION(, THREADAPI_RESULT, ThreadAPI_Create, THREAD_HANDLE*, threadHandle, THREAD_START_FUNC, func, void*, arg);
```
dcristoloveanu: >DLLEXPORT [](start = 0, length = 9)

This relies of the fact that MOCKABLE_FUNCTION must generate a function as the first thing it does. Putting this assumtion in is a failure waiting to happen if MOCKABLE_FUNCTION changes slightly.
Can you just put DLLEXPORT as the first argument in the MOCKABLE_FUNCTION macro?

inc/azure_c_shared_utility/tlsio_schannel.h
```diff
@@ -14,15 +14,27 @@ extern "C" {
...
+    TLSIO_STATE_NOT_OPEN,                         \
+    TLSIO_STATE_OPENING_UNDERLYING_IO,            \
+    TLSIO_STATE_HANDSHAKE_CLIENT_HELLO_SENT,      \
+    TLSIO_STATE_HANDSHAKE_SERVER_HELLO_RECEIVED,  \
+    TLSIO_STATE_RENEGOTIATE,                      \
+    TLSIO_STATE_OPEN,                             \
+    TLSIO_STATE_CLOSING,                          \
+    TLSIO_STATE_ERROR
+
+DEFINE_ENUM(TLSIO_STATE, TLSIO_STATE_VALUES);
```
dcristoloveanu: Sneaky. This change has nothing to do with DLLEXPORT.
And it should not go in as the states are actually not used in the interface anywhere. Thus they are not really mandated.

CMakeLists.txt
markrad: While I agree it makes it more readable to non-cmake experts I would defend this type of usage because in the event of mismatched if, elseif, else constructs first it is easier to see them and secondly cmake is very good at pointing out exactly where your error has occurred. I would advocate for using this construction for all if, elseif, else constructs but, if you really want me to, I will take it out.

---
In reply to: [184149662](https://github.com/Azure/azure-c-shared-utility/pull/152#discussion_r184149662) [](ancestors = 184149662)

CMakeLists.txt
markrad: One of the clean up items on my todo list.

---
In reply to: [184151426](https://github.com/Azure/azure-c-shared-utility/pull/152#discussion_r184151426) [](ancestors = 184151426)

inc/azure_c_shared_utility/umock_c_prod.h
markrad: _DLL is automatically defined by the MS build system when the target is a dll

---
In reply to: [184152808](https://github.com/Azure/azure-c-shared-utility/pull/152#discussion_r184152808) [](ancestors = 184152808)

inc/azure_c_shared_utility/threadapi.h
markrad: No. I tried that. The issue is that the MOCKABLE_FUNCTION macro generates (in this example) THREADAPI_RESULT modifier ThreadAPI_Create(args...). In some instances in the SDK we already have the value WINAPI passed into the MOCKABLE_FUNCTION macro which works fine because __stdcall _must_ be declared after the return type. However, __declspec(dllexport) needs to be declared _before_ the return type so it needs to read __declspec(dllexport) THREADAPI_RESULT ThreadAPI_Create. I toyed with the idea of a premodifier and a postmodifier but really didn't want to mess about with the MOCKABLE _FUNCTION macro. 

I am open to ideas...

---
In reply to: [184155255](https://github.com/Azure/azure-c-shared-utility/pull/152#discussion_r184155255) [](ancestors = 184155255)

inc/azure_c_shared_utility/tlsio_schannel.h
markrad: Actually not so sneaky. In order export TLSIO_STATE_FromString and TLSIO_STATEStrings I had to put the define in the header file since the export macro is only used in DEFINE_ENUM. All of these enum string functions are exported for the various return values. If they are not used in the interface anywhere then I think it should have been defined with DEFINE_LOCAL_ENUM but I'm not sure. Those are not exported.

---
In reply to: [184158329](https://github.com/Azure/azure-c-shared-utility/pull/152#discussion_r184158329) [](ancestors = 184158329)

inc/azure_c_shared_utility/agenttime.h
```diff
@@ -19,42 +19,46 @@
 extern "C"
 {
 #endif
+//#include <windows.h>
+
```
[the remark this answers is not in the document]
markrad: Another clean up item.

---
In reply to: [184150319](https://github.com/Azure/azure-c-shared-utility/pull/152#discussion_r184150319) [](ancestors = 184150319)

inc/azure_c_shared_utility/umock_c_prod.h
markrad: My bad.

---
In reply to: [184152458](https://github.com/Azure/azure-c-shared-utility/pull/152#discussion_r184152458) [](ancestors = 184152458)

inc/azure_c_shared_utility/umock_c_prod.h
markrad: Agreed. This was for the purposes of testing. I had considered if this functionality could be incorporated into the MOCKABLE_FUNCTION macro but decided better of it. I have now defined it in macro_utils.h which seemed an appropriate place. Happy to put it in its own file though if you so desire. I'll get rid of the commits in this file before I send it to the build gate.

---
In reply to: [184154823](https://github.com/Azure/azure-c-shared-utility/pull/152#discussion_r184154823) [](ancestors = 184154823,184153426,184153082)

inc/azure_c_shared_utility/agenttime.h
dcristoloveanu: We (me and Andrei) actually have a better (we believe) proposal.

Have one .def file for the most of the SDK.
Have another .def file for the upload to blob part, since that is an optional feature.

If the upload to blob is enabled then put both .def files in the files list in cmake:

set(iothub_client_def_files
    file1.def
    file2.def
)

add_library(iothub_client_dll SHARED
    ${iothub_client_c_files} 
    ${iothub_client_h_files}
    ${iothub_client_def_files}
)

Basically the idea is that DLLEXPORT is special packaging needed for Windows DLL targets.

The source code and packaging should be kept as far as possible from each other.
Or put it another way: the source should have as few knowledge as possible of stuff that is related to packaging.

That keeps code clean and things do not creep in.

HTH.

---
In reply to: [184153567](https://github.com/Azure/azure-c-shared-utility/pull/152#discussion_r184153567) [](ancestors = 184153567,184147428)

inc/azure_c_shared_utility/tlsio_schannel.h
dcristoloveanu: What I mean is: since the DEFINE_ENUM was not in there up to now,m why does it need to be in now?

---
In reply to: [184223100](https://github.com/Azure/azure-c-shared-utility/pull/152#discussion_r184223100) [](ancestors = 184223100,184218857,184158329)

jspaith:
Closing this because we've moved to multiple .def files instead.  Thanks.

PR #153, opened 2018-04-30
Added support to set the TLS and Ciphers

jkriba:
**Description of the problem**
This is an Enhancement based upon Critical Customer requirement and security audit finding towards Enforce TLS version and Cipher Suites for Honeywell IoT Devices [Gateways]. As a platform all device connection should be made via approved, secure cipher suites. However, today on Azure we don't control the cipher suites supported by Microsoft for Azure IoT Hub as they have to support many customers and devices which all don't have the latest ciphers.
As per our customer requirements in Honeywell we wanted to provide an option for the users to set the configuration of TLS version & cipher suites which should help the users to solve the above mentioned problem.

**Description of the solution**
The main modification is in the ‘c-utility’ module. We’ve modified two files, the full file names are:
c-utility/inc/azure_c_shared_utility/shared_util_options.h
c-utility/src/tlsio_openssl.c
and provided options for the user to set the relevent TLS version and Cipher suites from the Edge in order to support selection of them which is secure and relevant for the device environment.

### DIFF
--- a/inc/azure_c_shared_utility/shared_util_options.h
+++ b/inc/azure_c_shared_utility/shared_util_options.h
@@ -23,6 +23,7 @@ extern "C"
     static STATIC_VAR_UNUSED const char* const OPTION_HTTP_TIMEOUT = "timeout";
 
     static STATIC_VAR_UNUSED const char* const OPTION_TRUSTED_CERT = "TrustedCerts";
+    static STATIC_VAR_UNUSED const char* const OPTION_CIPHER_SUITE = "CipherSuite";
 
     static STATIC_VAR_UNUSED const char* const SU_OPTION_X509_CERT = "x509certificate";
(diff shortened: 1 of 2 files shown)


jebrando:
Thank you for the update, I will have to talk to compliance to see if we would want to take this change.  Stay tuned for updates.

jspaith:
We're concerned about this proposal.  Let's say that an app sets cipher to "BestInClass-2018" and then feels good.  Fast forward to 2028.  Let's say this algorithm is no longer best practice and may even have weaknesses (but the server may accept these connections anyway for back-compat).  It happens very frequently that app devs set and forget these type of custom options, even if security community tells them to move on, so the 2028 code would still be tightly coupled to old tech by hard-coding in the 2018 ciphers.

By letting the SSL layer between the client and the server handle this negotiation, in 2028 their defaults *should* be BestInClass-2028.  Running 'openssl ciphers' on Ubuntu 18.04 shows what I believe to be the default order that it will chose.

Are there specific scenarios where this may not be the case?

PR #155, opened 2018-05-15
Detect host closure on socket read

JetstreamRoySprowl:


pal/ios-osx/tlsio_appleios.c
```diff
@@ -416,23 +416,35 @@ static void dowork_read(TLS_IO_INSTANCE* tls_io_instance)
...
-        
-        if (rcv_bytes > 0)
+        CFStreamStatus read_status = CFReadStreamGetStatus(tls_io_instance->sockRead);
+        if (read_status == kCFStreamStatusAtEnd)
         {
-            // tls_io_instance->on_bytes_received was already checked for NULL
-            // in the call to tlsio_appleios_open_async
-            /* Codes_SRS_TLSIO_30_100: [ As long as the TLS connection is able to provide received data, tlsio_dowork shall repeatedly read this data and call on_bytes_received with the pointer to the buffer containing the data, the number of bytes received, and the on_bytes_received_context. ]*/
-            tls_io_instance->on_bytes_received(tls_io_instance->on_bytes_received_context, buffer, rcv_bytes);
+            LogInfo("Communications closed by host");
```
jebrando: >LogInfo [](start = 12, length = 7)

Change to LogError, LogInfo will need to be turned off, but we don't have that ability yet.

Issue #154, opened 2018-05-04
uws_client is dropping continuation frames


blico:
Currently uws_client is not handling frame fragmentation. 
o **SRS_UWS_CLIENT_01_213**: [ A fragmented message consists of a single frame with the FIN bit clear and an opcode other than 0, followed by zero or more frames with the FIN bit clear and the opcode set to 0, and terminated by a single frame with the FIN bit set and an opcode of 0. ]
o **SRS_UWS_CLIENT_01_220**: [ Clients and servers MUST support receiving both fragmented and unfragmented messages. ]

The first fragmented frame has an opcode of either TEXT or BINARY, but the frames after have an opcode of WS_CONTINUATION_FRAME, which uws_client drops.

One fix would be to buffer a fragmented frame within uws_client until it has been fully received, and then send it up. However, this technically breaks the Websocket protocol:
**SRS_UWS_CLIENT_01_227**: [ In the absence of any extension, a receiver doesn't have to buffer the whole frame in order to process it. ]
For example, if a streaming API is used, a part of a frame can be delivered to the application. However, note that this assumption might not hold true for all future WebSocket extensions.

A proper fix would be to start bubbling up CONTINUATION_FRAMES to clients of uws_client. One complication with this fix is that currently the on_ws_frame_received callback doesn’t send the first two bytes of the frame. The most significant bit of the frame is needed by the client to reassemble fragmented frames (the FIN bit notifies that a frame is fragmented).  One way to get around this would be to add a “isFinal” flag to the on_ws_frame_received callback. However, either way this would involve a breaking change to the client.

PR #156, opened 2018-05-15
Add UWS support for handling websocket frame fragmentation

blico:
Fix for issue #154 

uws_client was not handling frame fragmentation, resulting in the dropping of fragmented frames:

- SRS_UWS_CLIENT_01_213: [ A fragmented message consists of a single frame with the FIN bit clear and an opcode other than 0, followed by zero or more frames with the FIN bit clear and the opcode set to 0, and terminated by a single frame with the FIN bit set and an opcode of 0. ]

- SRS_UWS_CLIENT_01_220: [ Clients and servers MUST support receiving both fragmented and unfragmented messages. ]

Fixed by buffering fragmented frames, and not bubbling up to user until the frame is fully received. 
Tested by ensuring it fixes our broken scenario, and added unit tests to uws_client_ut.


src/uws_client.c
```diff
@@ -1356,10 +1477,16 @@ static void on_underlying_io_bytes_received(void* context, const unsigned char*
                                 size_t pong_frame_length;
                                 BUFFER_HANDLE pong_frame_buffer;
 
-                                uws_client->uws_state = UWS_STATE_ERROR;
```
blico: Worth calling out that I removed this line. It seems like a bug that we are setting the state to error when receiving a ping frame, but I could be missing something.

src/uws_client.c
```diff
@@ -1184,54 +1185,173 @@ static void on_underlying_io_bytes_received(void* context, const unsigned char*
...
+                                    else
+                                    {
+                                        uws_client->fragment_buffer = new_fragment_bytes;
+                                        (void)memcpy(uws_client->fragment_buffer + uws_client->fragment_buffer_count, uws_client->stream_buffer + needed_bytes - length, length);
+                                        uws_client->fragment_buffer_count = length;
+                                    }
+                                    /* Codes_SRS_UWS_CLIENT_01_225: [ As a consequence of these rules, all fragments of a message are of the same type, as set by the first fragment's opcode. ]*/
+                                    /* Codes_SRS_UWS_CLIENT_01_226: [ Since control frames cannot be fragmented, the type for all fragments in a message MUST be either text, binary, or one of the reserved opcodes. ]*/
+                                    uws_client->fragmented_frame_type = WS_FRAME_TYPE_TEXT;
+                                }
```
jebrando: This section of code is repeated below for binary, can you please create a function to reduce duplication.

src/uws_client.c
jebrando: I talked to the implementer and you are correct.

---
In reply to: [188471094](https://github.com/Azure/azure-c-shared-utility/pull/156#discussion_r188471094) [](ancestors = 188471094)

tests/uws_client_ut/uws_client_ut.c
```diff
@@ -3932,6 +3937,319 @@ TEST_FUNCTION(when_a_0_bytes_text_frame_is_received_it_shall_be_indicated_to_the
...
+    // act
+    g_on_bytes_received(g_on_bytes_received_context, test_stream, sizeof(test_stream));
+    // assert
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+
+    // cleanup
+    uws_client_destroy(uws_client);
+}
+
+/* Codes_SRS_UWS_CLIENT_01_213: [ A fragmented message consists of a single frame with the FIN bit clear and an opcode other than 0, followed by zero or more frames with the FIN bit clear and the opcode set to 0, and terminated by a single frame with the FIN bit set and an opcode of 0. ]*/
```
jebrando: >Codes [](start = 3, length = 5)

Can you please change these to Tests instead of Codes.

blico: done

src/uws_client.c
blico: added process_frame_fragment() function

blico:
@jebrando I addressed your feedback in my latest commit. Let me know if you have any other feedback.

PR #157, opened 2018-05-30
Enabling stricter warnings in c shared utilities library

mrohera:


mrohera:
PR for ctest and umock-c have been completed and merged.

JetstreamRoySprowl:
Damon’s right – make sure you check with Dan Cristoloveanu about the change to micromock.
Roy

configs/azure_iot_build_rules.cmake
```diff
@@ -191,12 +191,12 @@ function(add_files_to_install filesToBeInstalled)
...
 
-# XCode warns about unused variables and unused static functions,
-# both of which are produced by serializer
+# XCode and stricter warning levels such as -Wall and -Wextra warn about unused
+# variables and unused static functions, both of which are produced by serializer
 function(usePermissiveRulesForSamplesAndTests)
-    if(XCODE)
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-variable  -Wno-unused-function -Wno-missing-braces"  PARENT_SCOPE)
-        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-unused-variable  -Wno-unused-function -Wno-missing-braces" PARENT_SCOPE)
+    if(NOT WIN32)
```
dcristoloveanu: >NOT WIN32 [](start = 7, length = 9)

This sounds more like if GCC

adapters/httpapi_compact.c
```diff
@@ -832,7 +832,7 @@ static HTTPAPI_RESULT SendHeadsToXIO(HTTP_HANDLE_DATA* http_instance, HTTPAPI_RE
     /*Codes_SRS_HTTPAPI_COMPACT_21_038: [ The HTTPAPI_ExecuteRequest shall execute the resquest for the path in relativePath parameter. ]*/
     /*Codes_SRS_HTTPAPI_COMPACT_21_036: [ The request type shall be provided in the parameter requestType. ]*/
     if (((ret = snprintf(buf, sizeof(buf), "%s %s HTTP/1.1\r\n", get_request_type(requestType), relativePath)) < 0) ||
-        (ret >= sizeof(buf)))
+        ((size_t)ret >= sizeof(buf)))
```
dcristoloveanu: >ret [](start = 17, length = 3)

Sounds like this should be a size_t?

mrohera: [https://linux.die.net/man/3/snprintf](https://linux.die.net/man/3/snprintf) returns an int and a negative number indicates an error. Not sure if there is anything else we can do here other than "up" type-casting 'ret' after checking that it is non negative.

configs/azure_iot_build_rules.cmake
mrohera: Yes you are right, I misunderstood the WIN32 flag. Would this check be okay instead: if (NOT MSVC)?

mrohera: Also I prefer the NOT expression because these flags are relevant for Clang, GCC etc.

PR #158, opened 2018-06-06
Domain sockets

jspaith:
Enable unix domain sockets for Edge.  Did a mini-refactor of code while at it as it would've got unweildy.


adapters/socketio_berkeley.c
```diff
@@ -248,6 +256,144 @@ static STATIC_VAR_UNUSED void signal_callback(int signum)
...
+    
+    do
+    {
+        retval = select(socket_io_instance->socket + 1, NULL, &fdset, NULL, &tv);
+    
+        if (retval < 0)
+        {
+            select_errno = errno;
+        }
+    } while (retval < 0 && select_errno == EINTR);
```
jebrando: This is a possible endless loop, on some obscure distros.  Maybe you should put a max times this loops should execute.

adapters/socketio_berkeley.c
```diff
@@ -569,8 +716,6 @@ void socketio_destroy(CONCRETE_IO_HANDLE socket_io)
 int socketio_open(CONCRETE_IO_HANDLE socket_io, ON_IO_OPEN_COMPLETE on_io_open_complete, void* on_io_open_complete_context, ON_BYTES_RECEIVED on_bytes_received, void* on_bytes_received_context, ON_IO_ERROR on_io_error, void* on_io_error_context)
```
jebrando: >socketio_open [](start = 4, length = 13)

You forgot to fix the tests for this and add a couple for the new functionality.  Testing for this unit is very hit and miss.  I'd like to put in on my backlog to add more.

adapters/socketio_berkeley.c
jspaith: Will open a PR on this later; this was as it was before.

---
In reply to: [193445412](https://github.com/Azure/azure-c-shared-utility/pull/158#discussion_r193445412) [](ancestors = 193445412)

PR #160, opened 2018-06-20
add ability to disable ssl verification in HTTPAPI

alexmrtn:
Hi,

We would like to contribute to the c-shared-utility by adding the ability to trust insecure certificates for HTTPS connections. Our use case is when devices are out of time sync and they can't connect to the iothub. We can manually contact a time server using the HTTPAPI to sync the time of the device. As the HTTPAPI only support HTTPS connection, we have to bypass the SSL verification of the time server.

Alexis.

### DIFF
--- a/adapters/httpapi_curl.c
+++ b/adapters/httpapi_curl.c
@@ -37,6 +37,7 @@ typedef struct HTTP_HANDLE_DATA_TAG
     long forbidReuse;
     long freshConnect;
     long verbose;
+    long ssl_verifypeer;
     const char* x509privatekey;
     const char* x509certificate;
     const char* certificates; /*a list of CA certificates*/
@@ -132,6 +133,7 @@ HTTP_HANDLE HTTPAPI_CreateConnection(const char* hostName)
                         httpHandleData->forbidReuse = 0;
                         httpHandleData->freshConnect = 0;
                         httpHandleData->verbose = 0;
+                        httpHandleData->ssl_verifypeer = 1;
                         httpHandleData->x509certificate = NULL;
                         httpHandleData->x509privatekey = NULL;
                         httpHandleData->certificates = NULL;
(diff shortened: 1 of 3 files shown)


jspaith:
@alexrtm - sorry for delayed response on this one.

Could you please talk directly to curl to get to the time server, where it gives you unlimited options, instead of through this?

We've seen *many* issues across multiple SDK's where the moment we plumb in a "disable security" option, its misused.  These disable certificate checks are particularly dangerous (a) because devs disable them to quickly get things going and (b) forget they're in, and the code they write is fine and yet silently vulnerable to man in the middle attacks.  

While Microsoft is certainly not perfect on security(!), we take it very seriously and especially for embedded devices.  Even here, we've had problems with analogous back-doors [here](https://www.symantec.com/security-center/vulnerabilities/writeup/104070).

alexmrtn:
@jspaith I totally understand the reason why adding a "disable security" option is a not so good idea and that it can be easily misused.

This solution was the "easy" way to solve our problem. Using the HTTPAPI from the c-utility that leverage the problem of multi-platform support (the product we are building must run on both Linux and Windows) would be for us the best way to avoid platform-specific code + code duplication (regarding the c-utility). 

Maybe a better solution would be to add the HTTP support in the HTTPAPI in addition to HTTPS. What do you think about that? I can work on this and submit a new PR if this is ok for you.

Alexis.

jspaith:
Another caveat -- we don't **directly** e2e test the HTTPAPI layers, nor do we have the same level of API contract promises that come from the Azure IoT C SDK.  Obviously its in our best interest never to break these things, but if something were to go wrong here we would likely only discover it when updating the submodule for the C SDK which is extensively e2e tested.

You could mitigate this to some extent by staying on same c-utility that C SDK's master points at since that implies its past the C SDK gates.

That said - if you're comfortable with these caveats, we'd be willing to take a PR for HTTP specific.  I'm closing this one per comments above, though, so we'd start clean.

Thanks

PR #161, opened 2018-06-22
add HEAD option in HTTPAPI

alexmrtn:
Hi

The HTTPAPI is missing the HEAD request, here a PR to fill this gap.

Alexis.


adapters/httpapi_curl.c
```diff
@@ -401,6 +401,31 @@ HTTPAPI_RESULT HTTPAPI_ExecuteRequest(HTTP_HANDLE handle, HTTPAPI_REQUEST_TYPE r
 
                     break;
 
+                case HTTPAPI_REQUEST_HEAD:
+                    if (curl_easy_setopt(httpHandleData->curl, CURLOPT_HTTPGET, 1L) != CURLE_OK)
+                    {
+                        result = HTTPAPI_SET_OPTION_FAILED;
+                        LogError("(result = %s)", ENUM_TO_STRING(HTTPAPI_RESULT, result));
+                    }
+                    else
```
jspaith: Instead of the 
else 
{
    if (...)
[...]

Could you please change this to be more of a block of

else if (doFoo() != OK) 
{ 
  ...
}
else if (doBar()) !+ OK) 
{ 
 
}

I know this isn't current pattern in this file but its what rest of SDK is moving to since its way less indentation and easier to follow code flow.

adapters/httpapi_curl.c
```diff
@@ -401,6 +401,31 @@ HTTPAPI_RESULT HTTPAPI_ExecuteRequest(HTTP_HANDLE handle, HTTPAPI_REQUEST_TYPE r
 
                     break;
 
+                case HTTPAPI_REQUEST_HEAD:
+                    if (curl_easy_setopt(httpHandleData->curl, CURLOPT_HTTPGET, 1L) != CURLE_OK)
```
jspaith: Do you need to explicitly set the CURLOPT_HTTPGET?  My read of https://curl.haxx.se/libcurl/c/CURLOPT_NOBODY.html is that the NOBODY option is enough.

jspaith:
Looks good.  If you could address my minor comments and squash, we can get this in.  Thanks.

jspaith:
@alexmrtn - is this something you're still interested in?  If yes could you please make the changes when you have cycles, or otherwise close the PR?  Thanks

Issue #151, opened 2018-04-19
refcount.h should not assign result->count = 1


anporumb:
in line 58 of refcount.h the reference count of the object is set to "1" by assignment.

On windows at least, it should be set with InterlockedExchange function.

Best Regards,
Andrei Porumb

jebrando:
Is there a Linux equivalent to this?

dcristoloveanu:
The current Linux implementation of refcount_os.h uses either atomic_fetch_add or __sync_add_and_fetch for increment.

Similarly for assignment something that ensures a memory barrier should be used. That would be atomic_store if C11 atomics are available (or __sync_synchronize if gnu atomics are used).

jebrando:
I looked at this and it will require some changes.  I will set it as the appropriate label.

dcristoloveanu:
This has been fixed.

PR #165, opened 2018-07-09
Fix refcount to use full memory barriers

dcristoloveanu:
Fix refcount to use full memory barriers.
This addrresses #151.

### DIFF
--- a/inc/azure_c_shared_utility/refcount.h
+++ b/inc/azure_c_shared_utility/refcount.h
@@ -48,14 +48,14 @@ C2(REFCOUNT_, type)
 REFCOUNT_TYPE(type)                                                                                  \
 {                                                                                                    \
     type counted;                                                                                    \
-    COUNT_TYPE count;                                                                                  \
+    COUNT_TYPE count;                                                                                \
 };                                                                                                   \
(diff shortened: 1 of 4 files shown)


inc/azure_c_shared_utility/refcount.h
```diff
@@ -69,6 +69,9 @@ static type* REFCOUNT_TYPE_DECLARE_CREATE(type) (void)
 #ifndef DEC_REF
 #error refcount_os.h does not define DEC_REF
 #endif // !DEC_REF
+#ifndef INIT_REF
+#error refcount_os.h does not define INIT_REF
+#endif // !INIT_REF
```
jebrando: Why are these necessary?  Won't the linker complain if the defines are not there?

dcristoloveanu: I think Roy added these to have some ... nicer error messages. I do not remember them existing in the first version of refcount.h done eons ago. But anyhow, I guess I am Ok with having them for slightly better errors. Was not that painful to add :-)

---
In reply to: [201130723](https://github.com/Azure/azure-c-shared-utility/pull/165#discussion_r201130723) [](ancestors = 201130723)

PR #167, opened 2018-07-19
Fix compiling for MinGW-W64

vkuzmichev:
# Checklist
- [x] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-c/blob/master/.github/CONTRIBUTING.md).
- [ ] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).
- If this is a modification that impacts the behavior of a public API
  - [ ] I edited the corresponding document in the `devdoc` folder and added or modified requirements.
- I submitted this PR against the correct branch: 
  - [x] This pull-request is submitted against the `master` branch. 
  - [ ] I have merged the latest `master` branch prior to submission and re-merged as needed after I took any feedback.
  - [ ] I have squashed my changes into one with a clear description of the change.

# Reference/Link to the issue solved with this PR (if any)
None.

# Description of the problem
Azure IoT SDK does not compile in Windows using MinGW-W64 compiler.

# Description of the solution
1) Disabled secure string functions "strcpy_s", "strcat_s", "strncpy_s", "sprintf_s" in 'azure-c-shared-utility' if they are already defined and implemented in MinGW-W64.
2) Added missing LogErrorWinHTTPWithGetLastErrorAsString macro definition for non-Microsoft compilers.

### DIFF
--- a/inc/azure_c_shared_utility/crt_abstractions.h
+++ b/inc/azure_c_shared_utility/crt_abstractions.h
@@ -81,8 +81,8 @@ typedef unsigned char bool;
 /* Codes_SRS_CRT_ABSTRACTIONS_99_001:[The module shall not redefine the secure functions implemented by Microsoft CRT.] */
 /* Codes_SRS_CRT_ABSTRACTIONS_99_040 : [The module shall still compile when building on a Microsoft platform.] */
 /* Codes_SRS_CRT_ABSTRACTIONS_99_002: [CRTAbstractions module shall expose the following API]*/
-#ifdef _MSC_VER
-#else // _MSC_VER
+#if defined (_MSC_VER) || defined (MINGW_HAS_SECURE_API)
+#else // _MSC_VER || MINGW_HAS_SECURE_API
 
 /* Adding definitions from errno.h & crtdefs.h */
 #if !defined (_TRUNCATE)
@@ -97,7 +97,7 @@ extern int strcpy_s(char* dst, size_t dstSizeInBytes, const char* src);
 extern int strcat_s(char* dst, size_t dstSizeInBytes, const char* src);
 extern int strncpy_s(char* dst, size_t dstSizeInBytes, const char* src, size_t maxCount);
 extern int sprintf_s(char* dst, size_t dstSizeInBytes, const char* format, ...);
-#endif // _MSC_VER
+#endif // _MSC_VER || MINGW_HAS_SECURE_API
 
 extern unsigned long long strtoull_s(const char* nptr, char** endPtr, int base);
 extern float strtof_s(const char* nptr, char** endPtr);
--- a/inc/azure_c_shared_utility/xlogging.h
+++ b/inc/azure_c_shared_utility/xlogging.h
@@ -135,9 +135,43 @@ extern LOGGER_LOG_GETLASTERROR xlogging_get_log_function_GetLastError(void);
                     }\
                 }\
             } while((void)0,0)
-#else
+#else // _MSC_VER
 #define LogError(FORMAT, ...) do{ LOG(AZ_LOG_ERROR, LOG_LINE, FORMAT, ##__VA_ARGS__); }while((void)0,0)
-#endif
+
+#ifdef WIN32
+#define LogErrorWinHTTPWithGetLastErrorAsString(FORMAT, ...) do { \
+                DWORD errorMessageID = GetLastError(); \
+                char messageBuffer[MESSAGE_BUFFER_SIZE]; \
+                LogError(FORMAT, ##__VA_ARGS__); \
+                if (errorMessageID == 0) \
+                {\
+                    LogError("GetLastError() returned 0. Make sure you are calling this right after the code that failed. "); \
+                } \
+                else\
+                {\
+                    int size = FormatMessage(FORMAT_MESSAGE_FROM_HMODULE | FORMAT_MESSAGE_IGNORE_INSERTS, \
+                        GetModuleHandle("WinHttp"), errorMessageID, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), messageBuffer, MESSAGE_BUFFER_SIZE, NULL); \
+                    if (size == 0)\
+                    {\
+                        size = FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS, NULL, errorMessageID, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), messageBuffer, MESSAGE_BUFFER_SIZE, NULL); \
+                        if (size == 0)\
+                        {\
+                            LogError("GetLastError Code: %d. ", errorMessageID); \
+                        }\
+                        else\
+                        {\
+                            LogError("GetLastError: %s.", messageBuffer); \
+                        }\
+                    }\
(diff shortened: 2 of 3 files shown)


jspaith:
The `MINGW_HAS_SECURE_API` is quite eloquent - really like that.  

The feedback on the `LogErrorWinHTTPWithGetLastErrorAsString` is similar to the MinGW PR in C SDK proper.  The (existing) code is confusing the compiler with the platform the code is run on.  Pretty sure that the `LogErrorWinHTTPWithGetLastErrorAsString` should be wrapped in an `#ifdef WIN32` as it needs Win32 primitives.  The LogError itself - on quick examination - seems to need the MSVC wrapper per differences in how preprocessor behaves.

Like other change, if you agree & can make change & squash (:)), we can get this in.  

As a heads up, we update the c-sdk to point to latest c-utility submodule every few weeks or so.  We don't have a submodule update planned at present so it may be a bit until this reaches C SDK master, but it will arrive sooner than later.

Thanks again for sharing your fixes for MinGW.

vkuzmichev:
@jspaith 
Thank you for your review.

> Pretty sure that the ```LogErrorWinHTTPWithGetLastErrorAsString``` should be wrapped in an ```#ifdef WIN32``` as it needs Win32 primitives.

Right! I missed this out. Perhaps, because I did not try to compile my patch in Linux.
However, I have just verified that it builds as is. I guess this macro is used only from sources compiled for Windows only.
But I agree, it should look like the following:
```
#else // _MSC_VER
#define LogError(FORMAT, ...) do{ LOG(AZ_LOG_ERROR, LOG_LINE, FORMAT, ##__VA_ARGS__); }while((void)0,0)
#ifdef WIN32
#define LogErrorWinHTTPWithGetLastErrorAsString(FORMAT, ...) do { \
[...]
#else // WIN32
/* This actually should not be used from non-WIN32 sources */
#define LogErrorWinHTTPWithGetLastErrorAsString LogError
#endif // WIN32
#endif // _MSC_VER
```
The very first definition of  ```LogErrorWinHTTPWithGetLastErrorAsString``` under ```#if defined _MSC_VER``` does not look like requiring ```#ifdef WIN32```, since ```_MSC_VER``` is available in Windows only, so already implies ```WIN32```. This should work as is. Unless MS intend to release Visual Studio for Linux :)

> The LogError itself - on quick examination - seems to need the MSVC wrapper per differences in how preprocessor behaves.

Hmm. Do you mean that different MS compiler may process LogError differently?
I'm not so familiar with that, so it would be better if you prepare your own commit.
This change, however, does not look like mingw specific, so is a different topic?

----

Also I just spotted that ```LogLastError``` is defined for _MSC_VER:
```
#if defined _MSC_VER

#if !defined(WINCE)
extern void xlogging_set_log_function_GetLastError(LOGGER_LOG_GETLASTERROR log_function);
extern LOGGER_LOG_GETLASTERROR xlogging_get_log_function_GetLastError(void);
#define LogLastError(FORMAT, ...) do{ LOGGER_LOG_GETLASTERROR l = xlogging_get_log_function_GetLastError(); if(l!=NULL) l(__FILE__, FUNC_NAME, __LINE__, FORMAT, __VA_ARGS__); }while((void)0,0)
#endif

#define LogError[...]
```
But I can not find any use of ```LogLastError``` in Azure SDK. Probably there is no one, and this is why missing non-```_MSC_VER``` definition does not produce any compile errors.
Should I add definition for it similar way I define ```LogErrorWinHTTPWithGetLastErrorAsString```? I think, it should be added.

jspaith:
If there's a full Visual Studio coming out for Linux, knowledge of such is way beyond my pay grade!

As a heads up I'm about to be gone for ~1 week.  I've provided some context to my coworkers, but summer means we have some folks out so there maybe some delays in getting in.  It seems you're unblocked at least :) since you have the PR.

I took a quick look at the `LogLastError ` code and it looks like it's some sort of ETW tie in *in addition to the xlogging.h* but it's unclear thoes are compatible at same time even.  `EXTERN_C __declspec(selectany) const EVENT_DESCRIPTOR LogLastError = {0x4, 0x0, 0x11, 0x2, 0xd, 0x1, 0x4000000000000008};`  I haven't really dug into this much, but ETW is the Windows event tracer that is pretty heavily dependent on GUIDs.  As I said though I can't see how that and a macro defining something differert would work in practice in quirk review.

For the `LogErrorWinHTTPWithGetLastErrorAsString` for non Win32, it's quite fine if that's a #error.  Both WinHTTP and GetLastError are joined at hip to Win32 so we're not doing any favors to folks to let that code compile on Linux if they try to.  I agree the same pattern for defining `LogLastError` for non-Win32 - with a #error - should be appropriate.

vkuzmichev:
@jspaith 
Finally I have updated the branch with fresh version of logging related commit.
If I put ```#error``` in ```#else``` section it will not let to compile SDK in Linux.
So I left LogErrorWinHTTPWithGetLastErrorAsString undefined for non-WIN32, so first use of LogErrorWinHTTPWithGetLastErrorAsString in non-Win32 source code will cause compile error (missing definition).

jspaith:
Could you please take a look at https://github.com/Azure/azure-c-shared-utility/compare/minGw.  I've taken what you've done, but I've moved the core Win32 helper code into a helper function since it was too much for a macro and worse when duplicated.  It builds fine for me on MSVC/Win32 and also Ubuntu/clang.

If it's good, I'll get it in and I won't squash this one so your contribution stays in history.

vkuzmichev:
@jspaith 
Looks good. There is one change in your commit to note. ```LogError``` macros used within new function ```xlogging_LogErrorWinHTTPWithGetLastErrorAsStringFormatter``` will use ```__FILE__``` == src/xlogging.c, ```FUNC_NAME``` == new function name, and ```__LINE__``` number where it is located in src/xlogging.c instead of location where the macro ```LogErrorWinHTTPWithGetLastErrorAsString``` is actually used.
I think this is acceptable regression since you left one ```LogError``` within ```LogErrorWinHTTPWithGetLastErrorAsString``` definition, so at least first error message should point to actual place where error is detected.

You can just push your commit into this PR branch (although it is in my fork of repository), so all the work can be merged together.
Thank you.

jspaith:
Ahh - good catch on the log lines having xlogging.c in them.  But I agree that the line above captures original error, and the good of moving from pre-processor outweighs the 1 wrong file.

RE checking this in, the way our gates for C work is that I can't just point at a PR but have to create a temporary branch anyway.  Since it's off the PR we don't lose history, but it means here I might as well just bring `minGW` in directly.  Done!  

Thanks again for this fix.

jspaith:
Ugh -  7948cf422a05a46ef2ba56ee4074d4ba428ae205 fixes a tricky one that slipped through (and clearly my fault).  Happy I found this early instead of having some horrible to debug customer issue losing error code.  Change should be in shortly.

vkuzmichev:
@jspaith 
Interestingly. I though ```printf``` can not change the error code variable used by ```SetLastError*```/```GetLastError*```. ```printf``` even should not change ```errno```, but even if it can, I expect that this is not the same variable.
However, I agree to save error code in advance.

jspaith:
I saw the error code changed to 0 in one run, which motivated this fix.  Even if `printf` is leaving the `LastError` alone (it probably is), one of printf()`s callees must be doing something.  In any event should've been like this in 1st place to future proof.

PR #172, opened 2018-08-02
Weixu/uwp format

lisaweixu:

PR #173, opened 2018-08-02
revert some of the 

lisaweixu:


jspaith:
Could you please explain how this is different than #172?  Should #172 be retired?  Also could you please provide a bit more background about the motivation behind this change in 1st place?  Thanks!

lisaweixu:
Sorry, as I explained in #172. I did this PR against the wrong branch. I am closing it.

PR #177, opened 2018-08-14
Fix #174: avoid new GCC 8.2 warnings (s/strncpy/memcpy/)

### LINKED ISSUE #174: GCC 8.2 generates warnings when compiling crt_abstractions.c
Using

    gcc (Debian 8.2.0-3) 8.2.0

I'm getting the following warnings (and thus errors due to `-Werror`) when compiling `crt_abstractions.c` and 

```Bash

azure-iot-sdk-c/c-utility/src/crt_abstractions.c: In function ‘strncpy_s’:
azure-iot-sdk-c/c-utility/src/crt_abstractions.c:163:19: error: ‘strncpy’ output truncated before terminating nul copying as many bytes from a string as its length [-Werror=stringop-truncation]
             (void)strncpy(dst, src, srcLength);
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
azure-iot-sdk-c/c-utility/src/crt_abstractions.c:132:28: note: length computed here
         size_t srcLength = strlen(src);
                            ^~~~~~~~~~~
azure-iot-sdk-c/c-utility/src/crt_abstractions.c:149:23: error: ‘strncpy’ specified bound depends on the length of the source argument [-Werror=stringop-overflow=]
                 (void)strncpy(dst, src, srcLength);
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
azure/azure-iot-sdk-c/c-utility/src/crt_abstractions.c:132:28: note: length computed here
         size_t srcLength = strlen(src);
                            ^~~~~~~~~~~
```

GCC seems unable to determine that srcLength has been "corrected" to not exceed destlen. Maybe `memcpy()` should be used. NUL-termination is done manually right after.

pboettch:
GCC 8.2 is not able to determine that the length is OK for dst
and thus moans about the usage of strncpy.

How it is done, memcpy seems more appropriate to be used.

Maybe in the future this code can be even more simplified.

### DIFF
--- a/src/crt_abstractions.c
+++ b/src/crt_abstractions.c
@@ -146,7 +146,7 @@ int strncpy_s(char* dst, size_t dstSizeInBytes, const char* src, size_t maxCount
             }
             else
             {
-                (void)strncpy(dst, src, srcLength);
+                (void)memcpy(dst, src, srcLength);
                 dst[srcLength] = '\0';
                 /*Codes_SRS_CRT_ABSTRACTIONS_99_018: [strncpy_s shall return Zero upon success]*/
                 result = 0;
@@ -160,7 +160,7 @@ int strncpy_s(char* dst, size_t dstSizeInBytes, const char* src, size_t maxCount
                 srcLength = dstSizeInBytes - 1;
                 truncationFlag = 1;
             }
-            (void)strncpy(dst, src, srcLength);
+            (void)memcpy(dst, src, srcLength);
             dst[srcLength] = '\0';
             result = 0;
         }


zolvarga:
@pboettch 
Thank you for contributing to our repo!

It is a good idea to switch to memcpy and in this case it won't cause a problem.

I am accepting the PR, ad will close it when it has been merged.

Best Regards,
Zoltan

zolvarga:
@pboettch 

The PR has been merged.

Best Regards,
Zoltan

Issue #162, opened 2018-06-26
UUID define in uuid.h collides with windows.h


dcristoloveanu:
Including uuid.h and windows.h in the same project results in warnings of redefinition for UUID.
It would be awesome if uuid.h would choose a type name that is not really that widely used as to be present in windows.h.

Thanks!
/Dan

ewertons:
Makes sense. We will try to address that soon.

dcristoloveanu:
This has been addressed and it is merged into master.

az-iot-builder-01:
@dcristoloveanu, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #178, opened 2018-08-28
Rename UUID type to UUID_T to avoid collisions with Windows types

dcristoloveanu:
This addresses the collision reported in #162.
After this is merged in a PR for uamqp will be issued, since that seems to be the only downstream repo that needs it.

### DIFF
--- a/inc/azure_c_shared_utility/uuid.h
+++ b/inc/azure_c_shared_utility/uuid.h
@@ -16,26 +16,26 @@ extern "C" {
 
 #include "azure_c_shared_utility/umock_c_prod.h"
 
-typedef unsigned char UUID[16];
+typedef unsigned char UUID_T[16];
 
 /* @brief               Generates a true UUID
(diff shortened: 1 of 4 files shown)


inc/azure_c_shared_utility/uuid.h
```diff
@@ -16,26 +16,26 @@ extern "C" {
 
 #include "azure_c_shared_utility/umock_c_prod.h"
 
-typedef unsigned char UUID[16];
+typedef unsigned char UUID_T[16];
```
jebrando: >UUID_T [](start = 22, length = 6)

I don't particularly like the _T because it's used so often, but I can't come up with something better, so carry on.

ewertons: To be more similar to other *_t types, "uuid_t" seems like a better fit I think.

But it's pure style choice, your changes are good.

PR #185, opened 2018-09-11
Link uuid library by specific directories.

qiying:
If libuuid.so and uuid.h are under a specific path, CMakeLists.txt should provide variables for setting.


ewertons:
Thanks for your contribution.

This is being replaced by https://github.com/Azure/azure-c-shared-utility/pull/627

-Azure IoT C SDK Team.

PR #198, opened 2018-09-29
Improve const buffer to do only one malloc

dcristoloveanu:
- Improve const buffer to do only one malloc
- Created INIT_VAR/INC_VAR/DEC_VAR macros in the refcount_os implementations for Windows/Linux/Generic. This is in order that we can use this abstraction without being bound to a ref count type from refcount.h. This will also help us in other projects :-).
- Removed some oldies but goldies from the const buffer tests

Having the specs and test suite there was so awesome!

### DIFF
--- a/inc/azure_c_shared_utility/refcount.h
+++ b/inc/azure_c_shared_utility/refcount.h
@@ -63,15 +63,19 @@ static type* REFCOUNT_TYPE_DECLARE_CREATE(type) (void)                          
 #ifndef DEC_RETURN_ZERO
 #error refcount_os.h does not define DEC_RETURN_ZERO
 #endif // !DEC_RETURN_ZERO
-#ifndef INC_REF
-#error refcount_os.h does not define INC_REF
+#ifndef INC_REF_VAR
+#error refcount_os.h does not define INC_REF_VAR
 #endif // !INC_REF
-#ifndef DEC_REF
-#error refcount_os.h does not define DEC_REF
+#ifndef DEC_REF_VAR
+#error refcount_os.h does not define DEC_REF_VAR
 #endif // !DEC_REF
-#ifndef INIT_REF
-#error refcount_os.h does not define INIT_REF
+#ifndef INIT_REF_VAR
+#error refcount_os.h does not define INIT_REF_VAR
 #endif // !INIT_REF
+
+#define INC_REF(type, var) INC_REF_VAR(((REFCOUNT_TYPE(type)*)var)->count)
+#define DEC_REF(type, var) DEC_REF_VAR(((REFCOUNT_TYPE(type)*)var)->count)
+#define INIT_REF(type, var) INIT_REF_VAR(((REFCOUNT_TYPE(type)*)var)->count)
 
 #ifdef __cplusplus
 }
--- a/pal/linux/refcount_os.h
+++ b/pal/linux/refcount_os.h
@@ -57,22 +57,22 @@ gcc
 /*if macro DEC_REF returns DEC_RETURN_ZERO that means the ref count has reached zero.*/
 #if defined(REFCOUNT_ATOMIC_DONTCARE)
 #define DEC_RETURN_ZERO (0)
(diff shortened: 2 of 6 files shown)


src/constbuffer.c
```diff
@@ -14,16 +11,15 @@
 typedef struct CONSTBUFFER_HANDLE_DATA_TAG
 {
     CONSTBUFFER alias;
-}CONSTBUFFER_HANDLE_DATA;
-
-DEFINE_REFCOUNT_TYPE(CONSTBUFFER_HANDLE_DATA);
+    COUNT_TYPE count;
+} CONSTBUFFER_HANDLE_DATA;
```
massand: please add static compile time validation for the existence of count type here, in lieu of DEFINE_REFCOUNT_TYPE. It also helps clarify the requirements of a "REF type" in our code.

pal/linux/refcount_os.h
```diff
@@ -57,22 +57,26 @@ gcc
 /*if macro DEC_REF returns DEC_RETURN_ZERO that means the ref count has reached zero.*/
 #if defined(REFCOUNT_ATOMIC_DONTCARE)
 #define DEC_RETURN_ZERO (0)
+#define INC_REF_VAR(count) ++(count)
+#define DEC_REF_VAR(count) --(count)
+#define INIT_REF_VAR(count) do { count = 1; } while((void)0,0)
+
 #define INC_REF(type, var) ++((((REFCOUNT_TYPE(type)*)var)->count))
```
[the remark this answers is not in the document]
dcristoloveanu: Nope, these are not needed here anymore and in fact they would collide with the defines in refcount.h. Removed.

---
In reply to: [221828317](https://github.com/Azure/azure-c-shared-utility/pull/198#discussion_r221828317) [](ancestors = 221828317)

inc/azure_c_shared_utility/refcount.h
```diff
@@ -63,16 +63,20 @@ static type* REFCOUNT_TYPE_DECLARE_CREATE(type) (void)
 #ifndef DEC_RETURN_ZERO
 #error refcount_os.h does not define DEC_RETURN_ZERO
 #endif // !DEC_RETURN_ZERO
-#ifndef INC_REF
+#ifndef INC_REF_VAR
 #error refcount_os.h does not define INC_REF
```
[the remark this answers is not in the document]
dcristoloveanu: Updated.

---
In reply to: [221829220](https://github.com/Azure/azure-c-shared-utility/pull/198#discussion_r221829220) [](ancestors = 221829220)

src/constbuffer.c
dcristoloveanu: I am not sure I understand this comment.
Are you suggesting havinf an if defined(COUNT_TYPE)?
If so the only thing we could do in case the type is not defined is to #error.

But the compiler would also error if COUNT_TYPE is not defined.

So the check would add no value...

Or maybe you are referring to another type of check. Can you please clarify?

---
In reply to: [221837290](https://github.com/Azure/azure-c-shared-utility/pull/198#discussion_r221837290) [](ancestors = 221837290)

massand: This is a minor suggestion to codify the relationship between the declared type (CONSTBUFFER) and the REF macros. Since we aren''t replacing all instances of REFTYPE_COUNT_CREATE

dcristoloveanu: Is the suggestion to add the following code?

#if defined(COUNT_TYPE)
#error COUNT_TYPE not defined
#endif

We do not need to add this as it does not add any value (compiler detects the error by the fact that COUNT_TYPE does not exist).

CONST_BUFFER uses the COUNT_TYPE and the INC_REF_VAR/DEC_REF_VAR/INIT_REF_VAR macros in order to increment/decrement/init the ref count in a platform agnostic way. It does not use the rest of the macros in refcount which are meant for wrapping a structure in a ref counted type..

Or if your suggestion is another, can you please draft it? Thanks!

---
In reply to: [222053358](https://github.com/Azure/azure-c-shared-utility/pull/198#discussion_r222053358) [](ancestors = 222053358)

PR #199, opened 2018-10-01
have add head for singlylinked list

anporumb:


src/singlylinkedlist.c
```diff
@@ -365,3 +365,44 @@ int singlylinkedlist_foreach(SINGLYLINKEDLIST_HANDLE list, LIST_ACTION_FUNCTION
...
+        result = NULL;
+    }
+    else
+    {
+        result = malloc(sizeof(LIST_ITEM_INSTANCE));
+
+        if (result == NULL)
+        {
+            /*Codes_SRS_LIST_02_003: [ If there are any failures then singlylinkedlist_add_head shall fail and return NULL. ]*/
+            /*return as is*/
```
jebrando: >/*return as is*/ [](start = 12, length = 16)

Please add LogError here so the failure can be known.

src/singlylinkedlist.c
```diff
@@ -365,3 +365,44 @@ int singlylinkedlist_foreach(SINGLYLINKEDLIST_HANDLE list, LIST_ACTION_FUNCTION
...
 }
+
+LIST_ITEM_HANDLE singlylinkedlist_add_head(SINGLYLINKEDLIST_HANDLE list, const void* item)
+{
+    LIST_ITEM_HANDLE result;
+
+    /* Codes_SRS_LIST_02_001: [ If list is NULL then singlylinkedlist_add_head shall fail and return NULL. ]*/
+    if (list == NULL)
+    {
+        LogError("Invalid argument SINGLYLINKEDLIST_HANDLE list=%p, const void* item=%p", list, item);
```
jebrando: >, item [](start = 94, length = 6)

Putting the variable 'item' in the LogError will only add confusion since it doesn't matter if it's NULL or NOT.  Please Remove.

src/singlylinkedlist.c
anporumb: Added LogError

---
In reply to: [221642092](https://github.com/Azure/azure-c-shared-utility/pull/199#discussion_r221642092) [](ancestors = 221642092)

src/singlylinkedlist.c
anporumb: Removed printing item

---
In reply to: [221642887](https://github.com/Azure/azure-c-shared-utility/pull/199#discussion_r221642887) [](ancestors = 221642887)

PR #203, opened 2018-10-10
Use stub for HTTP proxy when SocketIO isn't enabled

BigPeteB:
Fix for Azure/azure-iot-sdk-c#456

### DIFF
--- a/src/http_proxy_stub.c
+++ b/src/http_proxy_stub.c
@@ -1,0 +1,131 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <stdbool.h>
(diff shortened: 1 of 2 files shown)


src/http_proxy_stub.c
```diff
@@ -0,0 +1,131 @@
...
+    (void)http_proxy_io;
+    (void)on_io_open_complete;
+    (void)on_io_open_complete_context;
+    (void)on_bytes_received;
+    (void)on_bytes_received_context;
+    (void)on_io_error;
+    (void)on_io_error_context;
+
+    LogError("Function %s is a stub and should never be called", __FUNCTION__);
+    result = __LINE__;
```
jebrando: >__LINE__ [](start = 13, length = 8)

Can you change to __FAILURE__ for some devices.

massand: In case the markdown threw us off - its `__FAILURE__`

PR #220, opened 2018-10-23
Put VS2017 folders in .gitignore

schultetwin1:
When opening azure-c-shared-utility in VS2017 (using its CMake support),
VS2017 creates a couple folders.

* .vs
* obj folders next to .slns

This adds those folders into the .gitignore file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,10 @@
 /cmake
 /devdoc/~$o_requirements.docm
 /.vscode
+/.vs
 *.jar
 *.aps
 /res
+/testtools/SerialPort/SerialPort/obj/
+/tools/compilembed/obj/
+/tools/macro_utils_h_generator/obj/


.gitignore
```diff
@@ -7,6 +7,10 @@
 /cmake
 /devdoc/~$o_requirements.docm
```
dcristoloveanu: This can also be removed.

.gitignore
```diff
@@ -7,6 +7,10 @@
...
 /.vscode
+/.vs
 *.jar
 *.aps
 /res
+/testtools/SerialPort/SerialPort/obj/
```
dcristoloveanu: >lPort/obj/ [](start = 27, length = 10)

Suggest a generic exclusion of obj folders:

**/obj/*

.gitignore
```diff
@@ -7,6 +7,10 @@
 /cmake
```
dcristoloveanu: Also suggest /cmake* as many times there are variations of the cmake folder being used.

dcristoloveanu:
Cheers for the contributions!
I am one of the admins that can merge PRs, sorry for the delay in replying.

.gitignore
jebrando: @schultetwin1 I'll take care of this one.

PR #223, opened 2018-11-08
Change compileAs to have per target statements

dcristoloveanu:
Change compileAs to have per target statements. This is needed as we have a project that needs to be built with C11. The compileAsC99 that was forced by C shared did not allow for that so the correct way of doing this is actually having the setting per project.


CMakeLists.txt
```diff
@@ -628,3 +628,4 @@ else()
         ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
 endif()
 
+compileTargetAsC99(aziotsharedutil)
```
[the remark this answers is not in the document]
dcristoloveanu: Well, there are 2 different things:
Setting the standard globally (done by CMAKE_C_STANDARD for example) and setting the standard to be used for a given target (PROPERTIES C_STANDARD 11 for example).

The problem with using the global CMAKE settings is that the first one sticks.

That is why I was forced to add the function that does this per project.
So it logically follows that each project should set its own settings for what standard it uses.
Thus this appeared here ...

---
In reply to: [232010290](https://github.com/Azure/azure-c-shared-utility/pull/223#discussion_r232010290) [](ancestors = 232010290)

configs/azure_iot_build_rules.cmake
```diff
@@ -168,7 +168,27 @@ macro(compileAsC11)
   endif()
 endmacro(compileAsC11)
 
-compileAsC99()
```
jebrando: Please leave for backward compatibility

jebrando: talked offline.

---
In reply to: [232019614](https://github.com/Azure/azure-c-shared-utility/pull/223#discussion_r232019614) [](ancestors = 232019614)

PR #225, opened 2018-11-16
Reordered waiting_thread_count++ and unlock so that ++ happen under t…

wes-b:
Reordered waiting_thread_count++ and unlock so that ++ happen under the lock, removed interlocked functions as all touches are now under the lock. Also fixed error checking in Condition_Init as it related to CreateEvent, which should be compared against NULL to detect a failure.

This addresses open issues 189, 195, and 196.

### DIFF
--- a/adapters/condition_win32.c
+++ b/adapters/condition_win32.c
@@ -28,7 +28,7 @@ COND_HANDLE Condition_Init(void)
     {
         cond->event_handle = CreateEvent(NULL, FALSE, FALSE, NULL);
 
-        if (cond->event_handle == INVALID_HANDLE_VALUE)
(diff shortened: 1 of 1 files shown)


adapters/condition_win32.c
```diff
@@ -28,7 +28,7 @@ COND_HANDLE Condition_Init(void)
     {
         cond->event_handle = CreateEvent(NULL, FALSE, FALSE, NULL);
 
-        if (cond->event_handle == INVALID_HANDLE_VALUE)
+        if (cond->event_handle == NULL)
```
dcristoloveanu: >NULL [](start = 34, length = 4)

Thank you for fixing this

PR #232, opened 2018-11-28
Support UNIX domain sockets in Windows RS5

damonbarry:
Windows supports UNIX domain sockets, but only on RS5 or later. So this change includes logic in CMakeLists.txt which adds the compiler definition `AF_UNIX_ON_WINDOWS` when building on RS5 or later. It also makes the Winsock2 adapter for socketio look a lot more like the Berkeley adapter, including support for the `ADDRESS_TYPE` option in `socketio_setoption`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,6 +262,12 @@ if(${use_applessl})
     include_directories(./pal/ios-osx/)
 endif()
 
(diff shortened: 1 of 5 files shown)


damonbarry:
I opened this PR against my fork, but the build system needs a branch on the main repo. I opened #234; closing this one.

PR #233, opened 2018-11-28
Add CONSTBUFFER_CreateWithMoveMemory API in CONSTBUFFER

dcristoloveanu:
Add CONSTBUFFER_CreateWithMoveMemory API in CONSTBUFFER

### DIFF
--- a/tests/constbuffer_ut/constbuffer_ut.c
+++ b/tests/constbuffer_ut/constbuffer_ut.c
@@ -62,7 +62,7 @@ unsigned char* my_BUFFER_u_char(BUFFER_HANDLE handle)
     return result;
 }
 
-size_t my_BUFFER_length(BUFFER_HANDLE handle)
+static size_t my_BUFFER_length(BUFFER_HANDLE handle)
 {
     size_t result;
     if (handle == BUFFER1_HANDLE)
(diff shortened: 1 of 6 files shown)


tests/constbuffer_ut/constbuffer_ut.c
```diff
@@ -342,6 +324,101 @@ BEGIN_TEST_SUITE(constbuffer_unittests)
...
+        CONSTBUFFER_Destroy(handle);
+    }
+
+    /* Tests_SRS_CONSTBUFFER_01_002: [ Otherwise, CONSTBUFFER_CreateWithMoveMemory shall store the source and size and return a non-NULL handle to the newly created const buffer. ]*/
+    TEST_FUNCTION(CONSTBUFFER_CreateWithMoveMemory_with_0_size_succeeds)
+    {
+        ///arrange
+        CONSTBUFFER_HANDLE handle;
+        const CONSTBUFFER* content;
+        unsigned char* test_buffer = my_gballoc_malloc(2);
```
jebrando: >unsigned char* test_buffer = my_gballoc_malloc(2); [](start = 8, length = 50)

I believe you need to free this memory.

dcristoloveanu: This is actually on purpose. The buffer passed to this constructor is owned by CONSTBUFFER. It does not matter that it is NULL or non- NULL. malloc can return a non-NULL pointer for a 0 size allocation. Standard says. Imagine this code:

unsigned char* x = malloc(some_size_passed_as_arg);
CONSTBUFFER_HANDLE b = CONSTBUFFER_CreateWithMoveMemory(x, some_size_passed_as_arg);

If some_size_passed_as_arg is 0 malloc can return a non-NULL pointer and then a buffer can be created with that (with size 0).

Well same buffer needs to be freed by someone (the someone that owns it after the Move constructor is definitely the const buffer.

PR #234, opened 2018-11-29
Support UNIX domain sockets in Windows RS5

damonbarry:
Windows supports UNIX domain sockets, but only on RS5 or later. So this change includes logic in CMakeLists.txt which adds the compiler definition `AF_UNIX_ON_WINDOWS` when building on RS5 or later. It also makes the Winsock2 adapter for socketio look a lot more like the Berkeley adapter, including support for the `ADDRESS_TYPE` option in `socketio_setoption`.

### DIFF
--- a/adapters/socketio_win32.c
+++ b/adapters/socketio_win32.c
@@ -8,11 +8,15 @@
 #include <ws2tcpip.h>
 #include <windows.h>
 #include <mstcpip.h>
+#ifdef AF_UNIX_ON_WINDOWS
+#include <afunix.h>
+#endif
 #include "azure_c_shared_utility/socketio.h"
 #include "azure_c_shared_utility/singlylinkedlist.h"
 #include "azure_c_shared_utility/gballoc.h"
 #include "azure_c_shared_utility/gbnetwork.h"
 #include "azure_c_shared_utility/optimize_size.h"
+#include "azure_c_shared_utility/shared_util_options.h"
 #include "azure_c_shared_utility/xlogging.h"
 
 typedef enum IO_STATE_TAG
@@ -35,6 +39,7 @@ typedef struct PENDING_SOCKET_IO_TAG
 typedef struct SOCKET_IO_INSTANCE_TAG
 {
     SOCKET socket;
+    SOCKETIO_ADDRESS_TYPE address_type;
     ON_BYTES_RECEIVED on_bytes_received;
     ON_IO_ERROR on_io_error;
     void* on_bytes_received_context;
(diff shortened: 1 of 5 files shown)


damonbarry:
@jspaith I moved the parsing logic into socketio_win32 as discussed. Please review...

adapters/socketio_win32.c
```diff
@@ -245,6 +251,94 @@ void socketio_destroy(CONCRETE_IO_HANDLE socket_io)
...
+    {
+#endif
+        char portString[16];
+        ADDRINFO* addr_info = NULL;
+        ADDRINFO addrHint = { 0 };
+        addrHint.ai_family = AF_INET;
+        addrHint.ai_socktype = SOCK_STREAM;
+        addrHint.ai_protocol = 0;
+
+        sprintf(portString, "%u", socket_io_instance->port);
```
dcristoloveanu: >sprintf [](start = 8, length = 7)

This can fail

adapters/socketio_win32.c
```diff
@@ -245,6 +251,94 @@ void socketio_destroy(CONCRETE_IO_HANDLE socket_io)
...
+        if (getaddrinfo(socket_io_instance->hostname, portString, &addrHint, &addr_info) != 0)
+        {
+            LogError("Failure: getaddrinfo failure %d.", WSAGetLastError());
+            result = __FAILURE__;
+        }
+        else
+        {
+            result = connect_socket(socket_io_instance->socket, addr_info->ai_addr, sizeof(*addr_info->ai_addr));
```
dcristoloveanu: >connect_socket [](start = 21, length = 14)

Don't you want to log the error when this fails?

adapters/socketio_win32.c
```diff
@@ -245,6 +251,94 @@ void socketio_destroy(CONCRETE_IO_HANDLE socket_io)
...
+        }
+
+        if (path_len + 1 > sizeof(addr_un.sun_path))
+        {
+            LogError("Path '%s' is too long for a unix socket (max len = %zu)", path, sizeof(addr_un.sun_path));
+            result = __FAILURE__;
+        }
+        else
+        {
+            memset(&addr_un, 0, sizeof(addr_un));
```
dcristoloveanu: >memset [](start = 12, length = 6)

(void)

adapters/socketio_win32.c
damonbarry: connect_socket is an internal function, and it logs its own failures.

adapters/socketio_win32.c
damonbarry: Will do

adapters/socketio_win32.c
damonbarry: I'll add the error handler for this

damonbarry: Fixed

adapters/socketio_win32.c
damonbarry: Fixed

PR #238, opened 2018-12-05
Adding a Buffer create with size function for mqtt

jebrando:


tests/buffer_ut/buffer_ut.c
```diff
@@ -1561,6 +1558,81 @@ BEGIN_TEST_SUITE(Buffer_UnitTests)
         BUFFER_delete(res);
     }
 
+    TEST_FUNCTION(BUFFER_create_with_size_succeeds)
```
dcristoloveanu: >TEST_FUNCTION(BUFFER_create_with_size_succeeds) [](start = 4, length = 47)

Missing specs and tags in the code.
Since we're partying together on this code, could you please add them?

BUFFER does have specs/tests code so let's keep the bar up. "When in Rome"

src/buffer.c
```diff
@@ -91,6 +91,36 @@ BUFFER_HANDLE BUFFER_create(const unsigned char* source, size_t size)
     return (BUFFER_HANDLE)result;
 }
 
+BUFFER_HANDLE BUFFER_create_with_size(size_t buff_size)
+{
+    BUFFER* result;
+    if (buff_size == 0)
```
dcristoloveanu: >  if (buff_size == 0) [](start = 2, length = 21)

malloc allows zero sizes, BUFFER_create also does, so this should also allow it.

dcristoloveanu:
Waiting for specs, otherwise looks OK.
We could think that one could obtain the same by creating an empty buffer and then enlarging, but I am with you, I like having to call only one API, makes things easier.

devdoc/buffer_requirements.md
```diff
@@ -49,7 +52,26 @@ BUFFER_create creates a new buffer from the memory at source, having size "size"
...
+
+
+
+```c
+extern BUFFER_HANDLE BUFFER_create_with_size(size_t buff_size)
+```
+
+**SRS_BUFFER_07_029: [** BUFFER_create_with_size shall create a BUFFER_HANDLE with a pre allocated underlying buffer size. **]**
+
+**SRS_BUFFER_07_030: [** If buff_size is 0 BUFFER_create_with_size shall return NULL. **]**
```
dcristoloveanu: >buff_size is 0 BUFFER_create_with_size shall return NULL. [](start = 28, length = 57)

This should not return NULL. It should be possible to allocate a buffer with size 0.

For the same reason that malloc allows it. Here goes.

During writing SW, sometimes one would write stuff like:

size_t payload_size = get_my_payload_size(); // for example read it from a packet on the wire

BUFFER_HANDLE payload_buffer = BUFFER_create_with_size(payload_size);

Notice how the code does not need to differentiate between a zero size and non zero size payload. That is neat because why should we litter the callers of malloc/BUFFER_create_with_size with 0 checks? 0 or positive payload size are both valid cases.

Can you please address this? otherwise I fear that assumptions can be easily made that BUFFER_create_with_size will work with 0 and it could potentially introduce bugs down the road.

And now it is the time to change it as it is very cheap (we're in the code :-)).

dcristoloveanu:
Thanks for adding the specs.

---
In reply to: [444617227](https://github.com/Azure/azure-c-shared-utility/pull/238#issuecomment-444617227) [](ancestors = 444617227)

tests/buffer_ut/buffer_ut.c
```diff
@@ -1561,6 +1558,87 @@ BEGIN_TEST_SUITE(Buffer_UnitTests)
         BUFFER_delete(res);
     }
 
+    // Codes_SRS_BUFFER_07_029: [ BUFFER_create_with_size shall create a BUFFER_HANDLE with a pre allocated underlying buffer size.]
```
dcristoloveanu: >Codes [](start = 7, length = 5)

Tests

dcristoloveanu: Here and everywhere.

---
In reply to: [239323121](https://github.com/Azure/azure-c-shared-utility/pull/238#discussion_r239323121) [](ancestors = 239323121)

PR #259, opened 2019-01-17
Move memory_data to C shared utility

dcristoloveanu:
Move memory_data to C shared utility

### DIFF
--- a/src/memory_data.c
+++ b/src/memory_data.c
@@ -1,0 +1,169 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include <inttypes.h>
+
+#include "azure_c_shared_utility/uuid.h"
+
+#include "azure_c_shared_utility/memory_data.h"
+
+void read_uint8_t(const unsigned char* source, uint8_t* destination)
+{
+    /*Codes_SRS_MEMORY_DATA_02_041: [ read_uint8_t shall write in destination the byte at source ]*/
+    *destination = *source;
+}
+
+void read_uint16_t(const unsigned char* source, uint16_t* destination)
+{
+    /*Codes_SRS_MEMORY_DATA_02_042: [ read_uint16_t shall write in destination the bytes at source MSB first and return. ]*/
+    *destination = 
+        (source[0]<<8)+
+        (source[1])
+        ;
+}
+
+void read_uint32_t(const unsigned char* source, uint32_t* destination)
+{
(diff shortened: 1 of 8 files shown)


src/memory_data.c
```diff
@@ -0,0 +1,170 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include <inttypes.h>
+
+#include "azure_c_shared_utility/xlogging.h"
```
jebrando: Not needed.

src/memory_data.c
```diff
@@ -0,0 +1,170 @@
...
+{
+    /*Codes_SRS_MEMORY_DATA_02_041: [ read_uint8_t shall write in destination the byte at source ]*/
+    *destination = *source;
+}
+
+void read_uint16_t(const unsigned char* source, uint16_t* destination)
+{
+    /*Codes_SRS_MEMORY_DATA_02_042: [ read_uint16_t shall write in destination the bytes at source MSB first and return. ]*/
+    *destination = 
+        (source[0]<<8)+
```
jebrando: >source [](start = 9, length = 6)

I'm not sure why you're not checking the source here.  If someone pass NULL in any of these function you will get a Segfault.  I'm assuming you did this for a reason.

src/memory_data.c
dcristoloveanu: Removed

---
In reply to: [248874282](https://github.com/Azure/azure-c-shared-utility/pull/259#discussion_r248874282) [](ancestors = 248874282)

src/memory_data.c
dcristoloveanu: These functions are like memcpy/strcpy/etc in the C standard. They do not check for NULL. NULL is a special address which most of the times is used for signalling not initialized, but otherwise it is no different than address 0x42.

Imagine a case where in a very low level SW (like on a microcontroller in an init routine someone would actually want to write something at address 0. That would actually be valid.

NULL checks do make sense IMO when checking whether a handle has been initialized or not. But for functions which are supposed to directly manipulate memory, I think it's free for all on what address can be used. Like the C standard functions also behave.

---
In reply to: [248874615](https://github.com/Azure/azure-c-shared-utility/pull/259#discussion_r248874615) [](ancestors = 248874615)

jebrando: Can you please put a Spec note or something to state your case on this.  Someone who comes upon this module may be confused (as I am) about why this is doing no checking.

---
In reply to: [248929863](https://github.com/Azure/azure-c-shared-utility/pull/259#discussion_r248929863) [](ancestors = 248929863,248874615)

PR #182, opened 2018-09-03
MOD: %zu->%lu to mbed-related files

SeanZhang0920:
As %zu cannot be shown correctly on mbed os5, I modified %zu to %lu in some mbed-os5-related files.

PR #260, opened 2019-01-18
MOD: %zu->%lu to mbed-related files

ArthurMa1978:
As %zu cannot be shown correctly on mbed os5, I modified %zu to %lu in some mbed-os5-related files.

This is a Re-PR of https://github.com/Azure/azure-c-shared-utility/pull/182


ArthurMa1978:
Compare the old one the change is explicitly casting to (unsigned long) according to the comment

PR #262, opened 2019-01-25
Change constbuffer from CLONE to INC_REF/DEC_REF

dcristoloveanu:
Change constbuffer from CLONE to INC_REF/DEC_REF.

The motivation is that CONSTBUFFER does not allocate new resources for a clone, but in fact by the module contract does an inc_ref. And thus it is lighter on the user module to have an inc_ref/dec_ref API (less error checking and there is no need to treat the result as a potentially completely different handle.

### DIFF
--- a/devdoc/constbuffer_requirements.md
+++ b/devdoc/constbuffer_requirements.md
@@ -36,11 +36,11 @@ MOCKABLE_FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_CreateWithMoveMemory, unsign
 
 MOCKABLE_FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_CreateWithCustomFree, const unsigned char*, source, size_t, size, CONSTBUFFER_CUSTOM_FREE_FUNC, customFreeFunc, void*, customFreeFuncContext);
 
-MOCKABLE_FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_Clone, CONSTBUFFER_HANDLE, constbufferHandle);
+MOCKABLE_FUNCTION(, void, CONSTBUFFER_IncRef, CONSTBUFFER_HANDLE, constbufferHandle);
+
+MOCKABLE_FUNCTION(, void, CONSTBUFFER_DecRef, CONSTBUFFER_HANDLE, constbufferHandle);
 
 MOCKABLE_FUNCTION(, const CONSTBUFFER*, CONSTBUFFER_GetContent, CONSTBUFFER_HANDLE, constbufferHandle);
-
-MOCKABLE_FUNCTION(, void, CONSTBUFFER_Destroy, CONSTBUFFER_HANDLE, constbufferHandle);
 ```
 
 ###  CONSTBUFFER_Create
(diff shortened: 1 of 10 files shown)


devdoc/constbuffer_requirements.md
```diff
@@ -36,11 +36,11 @@ MOCKABLE_FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_CreateWithMoveMemory, unsign
 
 MOCKABLE_FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_CreateWithCustomFree, const unsigned char*, source, size_t, size, CONSTBUFFER_CUSTOM_FREE_FUNC, customFreeFunc, void*, customFreeFuncContext);
 
-MOCKABLE_FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_Clone, CONSTBUFFER_HANDLE, constbufferHandle);
```
jebrando: Why would you remove the clone API, this will require changes to the amqp transport that I'm sure people don't want to go through.  It would seem that you can keep the clone function  and also add the other IncRef and DecRef functions.

devdoc/constbuffer_requirements.md
```diff
@@ -36,11 +36,11 @@ MOCKABLE_FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_CreateWithMoveMemory, unsign
...
+MOCKABLE_FUNCTION(, void, CONSTBUFFER_IncRef, CONSTBUFFER_HANDLE, constbufferHandle);
 
-MOCKABLE_FUNCTION(, const CONSTBUFFER*, CONSTBUFFER_GetContent, CONSTBUFFER_HANDLE, constbufferHandle);
+MOCKABLE_FUNCTION(, void, CONSTBUFFER_DecRef, CONSTBUFFER_HANDLE, constbufferHandle);
 
-MOCKABLE_FUNCTION(, void, CONSTBUFFER_Destroy, CONSTBUFFER_HANDLE, constbufferHandle);
```
jebrando: >CONSTBUFFER_Destroy [](start = 26, length = 19)

Removing destroy is probably also not the most desirable.  I don't have a horse in the race on this one, but it is causing cascading changes through the repo.

devdoc/constbuffer_requirements.md
dcristoloveanu: Because Clone does not copy the contents. Semantically this is inc_ref. And since it is inc_ref then why not have it as such? Having clone that actually does inc_ref forces writing of code that checks the handle returned by Clone, but in reality that should never be checked, nor should it fail when Clone is called with a non-NULL handle.

The changes in the users are trivial (as seen in const buffer array which is a user of it).

---
In reply to: [251143791](https://github.com/Azure/azure-c-shared-utility/pull/262#discussion_r251143791) [](ancestors = 251143791)

devdoc/constbuffer_requirements.md
dcristoloveanu: I sense fear of change. But fear of change should always be overcome with the desire to progress :-).

---
In reply to: [251144218](https://github.com/Azure/azure-c-shared-utility/pull/262#discussion_r251144218) [](ancestors = 251144218)

dcristoloveanu: Technically Destroy is DecRef. Any ref counted object can only have DecRef, because noone can explictly free the memory for the object. A user can only "decrement ref count and if 0 then please free memory". Thus DecRef is the appropriate thing to have for an object that has IncRef.

And this is really a simple rename that should not panic anyone.

---
In reply to: [251150747](https://github.com/Azure/azure-c-shared-utility/pull/262#discussion_r251150747) [](ancestors = 251150747,251144218)

dcristoloveanu:
Hi Ewerton, just wiritng down my thoughts so we can discuss on Monday in person.

Regarding changes to depending modules: IoT SDK for C which I assume you are concerned about has exactly 1 usage of this. If you wish I can make the change for you so you do not need to bother about it.

As for whether it makes sense or not from a design point of view:
IncRef/DecRef explictly state that a new handle is not created. IncRef will not fail if the handle is non-NULL and thus the user code that calls IncRef does not need to check for any errors. It is lighter weight on the user code.

Clone generally implies that a new handle is created (a new entity is created that perfectly mirrors the original). Because this is a const object Clone does not create a new handle and uses internally ref count. But the user code of Clone cannot know that and has to assume that maybe the clone result handle is different and also has to check for success. This is heavier on the user code of Clone.

Looking forward to the discussion on Monday.
Cheers,
/Dan

PR #268, opened 2019-02-11
Add SQM ID to platform info string

jadeiceman:
Adding platform_get_platform_info_with_id to Win32 platform API so that SQM ID can be included in the product info.

### DIFF
--- a/inc/azure_c_shared_utility/platform.h
+++ b/inc/azure_c_shared_utility/platform.h
@@ -16,6 +16,9 @@ extern "C" {
     MOCKABLE_FUNCTION(, void, platform_deinit);
     MOCKABLE_FUNCTION(, const IO_INTERFACE_DESCRIPTION*, platform_get_default_tlsio);
     MOCKABLE_FUNCTION(, STRING_HANDLE, platform_get_platform_info);
+#ifdef WIN32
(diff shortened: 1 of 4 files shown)


inc/azure_c_shared_utility/platform.h
```diff
@@ -16,6 +16,9 @@ extern "C" {
     MOCKABLE_FUNCTION(, void, platform_deinit);
     MOCKABLE_FUNCTION(, const IO_INTERFACE_DESCRIPTION*, platform_get_default_tlsio);
     MOCKABLE_FUNCTION(, STRING_HANDLE, platform_get_platform_info);
+#ifdef WIN32
```
massand: Please remove checks for WIN32 preprocessor within the headers. We like to keep things consistent in headers and have each platform specific header provide the appopriate value for the platform_info.

samples/iot_c_utility/iot_c_utility.c
```diff
@@ -59,6 +59,15 @@ static void show_platform_info()
         (void)printf("%s\r\n", STRING_c_str(platform_info));
         STRING_delete(platform_info);
     }
+
+#ifdef WIN32
```
massand: Same comment here - we like samples to remain platform agnostic.

PR #269, opened 2019-02-12
Comment that OPTION_TLS_VERSION is deprecated

jspaith:
Attempt to steer users from OPTION_TLS_VERSION.


massand:
@jspaith - thanks! Please make the prov_client change in the other repo to remove the explicit setting of OPTION_TLS_VERSION there, since we always default to 1.2 for openSSL

PR #274, opened 2019-02-20
Allow openssl multi-init

EA007:
Allow openssl multi-init, for scenarios where the caller of a component using openssl can also use openssl.

### DIFF
--- a/adapters/tlsio_openssl.c
+++ b/adapters/tlsio_openssl.c
@@ -390,7 +390,7 @@ static const IO_INTERFACE_DESCRIPTION tlsio_openssl_interface_description =
 };
 
 static LOCK_HANDLE * openssl_locks = NULL;
-static int locks_init_count = 0;
+static int locks_install_count = 0;
 static int openssl_init_count = 0;
 
 static void openssl_lock_unlock_helper(LOCK_HANDLE lock, int lock_mode, const char* file, int line)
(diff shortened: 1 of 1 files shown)


adapters/tlsio_openssl.c
```diff
@@ -515,19 +516,26 @@ static void openssl_static_locks_uninstall(void)
 {
     if (openssl_locks != NULL)
     {
-        int i;
-        CRYPTO_set_locking_callback(NULL);
-
-        for (i = 0; i < CRYPTO_num_locks(); i++)
+        if (locks_init_count <= 0)
```
dcristoloveanu: ><= [](start = 29, length = 2)

I think < 0 is an error actually showing unbalanced install/uninstall calls.
So I think there should be 3 possible cases treated here:
< 0 => error
== 0 => cleanup
> 0 => decrement the nesting counter

dcristoloveanu:
Generally change looks good, I left some minor comments to address.

adapters/tlsio_openssl.c
```diff
@@ -390,7 +390,8 @@ static const IO_INTERFACE_DESCRIPTION tlsio_openssl_interface_description =
 };
 
 static LOCK_HANDLE * openssl_locks = NULL;
-
+static int locks_init_count = 0;
```
dcristoloveanu: >locks_init_count [](start = 11, length = 16)

Nitpick: Consider renaming to locks_install_count

adapters/tlsio_openssl.c
```diff
@@ -390,7 +390,8 @@ static const IO_INTERFACE_DESCRIPTION tlsio_openssl_interface_description =
...
+static int openssl_init_count = 0;
```
dcristoloveanu: >openssl_init_count [](start = 11, length = 18)

When I reached this line, I realized that my initial statement of the fact that openssl_static_locks_install should not error when multiple calls to init are made is wrong.

I understand the scenario is that the EH SDK is not the only user of openssl and some other users of openssl might initialize openssl also.

But since you are hitting the error in openssl_static_locks_install when openssl_locks is non-NULL, I understand that there are at least 2 calls to tlsio_openssl_init. I think that should not be the case. tlsio_openssl_init is called by platform_init IIRC and that is meant as a one time init for the platform, no nesting supported at the moment.

Nesting can be added, but the question is whether it is needed that platform_init is called multiple times.

If nesting is truly needed, the change looks generally OK. But I'd like to understand the multiple tlsio_openssl_init calls before merging the fix as this area has proven to be tricky in the past.

EA007: Since there is no coordination between the two openssl clients, and they just happen to both use openssl from the same executable but without any specific mutually agreed timeline or sequence, it appears that only way to allow that to happen is to allow each client to call platform_init and platform_deinit based on their own business logic.

adapters/tlsio_openssl.c
EA007: Fixed.

adapters/tlsio_openssl.c
EA007: Fixed.

ericwol-msft:
Elena, this issue is pretty old. We are cleaning this up. Is this still an issue for you?

EA007:
This is no longer needed. Thanks for checking!

-Elena Apreutesei

PR #275, opened 2019-02-21
Rename macro_utils DEC to CS_DEC (fix Arduino name collision)

ewertons:
This is one of the fixes for Arduino builds.
Related to https://github.com/Azure/azure-iot-sdk-c/issues/795


dcristoloveanu:
> /* and all of these copies must be located and replaced. */

The fact that these things are out of sync with the real macro_utils is a problem.
So let's fix it in the right way by doing as originally suggested by YoDaMa and have the PR to the macro utils repo ([https://github.com/Azure/azure-macro-utils-c](https://github.com/Azure/azure-macro-utils-c)).

---
Refers to: inc/azure_c_shared_utility/macro_utils.h:10 in 64ec0a6. [](commit_id = 64ec0a647facdb7e0a157e4054d463c0b8180917, deletion_comment = True)

dcristoloveanu:
> #define MU_TOSTRING_(x) #x

Especially if you're adding the prefix all over the place, this the divergence is getting really out of hand.

---
Refers to: inc/azure_c_shared_utility/macro_utils.h:17 in 64ec0a6. [](commit_id = 64ec0a647facdb7e0a157e4054d463c0b8180917, deletion_comment = False)

ewertons:
Too many changes have been made since this was approved. I'll resubmit this PR.

PR #278, opened 2019-02-27
Submodule restructure as to not force to have to do a recursive update

jebrando:


CMakeLists.txt
```diff
@@ -551,9 +548,22 @@ set(run_e2e_tests OFF)
...
-    add_subdirectory(testtools)
+if (NOT TARGET umock_c)
+    add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/deps/umock-c)
+endif()
+include_directories(${UMOCK_C_INC_FOLDER})
+
+if (${original_run_unittests} OR ${original_run_int_tests})
+    #Clone the test submodules if needed
+    if (NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/deps/testrunnerswitcher/CMakeLists.txt")
+        execute_process(COMMAND git clone https://github.com/Azure/azure-c-testrunnerswitcher.git -q ${CMAKE_CURRENT_SOURCE_DIR}/deps/testrunnerswitcher)
```
ewertons: With the disadvantage that by not having a fixed pointer to the submodule commit id, this will get the latest testrsw from master, and any breaking changes will be forced to be consumed right away.
Not a point to change or anything, just mentioning here to be aware.

CMakeLists.txt
```diff
@@ -551,9 +548,22 @@ set(run_e2e_tests OFF)
...
+    endif()
+    if (NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/deps/ctest/CMakeLists.txt")
+        execute_process(COMMAND git clone https://github.com/Azure/azure-ctest.git -q ${CMAKE_CURRENT_SOURCE_DIR}/deps/ctest)
+    endif()
```
ewertons: Same here.

configs/azure_iot_external_pal_unit_test_setup.cmake
```diff
@@ -10,9 +10,9 @@ set(ignore_me ${use_wsio})
 include("${CMAKE_CURRENT_LIST_DIR}/azure_iot_build_rules.cmake")
 include("${SHARED_UTIL_FOLDER}/dependencies-test.cmake")
 
-add_subdirectory(${SHARED_UTIL_FOLDER}/testtools/ctest)
-add_subdirectory(${SHARED_UTIL_FOLDER}/testtools/testrunner)
-add_subdirectory(${SHARED_UTIL_FOLDER}/testtools/umock-c)
+add_subdirectory(${SHARED_UTIL_FOLDER}/deps/ctest)
```
ewertons: Very nice. Thanks for this.

CMakeLists.txt
ewertons: This means that any breaking changes in these repos will break the c-sdk gated builds right away.

jebrando: You are correct.  Since they're testing repos we don't have breaking changes that much.

---
In reply to: [261692130](https://github.com/Azure/azure-c-shared-utility/pull/278#discussion_r261692130) [](ancestors = 261692130)

PR #280, opened 2019-03-01
Add SQM ID to platform_get_platform_info string

jadeiceman:
Adding an options parameter to platform_get_platform_info to add ability to include more information in returned string.  PLATFORM_OPTION_RETRIEVE_SQM adds the SQM ID to the string for Win32.


adapters/platform_win32.c
```diff
@@ -96,21 +155,43 @@ STRING_HANDLE platform_get_platform_info(void)
...
+        result = STRING_construct_sprintf("(native; WindowsProduct:Windows NT %d.%d; %s", LOBYTE(LOWORD(dwVersion)), HIBYTE(LOWORD(dwVersion)), arch);
     }
 #pragma warning(default:4996)
 
     if (result == NULL)
     {
         LogError("STRING_construct_sprintf failed");
     }
+
+    if (options & PLATFORM_OPTION_RETRIEVE_SQM)
```
ewertons: If STRING_construct has failed, this block should not be executed. The if-else blocks need to be rearranged to avoid that.

adapters/platform_esp8266.c
```diff
@@ -29,6 +29,15 @@ const IO_INTERFACE_DESCRIPTION* platform_get_default_tlsio(void)
     return tlsio_openssl_get_interface_description();
 }
 
+STRING_HANDLE platform_get_platform_info(uint32_t options)
```
ewertons: Same comment as bellow for all the "uint32_t" new options.

adapters/platform_win32.c
```diff
@@ -54,10 +54,69 @@ const IO_INTERFACE_DESCRIPTION* platform_get_default_tlsio(void)
...
 }
 
-STRING_HANDLE platform_get_platform_info(void)
+static char* get_win_sqm_info(void)
 {
-    // Expected format: "(<runtime name>; <operating system name>; <platform>)"
+    char* result;
+    LONG openKey;
+    DWORD options = 0;
+    HKEY openResult;
```
ewertons: What are the headers for these windows-specific types?
They should be explicitly included in this code file.

adapters/platform_win32.c
```diff
@@ -54,10 +54,69 @@ const IO_INTERFACE_DESCRIPTION* platform_get_default_tlsio(void)
...
+    LONG getRegValue = ERROR_INVALID_HANDLE;
+    DWORD dataType;
+    DWORD size = GUID_LENGTH;
+    PVOID pvData;
+
+    // SQM values are guids int he system, we will allocate enough space to hold that
+    if ((result = (char*)malloc(GUID_LENGTH)) == NULL)
+    {
+        LogError("Failure allocating sqm info");
+        result = NULL;
```
ewertons: Not needed.

adapters/platform_win32.c
```diff
@@ -96,21 +155,43 @@ STRING_HANDLE platform_get_platform_info(void)
...
         LogError("STRING_construct_sprintf failed");
     }
+
+    if (options & PLATFORM_OPTION_RETRIEVE_SQM)
+    {
+        // Failure here should continue
+        char* sqm_info = get_win_sqm_info();
+        if (sqm_info != NULL)
+        {
+            if (STRING_sprintf(result, "; %s)", sqm_info) != 0)
```
ewertons: Usually when there is a failure in the SDK we have been destroying whatever data that has been put together before in the function, and returning NULL.
Wondering which way would be preferred here. Looks like the way implemented here is better, but what Yi has to comment on this? Is partial info really better than no info for you?

samples/iot_c_utility/iot_c_utility.c
```diff
@@ -53,7 +53,7 @@ static void show_sastoken_example()
 
 static void show_platform_info()
 {
-    STRING_HANDLE platform_info = platform_get_platform_info();
+    STRING_HANDLE platform_info = platform_get_platform_info(PLATFORM_OPTION_RETRIEVE_SQM);
```
ewertons: What is the reason to expose this option to the user?
Can it be just assumed that if the application is running on Windows it will always try to get the SQM info?

jadeiceman: There's a requirement for the SQM ID to not be included when using HTTP in the Azure C SDK to limit the size of the string in the request.  It can be assumed that if the app is running on Windows it'll always try to get the SQM ID, just not when the transport protocol is HTTP.  Exposing this option will also allow us to more easily modify the contents of the returned platform info in the future if need be.

adapters/platform_win32.c
jadeiceman: Yes, SQM ID is optional - if it exists we'd like to have it so we can join it with other telemetry we collect from Windows.  If it doesn't exist, it should at least return the default values in the string.  It's not a hard requirement that SQM ID needs to be there.

inc/azure_c_shared_utility/platform.h
```diff
@@ -8,14 +8,20 @@
...
+#include <cstdint>
+#else
+#include <stdint.h>
 #endif /* __cplusplus */
 
     MOCKABLE_FUNCTION(, int, platform_init);
     MOCKABLE_FUNCTION(, void, platform_deinit);
     MOCKABLE_FUNCTION(, const IO_INTERFACE_DESCRIPTION*, platform_get_default_tlsio);
-    MOCKABLE_FUNCTION(, STRING_HANDLE, platform_get_platform_info);
+    MOCKABLE_FUNCTION(, STRING_HANDLE, platform_get_platform_info, uint32_t, options);
```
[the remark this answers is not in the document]
jadeiceman: Makes sense, I'll change this.

adapters/platform_win32.c
jadeiceman: But it looks like there's an issue here since the parentheses aren't closed if the SQM ID isn't found.  I'll fix that.

jadeiceman:
@ewertons Thanks for the review!  I've made the requested changes, please let me know if there's anything else that needs to be addressed.

adapters/platform_linux.c
```diff
@@ -47,8 +47,10 @@ const IO_INTERFACE_DESCRIPTION* platform_get_default_tlsio(void)
 #endif
 }
 
-STRING_HANDLE platform_get_platform_info(void)
+STRING_HANDLE platform_get_platform_info(PLATFORM_INFO_OPTION options)
 {
+    (void)options;
```
YoDaMa: could you add a brief comment why we are passing options to each platform even though we are not setting any options here. It makes sense in the PR but in a year I can see someone look at this and being slightly confused. Just something along the lines of: // no valid options so ignoring options parameter

adapters/platform_mbed.cpp
```diff
@@ -64,8 +64,10 @@ const IO_INTERFACE_DESCRIPTION* platform_get_default_tlsio(void)
     return tlsio_wolfssl_get_interface_description();
 }
 
-STRING_HANDLE platform_get_platform_info(void)
+STRING_HANDLE platform_get_platform_info(PLATFORM_INFO_OPTION options)
 {
+    (void)options;
```
YoDaMa: same thing for each platform file where we are now passing in options but doing nothing with it.

adapters/platform_linux.c
jadeiceman: Sure, will do.

PR #282, opened 2019-03-04
Change lock_win32 implementation from semaphore to SRWLOCK

mattdurak:
SRWLOCK implementation is more performant on Windows than Semaphores.
Also means that Lock and Unlock on Windows always return LOCK_OK.


dcristoloveanu:
> LOCK_HANDLE Lock_Init(void)

Now if I say we should rename this to Lock_Creaate and Lock_Deinit to Lock_Destroy I think I would make many enemies. So don't :-).

---
Refers to: adapters/lock_win32.c:11 in 1516868. [](commit_id = 151686876b94e93c0592f72fe78956e54ca54605, deletion_comment = False)

adapters/lock_win32.c
```diff
@@ -12,11 +12,12 @@ LOCK_HANDLE Lock_Init(void)
...
     /* Codes_SRS_LOCK_10_002: [Lock_Init on success shall return a valid lock handle which should be a non NULL value] */
     /* Codes_SRS_LOCK_10_003: [Lock_Init on error shall return NULL ] */
-    HANDLE result = CreateSemaphoreW(NULL, 1, 1, NULL);
+    SRWLOCK* result = malloc(sizeof(SRWLOCK));
     if (result == NULL)
     {
-        LogError("CreateSemaphore failed.");
+        LogError("Allocate SRWLOCK failed.");
     }
+    InitializeSRWLock(result);
```
dcristoloveanu: >InitializeSRWLock(result); [](start = 4, length = 26)

This should only be executed when result is non-NULL.

dcristoloveanu: And thus there is some missing test somewhere.

---
In reply to: [262592173](https://github.com/Azure/azure-c-shared-utility/pull/282#discussion_r262592173) [](ancestors = 262592173)

mattdurak: That is true, whoops!
I'll add back the WIN32 section of tests and cover this with the mocked malloc/free

---
In reply to: [262592291](https://github.com/Azure/azure-c-shared-utility/pull/282#discussion_r262592291) [](ancestors = 262592291,262592173)

mattdurak: Also the "lock_ut" tests are really more "lock_int" tests because they don't mock anything (since different implementations would need different mocks).

---
In reply to: [262637800](https://github.com/Azure/azure-c-shared-utility/pull/282#discussion_r262637800) [](ancestors = 262637800,262592291,262592173)

dcristoloveanu: Yep they are int. At the time this was written the int section of tests did not even exist. :-)

---
In reply to: [262643377](https://github.com/Azure/azure-c-shared-utility/pull/282#discussion_r262643377) [](ancestors = 262643377,262637800,262592291,262592173)

PR #292, opened 2019-04-04
Add BearSSL TLS adapter layer

markrad:
Adds support for BearSSL TLS library. Primarily this is being added for ESP8266 that are being using the Arduino environment to build IoT software based upon our C SDK. This TLS library is shipped with the ESP8266 board support package so it makes sense to support it. Performs full server certificate verification and supports X.509 authentication.

Build system has also been updated to allow the BearSSL TLS adapter to be used on a standard Linux platform should a customer desire. It is very light weight and a good candidate for low resource Linux devices too.


massand:
@markrad - could you please work with @ewertons to add a quick e2e validation for BearSSL on Linux in our lab. It will help ensure we don't break it going forward. This will likely be written within the azure-iot-sdk-c repo. You can post a branch and link the PR here. Thanks

tests/tlsio_bearssl_ut/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+#Copyright (c) Microsoft. All rights reserved.
+#Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+cmake_minimum_required(VERSION 2.8.11)
+
+set(theseTestsName tlsio_bearssl_ut)
+
+#set(MBED_ROOT_DIR $ENV{MBED_ROOT_DIR} CACHE PATH "")
+set(BEARSSL_ROOT_DIR "C:/Users/markrad/S/BearSSL")
```
jbobotek: This should be a relative path ala $ENV{OpenSSLDir} CACHE PATH, etc.

tests/tlsio_bearssl_ut/tlsio_bearssl_ut.c
```diff
@@ -0,0 +1,972 @@
...
+typedef void(*f_append_dn)(void *, const void *, size_t);
+MOCKABLE_FUNCTION(, void, br_x509_decoder_init, br_x509_decoder_context*, ctx, f_append_dn, append_dn, void*, append_dn_ctx);
+MOCKABLE_FUNCTION(, void, br_x509_decoder_push, br_x509_decoder_context*, ctx, const void*, data, size_t, len);
+
+MOCKABLE_FUNCTION(, void, on_io_open_complete, void*, context, IO_OPEN_RESULT, open_result);
+MOCKABLE_FUNCTION(, void, on_bytes_received, void*, context, const unsigned char*, buffer, size_t, size);
+MOCKABLE_FUNCTION(, void, on_io_error, void*, context);
+MOCKABLE_FUNCTION(, void, on_io_close_complete, void*, context);
+MOCKABLE_FUNCTION(, void, on_send_complete, void*, context, IO_SEND_RESULT, send_result);
+
```
jbobotek: Do we really need to unit test every bear ssl function? I would imagine the author does that. We should probably narrow this to test functions associated with the adapter layer.

markrad: This is not testing any part of BearSSL. All of the BearSSL functions are mocked so the actual library is not being called at all. What this does do though is that it sets the expectation of the sequence of function calls and then compares that to the actual function calls made. If this fails then the code has been either inadvertently modified or the unit test needs to be updated.

---
In reply to: [305166517](https://github.com/Azure/azure-c-shared-utility/pull/292#discussion_r305166517) [](ancestors = 305166517)

tests/tlsio_bearssl_ut/CMakeLists.txt
markrad: Yes indeed. I inadvertently left this residue in from my testing. I'll take a look at it.

---
In reply to: [305153694](https://github.com/Azure/azure-c-shared-utility/pull/292#discussion_r305153694) [](ancestors = 305153694)

markrad:
I'm not that familiar with the E2E tests. Can't one just change the TLS library and run them? They don't do function mocking do they?

---
In reply to: [484196587](https://github.com/Azure/azure-c-shared-utility/pull/292#issuecomment-484196587) [](ancestors = 484196587)

PR #302, opened 2019-04-18
logerror captures stack traces

anporumb:


inc/azure_c_shared_utility/xlogging.h
```diff
@@ -109,6 +116,53 @@ typedef void(*LOGGER_LOG_GETLASTERROR)(const char* file, const char* func, int l
...
+                    l(log_category, __FILE__, FUNC_NAME, __LINE__, log_options, format, __VA_ARGS__);                                                                    \
+                }                                                                                                                                                        \
+                else                                                                                                                                                     \
+                {                                                                                                                                                        \
+                    size_t formatSize = strlen(format);                                                                                                                  \
+                    char* formatWithStack = (char*)malloc(formatSize + sizeof("STACK_PRINT_FORMAT"));                                                                    \
```
dcristoloveanu: >malloc [](start = 51, length = 6)

This needs to be logging_malloc

inc/azure_c_shared_utility/xlogging.h
```diff
@@ -109,6 +116,53 @@ typedef void(*LOGGER_LOG_GETLASTERROR)(const char* file, const char* func, int l
         } \
     } \
 }
+#else /*LOGERROR_CAPTURES_STACK_TRACES is defined*/ 
+extern char* getStackAsString(void);
+#define STACK_PRINT_FORMAT "\nStack:\n%s"
+#define LOG(log_category, log_options, format, ...)                                                                                                                      \
```
dcristoloveanu: >LOG(log [](start = 8, length = 7)

So according to an illustrious developer: what's the point if we don't do it for etw :-)?

inc/azure_c_shared_utility/xlogging.h
anporumb: Changed

---
In reply to: [276783533](https://github.com/Azure/azure-c-shared-utility/pull/302#discussion_r276783533) [](ancestors = 276783533)

inc/azure_c_shared_utility/xlogging.h
anporumb: Now we do.

---
In reply to: [276822018](https://github.com/Azure/azure-c-shared-utility/pull/302#discussion_r276822018) [](ancestors = 276822018)

src/etwxlogging.c
```diff
@@ -14,6 +17,124 @@
...
+
+typedef union SYMBOL_INFO_EXTENDED_TAG
+{
+    SYMBOL_INFO symbol;
+    unsigned char extendsUnion[sizeof(SYMBOL_INFO) + TRACE_MAX_SYMBOL_SIZE - 1]; /*this field only exists to extend the size of the union to encompass "CHAR    Name[1];" of the SYMBOL_INFO to be as big as TRACE_MAX_SYMBOL_SIZE - 1 */ /*and the reason why it is not malloc'd is to exactly avoid a malloc that cannot be LogError'd (how does one log errors in a log function?!)*/
+}SYMBOL_INFO_EXTENDED;
+
+
+static SRWLOCK lockOverSymCalls = SRWLOCK_INIT;
+char* getStackAsString(void)
```
dcristoloveanu: >getStackAsString [](start = 6, length = 16)

I'd really like if this would not be duplicated. It really does the same thing like the one in xlogging, right?

So I suggest having a logging_stacktrace.c or similar where this code that gets the stack trace string built lives.

src/xlogging.c
```diff
@@ -18,8 +38,123 @@ void xlogging_set_log_function(LOGGER_LOG log_function)
...
+                    {
+                        (void)strncat(destination, "snprintf failed\n", destinationSize - destinationPos);
+                    }
+                    else
+                    {
+                        (void)strncat(destination, resultLine, destinationSize - destinationPos);
+                    }
+                }
+            }
+            destinationPos = strlen(destination);
```
dcristoloveanu: >destinationPos = strlen(destination); [](start = 12, length = 37)

Consider: Calling strlen again is not really needed if one would add to destinationPos the number of chars that are strncat'd (the quantity is known in all cases - either COUNTOF("snprintf failed\n") or snprintfResult.

src/xlogging.c
```diff
@@ -18,8 +38,123 @@ void xlogging_set_log_function(LOGGER_LOG log_function)
...
+        if (state == 0)
+        {
+            (void)SymInitialize(GetCurrentProcess(), NULL, TRUE);
+            (void)InterlockedExchange(&doSymInit, 2);
+        }
+    }
+#endif
+#endif
     return global_log_function;
 }
```
dcristoloveanu: Frankly, this says we should add an xlogging_init/deinit.
Just so we don't piggy back the getter.

I don't think it would be hard to add. What says you?

src/xlogging.c
```diff
@@ -18,8 +38,123 @@ void xlogging_set_log_function(LOGGER_LOG log_function)
...
+            DWORD displacement = 0;
+
+            if (!SymFromAddr(process, address, NULL, symbol))
+            {
+                (void)strncat(destination, "SymFromAddr failed\n", destinationSize - destinationPos);
+            }
+            else
+            {
+                line.SizeOfStruct = sizeof(IMAGEHLP_LINE64);
+                if (SymGetLineFromAddr64(process, address, &displacement, &line))
```
dcristoloveanu: Cheers for adding this!

src/etwxlogging.c
anporumb: Agree. Let me see about it.

---
In reply to: [276825068](https://github.com/Azure/azure-c-shared-utility/pull/302#discussion_r276825068) [](ancestors = 276825068)

src/xlogging.c
anporumb: I've considered and strlen is no longer called. This was a great suggestion!

---
In reply to: [276825923](https://github.com/Azure/azure-c-shared-utility/pull/302#discussion_r276825923) [](ancestors = 276825923)

src/xlogging.c
anporumb: Me says "when time comes". 

---
In reply to: [276826626](https://github.com/Azure/azure-c-shared-utility/pull/302#discussion_r276826626) [](ancestors = 276826626)

src/etwxlogging.c
dcristoloveanu: Seems like it is still duplicated.

---
In reply to: [276827635](https://github.com/Azure/azure-c-shared-utility/pull/302#discussion_r276827635) [](ancestors = 276827635,276825068)

PR #311, opened 2019-05-08
Fix building and producing configs when using installed dependencies

ewertons:
Please look at code for details.


CMakeLists.txt
```diff
@@ -676,6 +678,7 @@ if(${use_installed_dependencies})
         FILES
             "configs/${PROJECT_NAME}Config.cmake"
             "configs/${PROJECT_NAME}Functions.cmake"
+           "configs/azure_iot_build_rules.cmake"
```
jspaith: >"configs/azure_iot_build_rules.cmake" [](start = 11, length = 37)

Nit - start of line does not align with others in same block.

ewertons: to be fixed.

ewertons: Done.

jenkins/linux_install_deps.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
jspaith: Please add this script to the .vsts-ci.yml for some platform(?).

ewertons: That will come afterwards. But it will happen

CMakeLists.txt
```diff
@@ -580,6 +580,8 @@ if(NOT ${use_installed_dependencies})
     endif()
 endif()
 
+include(dependencies.cmake)
```
jebrando: Not this time but the next, we need to consolidate the umock-c either put all of it in dependencies.cmake or have it here.

PR #314, opened 2019-05-10
UUID_from_string: fix UUID_T overflow

mahilleb-msft:
- Too few dashes in UUID string could have caused to write beyond UUID_T.
- Check strictly for expected position with dashes.


ewertons:
This PR is being replaced by https://github.com/Azure/azure-c-shared-utility/pull/626

PR #327, opened 2019-06-03
Add support for private keys in crypto devices

OYTIS:
The PR is related to this issue: https://github.com/Azure/azure-iot-sdk-c/issues/1022

I tried to make changes to the API minimal, therefore the old option to add plain text private key is still there. Nevertheless, some changes to the API were necessary, so probably some projects depending on the c-utility can fail to build. If you approve the idea behind this PR, I can make fixes to the other ones.

Only OpenSSL is supported so far, but nothing prevents supporting other libraries.

No tests are included in this PR, I will definitely need maintainers' help here.

A PR to uhttp utilizing this will be made shortly.

### DIFF
--- a/adapters/tlsio_openssl.c
+++ b/adapters/tlsio_openssl.c
@@ -12,6 +12,7 @@
 #include "azure_c_shared_utility/lock.h"
 #include "azure_c_shared_utility/tlsio.h"
 #include "azure_c_shared_utility/tlsio_openssl.h"
+#include "azure_c_shared_utility/tlsio_cryptodev.h"
 #include "azure_c_shared_utility/socketio.h"
 #include "azure_c_shared_utility/optimize_size.h"
 #include "azure_c_shared_utility/xlogging.h"
@@ -71,6 +72,7 @@ typedef struct TLS_IO_INSTANCE_TAG
     char* cipher_list;
     const char* x509_certificate;
     const char* x509_private_key;
+    TLSIO_CRYPTODEV_PKEY* x509_cryptodev_private_key;
     TLSIO_VERSION tls_version;
     TLS_CERTIFICATE_VALIDATION_CALLBACK tls_validation_callback;
     void* tls_validation_callback_data;
@@ -148,6 +150,16 @@ static void* tlsio_openssl_CloneOption(const char* name, const void* value)
             else
             {
                 /*return as is*/
+            }
+        }
+        else if (strcmp(name, SU_OPTION_X509_CRYPTODEV_PRIVATE_KEY) == 0)
+        {
+            result = malloc(sizeof(TLSIO_CRYPTODEV_PKEY));
+
+            if (result != NULL) {
+                memcpy(result, value, sizeof(TLSIO_CRYPTODEV_PKEY));
+            } else {
+                LogError("unable to allocate memory for TLSIO_CRYPTODEV_PKEY object");
             }
         }
         else if (strcmp(name, OPTION_X509_ECC_CERT) == 0)
@@ -249,6 +261,7 @@ static void tlsio_openssl_DestroyOption(const char* name, const void* value)
             (strcmp(name, OPTION_OPENSSL_CIPHER_SUITE) == 0) ||
             (strcmp(name, SU_OPTION_X509_CERT) == 0) ||
             (strcmp(name, SU_OPTION_X509_PRIVATE_KEY) == 0) ||
+            (strcmp(name, SU_OPTION_X509_CRYPTODEV_PRIVATE_KEY) == 0) ||
             (strcmp(name, OPTION_X509_ECC_CERT) == 0) ||
             (strcmp(name, OPTION_X509_ECC_KEY) == 0) ||
             (strcmp(name, OPTION_TLS_VERSION) == 0)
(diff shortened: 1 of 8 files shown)


adapters/tlsio_openssl.c
```diff
@@ -970,6 +1001,14 @@ static int create_openssl_instance(TLS_IO_INSTANCE* tlsInstance)
         log_ERR_get_error("unable to set cipher list.");
         result = MU_FAILURE;
     }
+    else if ((tlsInstance->sigalgs_list != NULL) &&
```
massand: >sigalgs_list [](start = 27, length = 12)

Is this required? How is it internally handled if we do not pass this in?

tests/x509_openssl_ut/x509_openssl_ut.c
```diff
@@ -368,7 +368,7 @@ BEGIN_TEST_SUITE(x509_openssl_unittests)
         //arrange
 
         //act
-        int result = x509_openssl_add_credentials(NULL, "certificate", "privatekey");
+        int result = x509_openssl_add_credentials(NULL, "certificate", "privatekey", NULL);
```
massand: >NULL [](start = 85, length = 4)

Require UT for new option

adapters/tlsio_openssl.c
```diff
@@ -150,6 +153,16 @@ static void* tlsio_openssl_CloneOption(const char* name, const void* value)
                 /*return as is*/
             }
         }
+        else if (strcmp(name, SU_OPTION_X509_CRYPTODEV_PRIVATE_KEY) == 0)
+        {
+	    result = malloc(sizeof(TLSIO_CRYPTODEV_PKEY));
+
+	    if (result != NULL) {
```
massand: >{ [](start = 25, length = 1)

nit: please adhere to existing style conventions in file.

adapters/tlsio_openssl.c
OYTIS: This is something that is required for our particular hardware (as it for the moment can't sign SHA512 hashes), but may also be required for someone else's. This option enables user to restrict signature algorithms used by OpenSSL. If you don't set it, OpenSSL will use its default list, which is what most users want.

I've made this change a separate commit, I can also make it a separate PR if you believe it's easier to review it this way.

adapters/tlsio_openssl.c
OYTIS: Thanks, I'll fix where I find it (is it tabs vs spaces?). Do you have some script to check it by the way?

tests/x509_openssl_ut/x509_openssl_ut.c
OYTIS: It looks like there is a requirement that `If any argument is NULL then x509_openssl_add_credentials shall fail and return a non-zero value.`. My design was to allow user to set either third or fourth argument to NULL if they use plaintext vs cryptodev private key.

Do you believe it would be better to introduce a completely new API call (e.g. `x509_openssl_app_cryptodev_key`)?

massand: Yes - I would be interested in seeing a different method in the API design for the private keys. We cannot break the existing "API contract" for backwards-compat reasons.

OYTIS: @massand Updated the PR to keep the old API untouched. So to add unit tests got the new call (`x509_openssl_add_credentials_cryptodev`) do we need to create the requirements somewhere, or I can just add them right away (adapt from those for `x509_openssl_add_credentials` probably)?

adapters/tlsio_openssl.c
OYTIS: Deleted tabs where I could find them.

tests/x509_openssl_ut/x509_openssl_ut.c
massand: Yes, please add the requirements adapted from an existing module (such as x509_openssl_add_credentials). I try to keep the requirements to a minimum of happy path requirements and requirements specifying major error conditions. I would also keep requirements scoped to public interface API methods, instead of each internal method. Thank you for maintaining the consistency again!

OYTIS: @massand Sorry, I've got a bit stuck with debugging the umock. So I extended `setup_load_credentials` to support crypto device scenario. My test crashes, and debugging shows it's because some of `SetReturn` calls are being ignored. As far as I understand it can happen if the sequence of `STRICT_EXPECTED_CALL`s is incorrect, so umock can't identify `SetReturn` with this particular call.
So how can I debug this issue further? How can I see if umock has 'checked' a `STRICT_EXPECTED_CALL` I've defined?

OYTIS: Added tests, please review.

Only tested on Linux with `OPENSSL_VERSION_NUMBER == 0x1010007f`, but I _tried_ to get the test numbering right.

OYTIS:
@jebrando @massand  @jspaith 
Rebased on current master.

ericwol-msft:
Anton, we are not going to take this PR. OpenSSL engine is now supported in this SDK. Also please consider using the embedded C SDK at https://github.com/Azure/azure-sdk-for-c

PR #332, opened 2019-06-14
Fix includes

dcristoloveanu:
Fix includes


mattdurak:
> #include <stdint.h>

missing if def c++ cstdint

---
Refers to: inc/azure_c_shared_utility/sha.h:30 in b77df06. [](commit_id = b77df0662c081853ea0a79801b42de5c1709dac3, deletion_comment = False)

PR #334, opened 2019-06-20
Define USE_OPENSSL macro if it's enabled in CMake

rcdailey:
In `platform_linux.c`, `USE_OPENSSL` is not defined if it's enabled in cmake via the `use_openssl` option. Because of this, the header never gets included:

```cpp
#ifdef USE_OPENSSL
#include "azure_c_shared_utility/tlsio_openssl.h"
#endif
```

This is slightly related to https://github.com/Azure/azure-iot-sdk-c/issues/999

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,10 @@ if(${use_custom_heap})
     add_definitions(-DGB_USE_CUSTOM_HEAP)
 endif()
 
+if(${use_openssl})
+    add_definitions(-DUSE_OPENSSL)
+endif()
+
 if(WIN32)
     option(use_schannel "set use_schannel to ON if schannel is to be used, set to OFF to not use schannel" ON)
     option(use_openssl "set use_openssl to ON if openssl is to be used, set to OFF to not use openssl" OFF)
@@ -673,4 +677,3 @@ DESTINATION
 )
 
 compileTargetAsC99(aziotsharedutil)
-


rcdailey:
@massand: If you have a moment...

massand:
@rcdailey thanks for the PR. The inclusion of USE_OPENSSL actually does occur on line 137 of the CMakeLists.txt. The higher order bit here that will help unblock here is the OS type. If it isn't set to MSVC, LINUX or APPLE, it ignores setting the CMAKE_C_FLAGS or CMAKE_CXX_FLAGS to include -DUSE_OPENSSL. We need to evaluate adding a new generic OS type or Android specifically to unblock folks, but at this point there aren't any plans to add this support.

massand:
Also, I think it is possible to follow the compileOption_C cmake option pattern established in our azure-iot-sdk-c repo.  
Here is how it is defined - https://github.com/Azure/azure-iot-sdk-c/blob/master/CMakeLists.txt#L77
Here is how it is used - https://github.com/Azure/azure-iot-sdk-c/blob/master/build_all/linux/build.sh#L130

rcdailey:
Not sure why you closed this. Even though I linked back to Android support, that's got nothing to do with this PR.

Enabling OpenSSL support is not a platform-specific behavior. The reason this isn't working is because you have the compiler flags duplicated under each platform-specific if condition. My solution is actually the correct one for a few reasons:

1. `CMAKE_C_FLAGS` and `CMAKE_CXX_FLAGS` are meant to be set from the command line. They aren't meant to be the first-class way of defining compiler flags for your targets. For that, you should be using `add_definition()` or (preferrably) `target_compile_definitions()`.
2. Mine sets the appropriate flags if OpenSSL is found and enabled. What determines whether or not the SSL-specific code should be enabled is the presence of OpenSSL, not a specific platform.

Please reconsider my proposal. I am happy to amend my commit with the removal of the compile flags you set around line 137. My change works with your existing setup and enables more use cases (such as Android, without you having to explicitly support it).

rcdailey:
@dcristoloveanu Could I get your opinion as well? I see you have a few commits in the repo.

PR #339, opened 2019-07-02
Add constbufferarray_create_with_move_memory tests and code

dcristoloveanu:
Add constbufferarray_create_with_move_memory tests and code

### DIFF
--- a/src/constbuffer_array.c
+++ b/src/constbuffer_array.c
@@ -15,16 +15,18 @@
 typedef struct CONSTBUFFER_ARRAY_HANDLE_DATA_TAG
 {
     uint32_t nBuffers;
+    bool created_with_moved_memory;
+    CONSTBUFFER_HANDLE* buffers;
 #ifdef _MSC_VER
     /*warning C4200: nonstandard extension used: zero-sized array in struct/union : looks very standard in C99 and it is called flexible array. Documentation-wise is a flexible array, but called "unsized" in Microsoft's docs*/ /*https://msdn.microsoft.com/en-us/library/b6fae073.aspx*/
 #pragma warning(disable:4200)
 #endif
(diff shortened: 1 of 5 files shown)


src/constbuffer_array.c
```diff
@@ -159,14 +195,16 @@ CONSTBUFFER_ARRAY_HANDLE constbuffer_array_create_from_array_array(const CONSTBU
...
+                    result->buffers = result->buffers_memory;
 
                     for (dest_idx = 0, array_idx = 0; array_idx < buffer_array_count; ++array_idx)
                     {
                         for (source_idx = 0; source_idx < buffer_arrays[array_idx]->nBuffers; ++source_idx, ++dest_idx)
                         {
                             /*Codes_SRS_CONSTBUFFER_ARRAY_42_004: [ constbuffer_array_create_from_array_array shall copy all of the CONSTBUFFER_HANDLES from each const buffer array in buffer_arrays to the newly constructed array by calling CONSTBUFFER_IncRef. ]*/
                             CONSTBUFFER_IncRef(buffer_arrays[array_idx]->buffers[source_idx]);
-                            result->buffers[dest_idx] = buffer_arrays[array_idx]->buffers[source_idx];
+                            result->buffers_memory[dest_idx] = buffer_arrays[array_idx]->buffers[source_idx];
```
mattdurak: >buffers_memory [](start = 36, length = 14)

consider: shouldn't this be `buffers` because then this line isn't dependent on lines 198-199? Seems consistent to always access through the same alias pointer. Same in the functions below.

dcristoloveanu: Makes sense, thank! Will change it

---
In reply to: [299264337](https://github.com/Azure/azure-c-shared-utility/pull/339#discussion_r299264337) [](ancestors = 299264337)

PR #342, opened 2019-07-09
Fixing mbed tls not to close the socket until the TLS layer notifies the peer of closure

jebrando:


adapters/tlsio_mbedtls.c
```diff
@@ -324,6 +323,24 @@ static int on_io_recv(void *context, unsigned char *buf, size_t sz)
...
 }
 
+static void on_send_complete(void* context, IO_SEND_RESULT send_result)
+{
+    TLS_IO_INSTANCE* tls_io_instance = (TLS_IO_INSTANCE *)context;
+    if (tls_io_instance != NULL)
+    {
+        // If the state is not open then this is probably an internal call
+        // So don't notify the upper level
+        if (tls_io_instance->tlsio_state == TLSIO_STATE_OPEN)
```
ewertons: This might be dangerous, because users could have sent something on the wire, then a state change occurs in the mbed tls layer that causes it not to be in OPEN state, and now the send confirmation for a valid piece of data will not be returned to the caller.

jebrando: Change to make sure we're not closing which r our messages.

---
In reply to: [301696099](https://github.com/Azure/azure-c-shared-utility/pull/342#discussion_r301696099) [](ancestors = 301696099)

PR #348, opened 2019-08-01
fix socketio queue bug on seeing EAGAIN

chlandsi:
Messages that cannot be sent right away because the call to `send` returns EAGAIN are not enqueued for retrying the send later, but discarded.


ewertons:
Thank you for your contribution, @chlandsi .

This issue was fixed by commit https://github.com/Azure/azure-c-shared-utility/commit/83f5ca94c6caf394f53611b014181d4c34843a66

PR #352, opened 2019-08-06
Extend support for hardware keys to mbedTLS and wolfSSL

OYTIS:
Added support for MbedTLS and wolfSSL to https://github.com/Azure/azure-c-shared-utility/pull/327

As there are some issues with wolfSSL and MbedTLS in the Azure SDK itself, it does not work perfectly, but it does show the same degree of correctness as cleartext keys.


ericwol-msft:
Anton, please evaluate the embedded C SDK. https://github.com/Azure/azure-sdk-for-c

PR #362, opened 2019-08-22
add a method for getting elapsed time in ms

amarsa-msft:
This method gets the elapsed time in milliseconds.

### DIFF
--- a/adapters/timer.c
+++ b/adapters/timer.c
@@ -60,6 +60,23 @@ double timer_get_elapsed(TIMER_HANDLE timer)
     return result;
 }
 
+double timer_get_elapsed_ms(TIMER_HANDLE timer)
+{
+    double result;
+    if (timer == NULL)
+    {
+        LogError("invalid arg TIMER_HANDLE timer=%p", timer);
+        result = -1.0;
+    }
+    else
+    {
+        LARGE_INTEGER stopTime;
+        (void)QueryPerformanceCounter(&stopTime);
+        result = ((double)(stopTime.QuadPart - timer->startTime.QuadPart) / (double)timer->freq.QuadPart) * 1000;
+    }
+    return result;
+}
+
 void timer_destroy(TIMER_HANDLE timer)
 {
     if (timer == NULL)


adapters/timer.c
```diff
@@ -60,6 +60,23 @@ double timer_get_elapsed(TIMER_HANDLE timer)
     return result;
 }
 
+double timer_get_elapsed_ms(TIMER_HANDLE timer)
```
jebrando: Do you have any tests for this?

tests/timer_ut/timer_ut.c
```diff
@@ -0,0 +1,229 @@
...
+    TIMER_HANDLE timer = timer_create();
+    umock_c_reset_all_calls();
+    STRICT_EXPECTED_CALL(mocked_QueryPerformanceCounter(IGNORED_PTR_ARG))
+        .CopyOutArgumentBuffer_lpPerformanceCount(&stop_time, sizeof(stop_time));
+    
+    //act
+    double elapsed_time = timer_get_elapsed(timer);
+
+    //assert
+    ASSERT_ARE_EQUAL(double, 9.9, elapsed_time);
```
mattdurak: > ASSERT_ARE_EQUAL(double, 9.9, elapsed_time); [](start = 3, length = 45)

Something tells me this may require a fix for the `double_Compare` defined by ctest.

int double_Compare(double left, double right)
{
    return left != right;
}


Because (9.9 == 99/10) is not necessarily true.

`double_Compare` may need to allow some small epsilon difference to make this behave as expected. Or maybe this test should just do that (e.g. make sure `elapsed_time` is > (9.9-epsilon) and `elapsed_time` is < (9.9+epsilon)).

mattdurak: Also, ctestrunnerswitcher doesn't seem to do the right thing for floating point compare when CppUnit is used.

```
#define ASSERT_ARE_EQUAL(type, A, B, ...) \
    do \
    { \
        char* ctrs_message = CONSTRUCT_CTRS_MESSAGE(__VA_ARGS__); \
        std::wstring cppUnitTestMessage = ToString(ctrs_message); \
        ctrs_sprintf_free(ctrs_message); \
        Assert::AreEqual((type)(A), (type)(B), cppUnitTestMessage.c_str()); \
    } while ((void)0, 0)
```

But the signature has a `tolerance` that we don't pass in.

```
  // double AreNotEqual comparison:
  static void AreNotEqual(double notExpected, double actual, double tolerance, const wchar_t* message = NULL, const __LineInfo* pLineInfo = NULL)
  {
   double diff = notExpected - actual;
   FailOnCondition(fabs(diff) > fabs(tolerance), NOT_EQUALS_MESSAGE(notExpected, actual, message), pLineInfo);
  }
```


---
In reply to: [317311507](https://github.com/Azure/azure-c-shared-utility/pull/362#discussion_r317311507) [](ancestors = 317311507)

amarsa-msft: had a discussion with Matt, keeping it as is to keep the test simple.

PR #365, opened 2019-08-29
stop printf logging when etw fails

anporumb:
ETW will make 1 printf and no more printf until ETW state is restored to working

### DIFF
--- a/src/etwlogger_driver.c
+++ b/src/etwlogger_driver.c
@@ -130,7 +130,7 @@ static void lazyRegisterEventProvider(void)
         {
             /*should check if the provider is registered with the system... maybe at a later time*/
             isETWLoggerInit = 1; /*and stays 1 until the process exits*/ /*sorry, no graceful exit with EventUnregisterMicrosoft_ServiceBus*/
-            LogInfo("EventRegisterMicrosoft_ServiceBus success"); /*selflogging that the log service has started*/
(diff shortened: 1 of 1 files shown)


src/etwlogger_driver.c
```diff
@@ -218,7 +224,14 @@ void etwlogger_log(LOG_CATEGORY log_category, const char* file, const char* func
                 if (event_write_result != ERROR_SUCCESS)
                 {
                     /*fallback on printf...*/
-                    (void)printf("failed in EventWriteLogInfoEvent: %lu\r\n", event_write_result);
+                    if (InterlockedCompareExchange(&stopPrintfing, 1, 0) == 0)
+                    {
+                        (void)printf("failed in EventWriteLogInfoEvent: %lu. Further failing calls to EventWriteLogInfoEvent will not result in printf\r\n", event_write_result);
```
jebrando: >printf("failed in EventWriteLogInfoEvent: %lu. Further failing calls to EventWriteLogInfoEvent will not result in printf\r\n", event_write_result); [](start = 30, length = 147)

Do you really need to say you're going to stop printing, instead of just not printing?

anporumb: This is printed just once - and at that moment at least 1x ETW event is lost. Otherwise there would be an empty console that doesn't say anything and the user is left wondering "where's my events..."

---
In reply to: [319255960](https://github.com/Azure/azure-c-shared-utility/pull/365#discussion_r319255960) [](ancestors = 319255960)

PR #373, opened 2019-09-10
Fix check for lock result in openssl tlsio

jebrando:


adapters/tlsio_openssl.c
```diff
@@ -402,14 +402,14 @@ static void openssl_lock_unlock_helper(LOCK_HANDLE lock, int lock_mode, const ch
 
     if (lock_mode & CRYPTO_LOCK)
     {
-        if (Lock(lock) != 0)
```
dcristoloveanu: >0) [](start = 26, length = 2)

Haha. How did this pass the gate? No tests?

PR #374, opened 2019-09-12
Async DNS resolution using c-ares

massand:
Adding a asynchronous DNS resolution helper for the socket_io layers that can support the c-ares library (which is quite portable across many platforms).

Currently, the async DNS functionality supports Linux sockets only. In order to use the new helper, a cmake option named 'use-cares' is used to reference the C-ares libraries (obtained via apt-get on the Linux build system). Therefore, it is assumed that c-ares is part of your toolchain's sysroot/lib and /include (i.e. we do not plan to submodule c-ares), akin to how OpenSSL dependency is structured.

The dns_resolver adds a thin interface atop the current DNS resolution functionality (i.e. getaddrinfo) and socket_win32 & socket_berkley changes were made to use that interface across both sync and async DNS implementations. 

The socket UT and e2e validation is complete. The e2e test infra changes are introduced in a separate azure-iot-sdk-c PR.


src/dns_resolver_ares.c
```diff
@@ -0,0 +1,283 @@
...
+            if (ms_result != 0)
+            {
+                /* Codes_SRS_dns_resolver_30_014: [ On any failure, dns_resolver_create shall log an error and return NULL. ]*/
+                LogError("dns_resolver_create: hostname allocation failed");
+                free(result);
+                result = NULL;
+            }
+
+            status = ares_library_init(ARES_LIB_INIT_WIN32);
+            if (status != ARES_SUCCESS)
```
jebrando: Result will fail and then throw an exception on line 96.  Need to put an else after the if block.

src/dns_resolver_ares.c
```diff
@@ -0,0 +1,283 @@
...
+            {
+                status = ares_init(&(result->ares_resolver));
+                if(status != ARES_SUCCESS)
+                {
+                    LogError("ares_init failed: %s\n", ares_strerror(status));
+                    result = NULL;
+                }
+            }
+
+            result->addrInfo = NULL;
```
jebrando: >result->addrInfo = NULL; [](start = 12, length = 24)

Need to have this above with the rest of the initialization.

src/dns_resolver_ares.c
```diff
@@ -0,0 +1,283 @@
...
+                LogError("ares_library_init failed: %s\n", ares_strerror(status));
+                result = NULL;
+            }
+            else 
+            {
+                status = ares_init(&(result->ares_resolver));
+                if(status != ARES_SUCCESS)
+                {
+                    LogError("ares_init failed: %s\n", ares_strerror(status));
+                    result = NULL;
```
jebrando: >result = NULL; [](start = 20, length = 14)

Need free before this to avoid mem leak

src/dns_resolver_ares.c
```diff
@@ -0,0 +1,283 @@
...
+}
+
+static void query_completed_cb(void *arg, int status, int timeouts, struct hostent *he)
+{
+    int i;
+    struct addrinfo *ptr = NULL;
+    struct sockaddr_in *addr;
+    
+    DNSRESOLVER_INSTANCE *dns = (DNSRESOLVER_INSTANCE *)arg;
+    (void)status;
```
jebrando: >(void)status; [](start = 4, length = 13)

unneeded because it's used on line 112.

src/dns_resolver_ares.c
```diff
@@ -0,0 +1,283 @@
...
+    (void)timeouts;
+
+    if(status != ARES_SUCCESS)
+    {
+        LogError("ARES error: %d", status);
+    }
+    else
+    {
+        dns->addrInfo = calloc(1, sizeof(struct addrinfo));
```
jebrando: >dns->addrInfo = calloc(1, sizeof(struct addrinfo)); [](start = 8, length = 51)

Check for failure.

tests/socketio_berkeley_ut/CMakeLists.txt
```diff
@@ -15,6 +15,8 @@ ${theseTestsName}.c
 
 set(${theseTestsName}_c_files
 ../../adapters/socketio_berkeley.c
+../../src/dns_resolver_sync.c
```
danewalton: test failed and i think you might need `dns_resolver_ares.c` here?

src/dns_resolver_ares.c
```diff
@@ -0,0 +1,311 @@
...
+#include "azure_c_shared_utility/dns_resolver.h"
+#include "azure_c_shared_utility/gballoc.h"
+#include "azure_c_shared_utility/crt_abstractions.h"
+#include "azure_c_shared_utility/xlogging.h"
+#include "ares.h"
+
+// EXTRACT_IPV4 pulls the uint32_t IPv4 address out of an addrinfo struct
+// This will not be needed for the asynchronous design
+#ifdef WIN32_NOT_USED
+//#define EXTRACT_IPV4(ptr) ((struct sockaddr_in *) ptr->ai_addr)->sin_addr.S_un.S_addr
```
jebrando: Commented code, please remove if necessary here and other files..

PR #380, opened 2019-09-19
Fix critical socketio_berkeley bug where socketio_send() will drop packets if send() returns -1 and errno=EAGAIN

ewertons:
This is a fix for a bug in socketio_berkeley where packets may be lost by socketio_send().
In socketio_send, if send() returns -1 and errno is EAGAIN, the packet does not get added to the pending I/O queue and gets dropped.

This was caught by attempting to send huge amounts of data in a single shot to a remote endpoint which is applying throttling.
After a few calls to socketio_send where packets were lost, on the first call to send() that succeeds, the remote endpoint receives data out of order (since previous packets got dropped by socketio), cannot decrypt it and severs the connection.

The other socketio adapters are not affected by this bug (see [socketio_mbed.c](https://github.com/Azure/azure-c-shared-utility/blob/255cbc81a0f7bd2a34664503ce1f0276da5bb374/adapters/socketio_mbed.c#L333), [socketio_mbed_os5.c](https://github.com/Azure/azure-c-shared-utility/blob/255cbc81a0f7bd2a34664503ce1f0276da5bb374/adapters/socketio_mbed_os5.c#L201), [socketio_win32.c](https://github.com/Azure/azure-c-shared-utility/blob/255cbc81a0f7bd2a34664503ce1f0276da5bb374/adapters/socketio_win32.c#L485)).

Note: there are no tests for this module. E2E tests would need to be added to correctly verify the behavior of socketio (and other xio layers) on key conditions (network latency/delay, high-throughput, etc).

For reference: On non-blocking mode send() will return -1 and errno=EAGAIN if the socket is not ready to send (see https://linux.die.net/man/3/send).

### DIFF
--- a/adapters/socketio_berkeley.c
+++ b/adapters/socketio_berkeley.c
@@ -919,7 +919,7 @@ int socketio_send(CONCRETE_IO_HANDLE socket_io, const void* buffer, size_t size,
                     {
                         /*send says "come back later" with EAGAIN - likely the socket buffer cannot accept more data*/
                         /* queue data */
-                        size_t bytes_sent = (send_result < 0 ? 0: send_result);
+                        size_t bytes_sent = (send_result < 0 ? 0 : send_result);
 
                         if (add_pending_io(socket_io_instance, buffer + bytes_sent, size - bytes_sent, on_send_complete, callback_context) != 0)
                         {
--- a/adapters/socketio_berkeley.c
+++ b/adapters/socketio_berkeley.c
@@ -910,32 +910,18 @@ int socketio_send(CONCRETE_IO_HANDLE socket_io, const void* buffer, size_t size,
                 ssize_t send_result = send(socket_io_instance->socket, buffer, size, 0);
                 if ((send_result < 0) || ((size_t)send_result != size))
                 {
-                    if (send_result == INVALID_SOCKET)
+                    if (errno != EAGAIN)
                     {
-                        if (errno == EAGAIN) /*send says "come back later" with EAGAIN - likely the socket buffer cannot accept more data*/
-                        {
-                            /* queue data */
-                            if (add_pending_io(socket_io_instance, buffer, size, on_send_complete, callback_context) != 0)
-                            {
-                                LogError("Failure: add_pending_io failed.");
-                                result = MU_FAILURE;
-                            }
-                            else
-                            {
-                                /*do nothing*/
-                                result = 0;
-                            }
-                        }
-                        else
-                        {
-                            LogError("Failure: sending socket failed. errno=%d (%s).", errno, strerror(errno));
-                            result = MU_FAILURE;
-                        }
+                        LogError("Failure: sending socket failed. errno=%d (%s).", errno, strerror(errno));
+                        result = MU_FAILURE;
                     }
                     else
                     {
+                        /*send says "come back later" with EAGAIN - likely the socket buffer cannot accept more data*/
                         /* queue data */
-                        if (add_pending_io(socket_io_instance, buffer + send_result, size - send_result, on_send_complete, callback_context) != 0)
+                        size_t bytes_sent = (send_result < 0 ? 0: send_result);
+
+                        if (add_pending_io(socket_io_instance, buffer + bytes_sent, size - bytes_sent, on_send_complete, callback_context) != 0)
                         {
                             LogError("Failure: add_pending_io failed.");
                             result = MU_FAILURE;


adapters/socketio_berkeley.c
```diff
@@ -910,32 +910,18 @@ int socketio_send(CONCRETE_IO_HANDLE socket_io, const void* buffer, size_t size,
...
-                                /*do nothing*/
-                                result = 0;
-                            }
-                        }
-                        else
-                        {
-                            LogError("Failure: sending socket failed. errno=%d (%s).", errno, strerror(errno));
-                            result = MU_FAILURE;
-                        }
+                        LogError("Failure: sending socket failed. errno=%d (%s).", errno, strerror(errno));
```
jebrando: >. errno=%d (%s).", errno, strerror(errno)); [](start = 64, length = 43)

Do we care how many bytes it sent?  I'm not sure.

adapters/socketio_berkeley.c
```diff
@@ -910,32 +910,18 @@ int socketio_send(CONCRETE_IO_HANDLE socket_io, const void* buffer, size_t size,
...
+                        result = MU_FAILURE;
                     }
                     else
                     {
+                        /*send says "come back later" with EAGAIN - likely the socket buffer cannot accept more data*/
                         /* queue data */
-                        if (add_pending_io(socket_io_instance, buffer + send_result, size - send_result, on_send_complete, callback_context) != 0)
+                        size_t bytes_sent = (send_result < 0 ? 0: send_result);
```
rafilho: >send_result [](start = 66, length = 11)

We should log in case it came here due to a EAGAIN (since it is still an error, even if treatable).

adapters/socketio_berkeley.c
```diff
@@ -910,32 +910,18 @@ int socketio_send(CONCRETE_IO_HANDLE socket_io, const void* buffer, size_t size,
...
-                            LogError("Failure: sending socket failed. errno=%d (%s).", errno, strerror(errno));
-                            result = MU_FAILURE;
-                        }
+                        LogError("Failure: sending socket failed. errno=%d (%s).", errno, strerror(errno));
+                        result = MU_FAILURE;
                     }
                     else
                     {
+                        /*send says "come back later" with EAGAIN - likely the socket buffer cannot accept more data*/
                         /* queue data */
```
rafilho: We should log in case it came here due to a EAGAIN (since it is still an error, even if treatable).

adapters/socketio_berkeley.c
rafilho: I think according to the documentation, an error => no bytes sent.

---
In reply to: [326645677](https://github.com/Azure/azure-c-shared-utility/pull/380#discussion_r326645677) [](ancestors = 326645677)

adapters/socketio_berkeley.c
```diff
@@ -910,32 +910,18 @@ int socketio_send(CONCRETE_IO_HANDLE socket_io, const void* buffer, size_t size,
                 ssize_t send_result = send(socket_io_instance->socket, buffer, size, 0);
                 if ((send_result < 0) || ((size_t)send_result != size))
                 {
-                    if (send_result == INVALID_SOCKET)
+                    if (errno != EAGAIN)
```
rafilho: How can we add a test for this?

adapters/socketio_berkeley.c
```diff
@@ -910,32 +910,18 @@ int socketio_send(CONCRETE_IO_HANDLE socket_io, const void* buffer, size_t size,
...
-                        {
-                            LogError("Failure: sending socket failed. errno=%d (%s).", errno, strerror(errno));
-                            result = MU_FAILURE;
-                        }
+                        LogError("Failure: sending socket failed. errno=%d (%s).", errno, strerror(errno));
+                        result = MU_FAILURE;
                     }
                     else
                     {
+                        /*send says "come back later" with EAGAIN - likely the socket buffer cannot accept more data*/
```
hihigupt: >*s [](start = 25, length = 2)

nit: hs

adapters/socketio_berkeley.c
hihigupt: >0: [](start = 63, length = 2)

nit:hs

adapters/socketio_berkeley.c
hihigupt: Do we also need to handle EWOULDBLOCK because socket is created in with non-blocking flag?

---
In reply to: [326714516](https://github.com/Azure/azure-c-shared-utility/pull/380#discussion_r326714516) [](ancestors = 326714516,326645677)

ewertons: I agree we should perhaps add EWOULDBLOCK, but that is an additional change in a hot path, and EAGAIN has sufficed for all the Linux platforms we use. After offline discussion, we agreed on not changing that how, since would require extra validation across all linuxes we support.

ewertons: @rafilho , it has been confirmed that on error no bytes are sent.

adapters/socketio_berkeley.c
ewertons: It would require an E2E test for it. We do not have E2E test modules specific for socketio_*. They do get tested indirectly through other E2E tests, but non would exercise this particular scenario.

adapters/socketio_berkeley.c
ewertons: Logging in this case is the same scenario as logging within DoWork. It would generate way too many statements, and not add up much to the debugging assets. 
Also, I would not suggest logging expected/handled failures.

adapters/socketio_berkeley.c
ewertons: You got it!

PR #386, opened 2019-09-27
Fixes socketio_berkeley  send() in case not all bytes are sent in one shot

ewertons:
We didn't get this failure until DeviceStreaming e2e tests failed on public-preview.

### DIFF
--- a/adapters/socketio_berkeley.c
+++ b/adapters/socketio_berkeley.c
@@ -50,6 +50,7 @@
 
 #define SOCKET_SUCCESS                 0
 #define INVALID_SOCKET                 -1
+#define SOCKET_SEND_FAILURE            -1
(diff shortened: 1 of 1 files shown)


ewertons:
For reference: https://linux.die.net/man/3/send

adapters/socketio_berkeley.c
```diff
@@ -50,6 +50,7 @@
 
 #define SOCKET_SUCCESS                 0
 #define INVALID_SOCKET                 -1
+#define SOCKET_SEND_FAILURE            -1
```
hihigupt: >SOCKET_SEND_FAILURE [](start = 8, length = 19)

SOCKET_SEND_ERROR?

ewertons: Yes, that's not to confuse about errno, which is another channel for error codes.

PR #388, opened 2019-10-01
better logging of the lost etw messages

anporumb:


src/etwlogger_driver.c
```diff
@@ -140,25 +141,30 @@ static void lazyRegisterEventProvider(void)
...
-static volatile LONG doPrintf = 1;
-
+/*the number of lost messages (messages for which EventWriteXXX (all of them) apis did not succeeds. 
+Gets incremented at every failure. When one call succeeds, a printf is produced containing the number of missed events and the variable is reset to 0.*/
+static volatile LONG nLostMessages = 0;
 void perform_EventWriteLogErrorEvent(const char* content, const char* file, const SYSTEMTIME* t, const char* func, int line)
 {
     ULONG event_write_result = EventWriteLogErrorEvent(content, file, t, func, line);
     if (event_write_result != ERROR_SUCCESS)
     {
```
mattdurak: Maybe nice to put all of this logic in a common helper function

anporumb: The new updateLostMessages function will take care of that, thanks!

---
In reply to: [330273824](https://github.com/Azure/azure-c-shared-utility/pull/388#discussion_r330273824) [](ancestors = 330273824)

PR #391, opened 2019-10-03
[SDL] Use secure versions of SSL/TLS

rafilho:
Addressing SDL issue https://msazure.visualstudio.com/One/_workitems/edit/5240489. Details below:

WinHttp API:

By default, only SSL 3.0 and TLS 1.0 are enabled. Specify which versions of SSL and TLS to enable by setting the WINHTTP_OPTION_SECURE_PROTOCOLS option through the WinHttpSetOption function.

Following flags must be set
    WINHTTP_FLAG_SECURE_PROTOCOL_TLS1_2 (enable support for TLS1.2).
    WINHTTP_FLAG_SECURE_PROTOCOL_TLS1_1 (Optional if TLS 1.1 protocol support if required).
Following flags must not be set
    WINHTTP_FLAG_SECURE_PROTOCOL_SSL3
    WINHTTP_FLAG_SECURE_PROTOCOL_SSL2
    WINHTTP_FLAG_SECURE_PROTOCOL_ALL

### DIFF
--- a/adapters/httpapi_winhttp.c
+++ b/adapters/httpapi_winhttp.c
@@ -434,6 +434,9 @@ static HTTPAPI_RESULT InitiateWinhttpRequest(HTTP_HANDLE_DATA* handleData, HTTPA
     const wchar_t* requestTypeString;
     size_t requiredCharactersForRelativePath;
     wchar_t* relativePathTemp = NULL;
+    DWORD supported_protocols = WINHTTP_FLAG_SECURE_PROTOCOL_TLS1 | // TLS 1.0 is support for back-compat reasons (https://docs.microsoft.com/en-us/azure/iot-fundamentals/iot-security-deployment)
+                                WINHTTP_FLAG_SECURE_PROTOCOL_TLS1_1 | 
+                                WINHTTP_FLAG_SECURE_PROTOCOL_TLS1_2; 
     
     if ((requestTypeString = GetHttpRequestString(requestType)) == NULL)
     {
@@ -476,7 +479,20 @@ static HTTPAPI_RESULT InitiateWinhttpRequest(HTTP_HANDLE_DATA* handleData, HTTPA
     ))
     {
         result = HTTPAPI_SET_X509_FAILURE;
-        LogErrorWinHTTPWithGetLastErrorAsString("unable to WinHttpSetOption");
+        LogErrorWinHTTPWithGetLastErrorAsString("unable to WinHttpSetOption (WINHTTP_OPTION_CLIENT_CERT_CONTEXT)");
+        (void)WinHttpCloseHandle(*requestHandle);
+        *requestHandle = NULL;
+    }
+    else if ((handleData->x509SchannelHandle != NULL) &&
+        !WinHttpSetOption(
+            *requestHandle,
+            WINHTTP_OPTION_SECURE_PROTOCOLS,
+            &supported_protocols,
+            sizeof(supported_protocols)
+        ))
+    {
+        result = HTTPAPI_SET_X509_FAILURE;
+        LogErrorWinHTTPWithGetLastErrorAsString("unable to WinHttpSetOption (WINHTTP_OPTION_SECURE_PROTOCOLS)");
         (void)WinHttpCloseHandle(*requestHandle);
         *requestHandle = NULL;
     }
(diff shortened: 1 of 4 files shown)


adapters/httpapi_winhttp.c
```diff
@@ -434,6 +434,9 @@ static HTTPAPI_RESULT InitiateWinhttpRequest(HTTP_HANDLE_DATA* handleData, HTTPA
     const wchar_t* requestTypeString;
     size_t requiredCharactersForRelativePath;
     wchar_t* relativePathTemp = NULL;
+    DWORD supported_protocols = WINHTTP_FLAG_SECURE_PROTOCOL_TLS1 | // TLS 1.0 is support for back-compat reasons (https://docs.microsoft.com/en-us/azure/iot-fundamentals/iot-security-deployment)
```
jspaith: Could we just go TLS1_2?  The doc referenced suggests using 1.2 and since Windows has this, couldn't we just lock on this?

adapters/httpapi_winhttp.c
```diff
@@ -476,7 +479,20 @@ static HTTPAPI_RESULT InitiateWinhttpRequest(HTTP_HANDLE_DATA* handleData, HTTPA
     ))
     {
         result = HTTPAPI_SET_X509_FAILURE;
-        LogErrorWinHTTPWithGetLastErrorAsString("unable to WinHttpSetOption");
+        LogErrorWinHTTPWithGetLastErrorAsString("unable to WinHttpSetOption (WINHTTP_OPTION_CLIENT_CERT_CONTEXT)");
+        (void)WinHttpCloseHandle(*requestHandle);
+        *requestHandle = NULL;
+    }
+    else if ((handleData->x509SchannelHandle != NULL) &&
```
jspaith: why are we checking `handleData->x509SchannelHandle` here?  Shouldn't this be set unconditionally?

adapters/httpapi_winhttp.c
```diff
@@ -476,7 +479,20 @@ static HTTPAPI_RESULT InitiateWinhttpRequest(HTTP_HANDLE_DATA* handleData, HTTPA
...
+        !WinHttpSetOption(
+            *requestHandle,
+            WINHTTP_OPTION_SECURE_PROTOCOLS,
+            &supported_protocols,
+            sizeof(supported_protocols)
+        ))
+    {
+        result = HTTPAPI_SET_X509_FAILURE;
```
jspaith: Recommend different, more generic error code?

adapters/httpapi_winhttp.c
rafilho: As discussed offline with John, I will keep 1.0 for back compat reasons, given that the SDL instructions do not comdemn it. 

---
In reply to: [331298041](https://github.com/Azure/azure-c-shared-utility/pull/391#discussion_r331298041) [](ancestors = 331298041)

adapters/httpapi_winhttp.c
rafilho: Went with a different/specific error code.

---
In reply to: [331298895](https://github.com/Azure/azure-c-shared-utility/pull/391#discussion_r331298895) [](ancestors = 331298895)

inc/azure_c_shared_utility/httpapi.h
```diff
@@ -51,7 +51,9 @@ HTTPAPI_ALLOC_FAILED,                        \
 HTTPAPI_INIT_FAILED,                         \
 HTTPAPI_INSUFFICIENT_RESPONSE_BUFFER,        \
 HTTPAPI_SET_X509_FAILURE,                    \
-HTTPAPI_SET_TIMEOUTS_FAILED                  \
+HTTPAPI_SET_TIMEOUTS_FAILED,                 \
+HTTPAPI_SET_SECURE_PROTOCOLS_FAILURE         \
```
jspaith: >HTTPAPI_SET_SECURE_PROTOCOLS_FAILURE [](start = 0, length = 36)

I took a look and these error codes haven't been modified since at least 2016.  And this is for now pretty Windows specific.

Even though we don't make same back-compat promises on c-util as we do other places in SDK, i'd feel better leaving these as is and returning "HTTPAPI_ERROR" if the SetSecure fails.

adapters/httpapi_winhttp.c
jspaith: >supported_protocols [](start = 10, length = 19)

Nit: Other variables in this function follow the thisIsVariableName pattern.  "supportedProtocols' would be following convention instead of "_".

rafilho: good point. My bad. Confusing the guidelines. Will address it.

---
In reply to: [331714624](https://github.com/Azure/azure-c-shared-utility/pull/391#discussion_r331714624) [](ancestors = 331714624)

PR #394, opened 2019-10-07
Adding Visual Studio debug window logging.

CIPop:
This change makes logging to be displayed in the Output window when tests are executed in _Debug_ mode within Visual Studio.
Example: 
image

The output should not be available in any other mode (Windows Release or non-Windows).

### DIFF
--- a/src/consolelogger.c
+++ b/src/consolelogger.c
@@ -203,6 +203,9 @@ void consolelogger_log(LOG_CATEGORY log_category, const char* file, const char* 
     time_t t;
     va_list args;
     va_start(args, format);
+#if (defined(_MSC_VER) && defined(_DEBUG))
+    char* debugPrint;
+#endif
 
     t = time(NULL);
 
@@ -210,9 +213,19 @@ void consolelogger_log(LOG_CATEGORY log_category, const char* file, const char* 
     {
     case AZ_LOG_INFO:
         (void)printf("Info: ");
+#if (defined(_MSC_VER) && defined(_DEBUG))
+        OutputDebugStringA("Info: ");
+#endif
         break;
     case AZ_LOG_ERROR:
         (void)printf("Error: Time:%.24s File:%s Func:%s Line:%d ", ctime(&t), file, func, line);
+#if (defined(_MSC_VER) && defined(_DEBUG))
+        if ((debugPrint = printf_alloc("Error: Time:%.24s File:%s Func:%s Line:%d ", ctime(&t), file, func, line)) != NULL)
+        {
+            OutputDebugStringA(debugPrint);
+            free(debugPrint);
+        }
+#endif
         break;
     default:
         break;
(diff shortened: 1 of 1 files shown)


src/consolelogger.c
```diff
@@ -203,16 +203,29 @@ void consolelogger_log(LOG_CATEGORY log_category, const char* file, const char*
...
     case AZ_LOG_INFO:
         (void)printf("Info: ");
+#if (defined(_MSC_VER) && defined(_DEBUG))
+        OutputDebugStringA("Info: ");
+#endif
         break;
     case AZ_LOG_ERROR:
         (void)printf("Error: Time:%.24s File:%s Func:%s Line:%d ", ctime(&t), file, func, line);
+#if (defined(_MSC_VER) && defined(_DEBUG))
+        if ((debugPrint = printf_alloc("Error: Time:%.24s File:%s Func:%s Line:%d ", ctime(&t), file, func, line)) != NULL)
```
dcristoloveanu: >printf_alloc [](start = 26, length = 12)

sprintf_char does the same thing. Seems that this console logger should use that so that the code is not duplicated.

src/consolelogger.c
```diff
@@ -203,16 +203,29 @@ void consolelogger_log(LOG_CATEGORY log_category, const char* file, const char*
...
+        {
+            OutputDebugStringA(debugPrint);
+            free(debugPrint);
+        }
```
dcristoloveanu: > } [](start = 7, length = 2)

At least would be nice to have the branch with a comment in it that we're explicitly not going to report the error.

CIPop: I assume you mean the allocation error: good point: I think I'll actually print something to show that we failed alloc.

src/consolelogger.c
CIPop: Makes sense. In the end then we shouldn't have `vprintf_alloc` and `printf_alloc` in here since we already have a string_utils.

src/consolelogger.c
```diff
@@ -221,10 +234,23 @@ void consolelogger_log(LOG_CATEGORY log_category, const char* file, const char*
     (void)vprintf(format, args);
     va_end(args);
 
+#if (defined(_MSC_VER) && defined(_DEBUG))
+    va_start(args, format);
```
jspaith: >va_start(args, format); [](start = 4, length = 23)

My varargs is rusty, but do we need to do the va_end() above here and another va_start here?  Can we just re-use the same args or is something happening under covers after initial vprintf() on 234 that requires this reset?

src/consolelogger.c
```diff
@@ -221,10 +234,23 @@ void consolelogger_log(LOG_CATEGORY log_category, const char* file, const char*
     (void)vprintf(format, args);
     va_end(args);
 
+#if (defined(_MSC_VER) && defined(_DEBUG))
```
jspaith: >_MSC_VER [](start = 13, length = 8)

If someone builds this in non-Visual Studio compiler for Windows (and I know customers do this on our SDK), I think they'd still want OutputDebugString().  Could we change the _MSC_VER to be an #ifdef WINDOWS instead?

src/consolelogger.c
dcristoloveanu: Yep, that's what I meant :-). Cheers!

src/consolelogger.c
dcristoloveanu: A very interesting point. Yeah. mingw for the win :-). This should be done throughout the codebase seems like.

---
In reply to: [332286415](https://github.com/Azure/azure-c-shared-utility/pull/394#discussion_r332286415) [](ancestors = 332286415)

dcristoloveanu: As a separate task of course :-)

---
In reply to: [332300304](https://github.com/Azure/azure-c-shared-utility/pull/394#discussion_r332300304) [](ancestors = 332300304,332286415)

balarajupetta (approved):

ericwol-msft:
Not supporting debug VS output.

PR #400, opened 2019-10-18
fix segfault resolving dns

mlilien:
on alpine (armhf) the system crashed sometimes if connection was lost.


ewertons:
Hi @mlilien ,
thanks for the PR. There are a few more changes needed to address the failures with the new DNS resolver, so we will consolidate that in another PR and add your fix. 
Thank you for your contribution.

PR #401, opened 2019-10-22
use tracelogging

anporumb:


CMakeLists.txt
```diff
@@ -43,8 +43,8 @@ if(WIN32)
     option(logerror_captures_stack_traces "LogError will make a reasonable attempt to capture the stack trace at the time of calling LogError. Only available on Windows" OFF)
     option(use_socketio "set use_socketio to ON if socketio is to be included in the library, set to OFF if a different implementation will be provided" ON)
 
-    set(use_etw OFF CACHE STRING "set use_etw to ON if ETW logging is to be used (OFF = No ETW logging, ON = Use ETW logging, ON_WITH_CONSOLE = Use ETW and call console logger. Default is OFF")
-    set_property(CACHE use_etw PROPERTY STRINGS ON_WITH_CONSOLE ON OFF)
+    set(use_etw OFF CACHE STRING "set use_etw to ON if ETW logging is to be used (OFF = No ETW logging, ON = Use ETW logging, TRACELOGGING = Use ETW from tracelogging.h, ON_WITH_CONSOLE = Use ETW and call console logger. TRACELOGGING_WITH_CONSOLE = Use ETW from tracelogging.h and call console logger. Default is OFF")
```
dcristoloveanu: >Use ETW from tracelogging.h, ON_WITH_CONSOLE = Use ETW and call console logger. TRACELOGGING_WITH_CONSOLE = Use ETW  [](start = 141, length = 116)

Can we please have only the new one? Pretty please? Because otherwise we'd have to come back to remove it. And otherwise it just grows as hairs ....

src/etwlogger_driver.c
```diff
@@ -9,9 +9,16 @@
...
-#include "azure_c_shared_utility/etwlogger.h"
+
 #include "azure_c_shared_utility/consolelogger.h"
 
+#ifdef USE_TRACELOGGING
+#include "TraceLoggingProvider.h"
+#include "evntrace.h"
+#else
+#include "azure_c_shared_utility/etwlogger.h"
+#endif
```
dcristoloveanu: Like said earlier, let's have just the tracelogging flavor.
And I shall not say it anymore now :_).

CMakeLists.txt
anporumb: Absolutely agree, but in order to have an easy migration between "old" and "new" we shall keep both of them in the code. Enable the "new" one (this one), see that it works, and then remove the "old" one.

Will be addressed by a follow up task done after the migration to the new provider is completed.

---
In reply to: [337707334](https://github.com/Azure/azure-c-shared-utility/pull/401#discussion_r337707334) [](ancestors = 337707334)

src/etwlogger_driver.c
anporumb: Agree. Will be done with a later task.

---
In reply to: [337707663](https://github.com/Azure/azure-c-shared-utility/pull/401#discussion_r337707663) [](ancestors = 337707663)

PR #405, opened 2019-10-25
Removing incorrect buffer error message.

CIPop:
Unbuilding a buffer that is not null is the common case. The error was likely missed when running  `no_logging=OFF`.
Since the application cannot verify if the buffer is empty or not, unbuild should always work without errors.


src/buffer.c
```diff
@@ -325,7 +325,6 @@ extern int BUFFER_unbuild(BUFFER_HANDLE handle)
         BUFFER* b = (BUFFER*)handle;
         if (b->buffer != NULL)
         {
-            LogError("Failure buffer data is NULL");
```
ewertons: Instead of removing it completely, could you move it to the block below (that handles when it's actually null)?

CIPop: No, tried that already: see the original PR explanation. 
The buffer unbuild function is called for _any_ buffer - null or not. Since there is no way for the app to check if the buffer is null or not before calling (similar to `free(void*)`) we cannot error on either of these cases.

CIPop: (I've talked about this change offline with @jebrando - he may have some background on original design considerations. )

dcristoloveanu: Frankly that is a bad thing from uhttp. There is never a need to call "destroy"/"close"/"unbuild" with a NULL handle.

It can be convenience of design, but we can see how this convenience of design can lead to ignoring errors.
Ignoring errors is not good because it can lead to missing coding errors in other cases which leads to badly written code.

So I would suggest fixing this in the proper way (add the error when the handle is NULL).
And then uhttp has to also be fixed honestly.

---
In reply to: [339236793](https://github.com/Azure/azure-c-shared-utility/pull/405#discussion_r339236793) [](ancestors = 339236793)

dcristoloveanu: By "fixed in the right way" I mean move the log error between ln. 320 and 321.

---
In reply to: [339405110](https://github.com/Azure/azure-c-shared-utility/pull/405#discussion_r339405110) [](ancestors = 339405110,339236793)

ewertons: Like here:

BUFFER_HANDLE BUFFER_create(const unsigned char* source, size_t size)
{
    BUFFER* result;
    /*Codes_SRS_BUFFER_02_001: [If source is NULL then BUFFER_create shall return NULL.]*/
    if (source == NULL)
    {
        LogError("invalid parameter source: %p", source);
        result = NULL;
    }

---
In reply to: [339405202](https://github.com/Azure/azure-c-shared-utility/pull/405#discussion_r339405202) [](ancestors = 339405202,339405110,339236793)

adapters/socketio_berkeley.c
```diff
@@ -954,6 +954,7 @@ void socketio_dowork(CONCRETE_IO_HANDLE socket_io)
     if (socket_io != NULL)
     {
         SOCKET_IO_INSTANCE* socket_io_instance = (SOCKET_IO_INSTANCE*)socket_io;
+        signal(SIGPIPE, SIG_IGN);
```
ewertons: When this was added, we tested it manually to make sure it was caught. I would suggest doing the same. Permission to write a smiley face. :)

adapters/socketio_berkeley.c
```diff
@@ -93,7 +93,7 @@ typedef struct SOCKET_IO_INSTANCE_TAG
     char* target_mac_address;
     IO_STATE io_state;
     SINGLYLINKEDLIST_HANDLE pending_io_list;
-    unsigned char recv_bytes[RECEIVE_BYTES_VALUE];
+    unsigned char recv_bytes[XIO_RECEIVE_BYTES_VALUE];
```
ewertons: Please forgive me :)
Sharing another option for naming: XIO_RECEIVE_BUFFER_SIZE
This would be P237 in the priority scale.

adapters/socketio_berkeley.c
CIPop: Done the manual tests you've recommended. This seems to be missing an e2e test - we could create a mock TCP server. Please open a tracking GH issue if required.

CIPop:
> Maybe we should have a section in this doc explaining how to use the XIO_RECEIVE_BYTES_VALUE

@ewertons please see https://github.com/Azure/azure-iot-sdk-c/pull/1305

PR #408, opened 2019-11-08
Fixing porting guide broken link

CIPop:
The tickcounter spec is missing. Referencing the tickcounter_freertos to fix the broken link for now.


CIPop:
@gregman-msft can't find any trace of the original spec file. According to git annotate 740df0965 was the commit bringing the porting guide in our repo.

PR #415, opened 2019-11-21
Fix SEGFAULT errors introduced by dns_resolver in socketio (gh# 1275)

ewertons:
Relates to: https://github.com/Azure/azure-iot-sdk-c/issues/1275

There were two instances of possible failures not properly handled on socketio related to dns_resolver:
- The result of dns_resolver_create() was not being checked for NULL (which could be caused by memory allocation failure);
- The result of dns_resolver_get_addrInfo() was not being checked for NULL, which could happen even if dns_resolver_is_lookup_complete returned true.

These failures affected socketio_berkeley.c and socketio_win32.c.
The currently remaining socketio implementations were not affected, since they do not use dns_resolver (socketio_mbed.c, socketio_mbed_os5.c).

Thanks @mlilien for the initial PR to fix this issue.

### DIFF
--- a/adapters/socketio_berkeley.c
+++ b/adapters/socketio_berkeley.c
@@ -653,6 +653,31 @@ static int set_target_network_interface(int socket, char* mac_address)
 }
 #endif //__APPLE__
 
+static void destroy_socket_io_instance(SOCKET_IO_INSTANCE* instance)
+{
+    if (instance->dns_resolver != NULL)
+    {
+        dns_resolver_destroy(instance->dns_resolver);
+    }
+
+    if (instance->hostname != NULL)
+    {
+        free(instance->hostname);
+    }
+
+    if (instance->target_mac_address != NULL)
+    {
+        free(instance->target_mac_address);
+    }
+
+    if (instance->pending_io_list != NULL)
+    {
+        singlylinkedlist_destroy(instance->pending_io_list);
+    }
+
+    free(instance);
+}
+
 CONCRETE_IO_HANDLE socketio_create(void* io_create_parameters)
 {
     SOCKETIO_CONFIG* socket_io_config = io_create_parameters;
@@ -668,12 +693,14 @@ CONCRETE_IO_HANDLE socketio_create(void* io_create_parameters)
         result = malloc(sizeof(SOCKET_IO_INSTANCE));
         if (result != NULL)
         {
+            (void)memset(result, 0, sizeof(SOCKET_IO_INSTANCE));
+
             result->address_type = ADDRESS_TYPE_IP;
             result->pending_io_list = singlylinkedlist_create();
             if (result->pending_io_list == NULL)
             {
                 LogError("Failure: singlylinkedlist_create unable to create pending list.");
-                free(result);
+                destroy_socket_io_instance(result);
                 result = NULL;
             }
             else
@@ -697,36 +724,25 @@ CONCRETE_IO_HANDLE socketio_create(void* io_create_parameters)
                 if ((result->hostname == NULL) && (result->socket == INVALID_SOCKET))
                 {
                     LogError("Failure: hostname == NULL and socket is invalid.");
-                    singlylinkedlist_destroy(result->pending_io_list);
-                    free(result);
+                    destroy_socket_io_instance(result);
+                    result = NULL;
+                }
+                else if ((result->dns_resolver = dns_resolver_create(result->hostname, result->port, NULL)) == NULL)
+                {
+                    LogError("Failure creating dns_resolver");
+                    destroy_socket_io_instance(result);
                     result = NULL;
                 }
                 else
                 {
-                    result->dns_resolver = dns_resolver_create(result->hostname, result->port, NULL);
-
-                    if (result->dns_resolver == NULL)
-                    {
-                        LogError("Failure creating dns_resolver");
-                        if (result->hostname != NULL)
-                        {
-                            free(result->hostname);
-                        }
-                        singlylinkedlist_destroy(result->pending_io_list);
-                        free(result);
-                        result = NULL;
-                    }
-                    else
-                    {
-                        result->port = socket_io_config->port;
-                        result->on_io_open_complete = NULL;
-                        result->target_mac_address = NULL;
-                        result->on_bytes_received = NULL;
-                        result->on_io_error = NULL;
-                        result->on_bytes_received_context = NULL;
-                        result->on_io_error_context = NULL;
-                        result->io_state = IO_STATE_CLOSED;
-                    }
+                    result->port = socket_io_config->port;
+                    result->on_io_open_complete = NULL;
+                    result->target_mac_address = NULL;
+                    result->on_bytes_received = NULL;
+                    result->on_io_error = NULL;
+                    result->on_bytes_received_context = NULL;
+                    result->on_io_error_context = NULL;
+                    result->io_state = IO_STATE_CLOSED;
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


adapters/socketio_berkeley.c
```diff
@@ -694,15 +703,26 @@ CONCRETE_IO_HANDLE socketio_create(void* io_create_parameters)
...
-                    result->on_bytes_received = NULL;
-                    result->on_io_error = NULL;
-                    result->on_bytes_received_context = NULL;
-                    result->on_io_error_context = NULL;
-                    result->io_state = IO_STATE_CLOSED;
+
+                    if (result->dns_resolver == NULL)
+                    {
+                        LogError("Failure creating dns_resolver");
+                        singlylinkedlist_destroy(result->pending_io_list);
```
danewalton: free `hostname` here?

danewalton: more specifically `result->hostname`

adapters/socketio_win32.c
```diff
@@ -342,17 +351,40 @@ CONCRETE_IO_HANDLE socketio_create(void* io_create_parameters)
...
-                        result->on_io_error = NULL;
-                        result->on_bytes_received_context = NULL;
-                        result->on_io_error_context = NULL;
-                        result->io_state = IO_STATE_CLOSED;
-                        result->keep_alive = tcp_keepalive;
+                        if (result->addrInfo->ai_addr == NULL)
+                        {
+                            LogError("Failure allocating ai_addr");
+                            free(result->addrInfo);
+                            singlylinkedlist_destroy(result->pending_io_list);
```
jebrando: >singlylinkedlist_destroy(result->pending_io_list); [](start = 28, length = 50)

You need to add the following on all failure items
if (result->hostname != NULL)
{
      free(result->hostname);
}

adapters/socketio_win32.c
```diff
@@ -342,17 +351,40 @@ CONCRETE_IO_HANDLE socketio_create(void* io_create_parameters)
...
-                        result->on_bytes_received = NULL;
-                        result->on_io_error = NULL;
-                        result->on_bytes_received_context = NULL;
-                        result->on_io_error_context = NULL;
-                        result->io_state = IO_STATE_CLOSED;
-                        result->keep_alive = tcp_keepalive;
+                        if (result->addrInfo->ai_addr == NULL)
+                        {
+                            LogError("Failure allocating ai_addr");
+                            free(result->addrInfo);
```
danewalton: `result->hostname` here too?

ewertons: Fixed.

adapters/socketio_win32.c
ewertons: fixed.

adapters/socketio_win32.c
```diff
@@ -331,9 +339,14 @@ CONCRETE_IO_HANDLE socketio_create(void* io_create_parameters)
                 else
                 {
                     result->addrInfo = calloc(1, sizeof(struct addrinfo));
+
                     if (result->addrInfo == NULL)
                     {
                         LogError("Failure: addrInfo == NULL.");
+                        if (result->hostname != NULL)
+                        {
+                            free(result->hostname);
```
jspaith: >free(result->hostname); [](start = 28, length = 23)

For here and in Berkley equivalent, could we please have a common cleanup path instead of having the free's in multiple places?  More likely to be issues here since we don't have UT error mocking on c-util stuff.

adapters/socketio_win32.c
```diff
@@ -331,9 +339,14 @@ CONCRETE_IO_HANDLE socketio_create(void* io_create_parameters)
                 else
```
jspaith: >else [](start = 16, length = 4)

Nit: the number of nested else's is getting deep.  Any chance we can move more to the 
else if (dosomething() == error)
{
   fail
}
else if (doNextSomething == error) 
{
...
}

Where it's a bit flatter?

adapters/socketio_win32.c
ewertons: I thought about that, but will do it separately as it would add more change to a sensitive component.

adapters/socketio_win32.c
ewertons: I like that way too. Shall be done.

adapters/socketio_berkeley.c
ewertons: Fixed.

adapters/socketio_win32.c
ewertons: Ok, I take it back. I'll do it.

adapters/socketio_win32.c
```diff
@@ -342,17 +355,48 @@ CONCRETE_IO_HANDLE socketio_create(void* io_create_parameters)
...
+                        else
+                        {
+                            result->dns_resolver = dns_resolver_create(result->hostname, result->port, NULL);
+
+                            if (result->dns_resolver == NULL)
+                            {
+                                LogError("Failure creating dns_resolver");
+                                free(result->addrInfo->ai_addr);
+                                free(result->addrInfo);
+                                if (result->hostname != NULL)
```
danewalton: technically the check whether it is `NULL` shouldn't be needed. it's a nop if it is `NULL`.

ewertons: We will need now, with the aux function I've added.

adapters/socketio_win32.c
ewertons: Done.

adapters/socketio_berkeley.c
```diff
@@ -644,6 +653,31 @@ static int set_target_network_interface(int socket, char* mac_address)
...
+    {
+        dns_resolver_destroy(instance->dns_resolver);
+    }
+
+    if (instance->hostname != NULL)
+    {
+        free(instance->hostname);
+    }
+
+    if (instance->target_mac_address != NULL)
```
danewalton: nit nit: i still don't think you need the checks but it's not a blocking issue.

adapters/socketio_win32.c
```diff
@@ -324,35 +364,37 @@ CONCRETE_IO_HANDLE socketio_create(void* io_create_parameters)
...
+                    destroy_socket_io_instance(result);
+                    result = NULL;
+                }
+                else if ((result->addrInfo->ai_addr = calloc(1, sizeof(struct sockaddr_in))) == NULL)
+                {
+                    LogError("Failure allocating ai_addr");
+                    destroy_socket_io_instance(result);
+                    result = NULL;
+                }
+                else if ((result->dns_resolver = dns_resolver_create(result->hostname, result->port, NULL)) == NULL)
```
hihigupt: >port [](start = 95, length = 4)

port =0? it should be config->port right?

adapters/socketio_berkeley.c
```diff
@@ -688,15 +724,19 @@ CONCRETE_IO_HANDLE socketio_create(void* io_create_parameters)
                 if ((result->hostname == NULL) && (result->socket == INVALID_SOCKET))
                 {
                     LogError("Failure: hostname == NULL and socket is invalid.");
-                    singlylinkedlist_destroy(result->pending_io_list);
-                    free(result);
+                    destroy_socket_io_instance(result);
+                    result = NULL;
+                }
+                else if ((result->dns_resolver = dns_resolver_create(result->hostname, result->port, NULL)) == NULL)
```
hihigupt: >port [](start = 95, length = 4)

same this value is not initialized

adapters/socketio_win32.c
```diff
@@ -324,35 +364,37 @@ CONCRETE_IO_HANDLE socketio_create(void* io_create_parameters)
...
+                {
+                    LogError("Failure creating dns_resolver");
```
hihigupt: >LogError("Failure creating dns_resolver"); [](start = 20, length = 42)

is it possible to run these port with sample... just to check the flow?

adapters/socketio_berkeley.c
ewertons: You are correct. From the c99 spec, "If ptr is a null pointer, no action occurs."
So there should be no check.

adapters/socketio_win32.c
ewertons: You are absolutely right. Fixed.
Thanks, you saved me here.

adapters/socketio_berkeley.c
ewertons: Fixed.

adapters/socketio_win32.c
ewertons: E2E tests will validate that.

PR #420, opened 2019-12-03
loop on mbedtls_ssl_write to handle partial writes

alexmrtn:
Current implementation of mbedtls tlsio caps the write to 16384 bytes. This limitation comes from [TLS RFC 6066](https://tools.ietf.org/html/rfc6066#section-4).

From the [mbedtls spec](https://tls.mbed.org/api/ssl_8h.html#a5bbda87d484de82df730758b475f32e5) of function `mbedtls_ssl_write`:

> This function will do partial writes in some cases. If the return value is non-negative but less than length, the function must be called again with updated arguments: buf + ret, len - ret (if ret is the return value) until it returns a value equal to the last 'len' argument.

This PR address this current limitation.

### DIFF
--- a/adapters/tlsio_mbedtls.c
+++ b/adapters/tlsio_mbedtls.c
@@ -63,6 +63,8 @@ typedef struct TLS_IO_INSTANCE_TAG
     unsigned char *socket_io_read_bytes;
     size_t socket_io_read_byte_count;
     SEND_COMPLETE_INFO send_complete_info;
+    size_t socket_io_write_byte_count;
+    size_t socket_io_total_byte_to_write;
 
     mbedtls_entropy_context entropy;
     mbedtls_ctr_drbg_context ctr_drbg;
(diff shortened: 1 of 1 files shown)


adapters/tlsio_mbedtls.c
```diff
@@ -677,10 +677,19 @@ int tlsio_mbedtls_send(CONCRETE_IO_HANDLE tls_io, const void *buffer, size_t siz
...
             tls_io_instance->send_complete_info.send_complete_count = 0;
-            int res = mbedtls_ssl_write(&tls_io_instance->ssl, buffer, size);
-            if (res != (int)size)
+
+            int bytes_sent = 0;
+            int res = 0;
+
+            do{
+                res = mbedtls_ssl_write(&tls_io_instance->ssl, buffer+bytes_sent, size-bytes_sent);
+                bytes_sent += res;
```
hihigupt: This will cause on_send_complete to be fired  on first chunk of data sent successfully ... but this should be opposite on_send_complete should fire when all the data push has been complete.

Check line  347 of this file.

alexmrtn: Did not see this problem... 

I added a `on_partial_send_complete` function with counters to call the actual `on_send_complete` function only when all chunk are sent. What do you think?

hihigupt: I have create a PR for this fix
https://github.com/Azure/azure-c-shared-utility/pull/421

Let me know if you still see this issue after applying my commit.

alexmrtn:
Closing as #421 fix this issue

Issue #381, opened 2019-09-20
tlsio_mbed.c leaks ~6000 bytes


uriel-kluk:
SDK 1.3.4
[path]\azure-iot-sdk-c\c-utility\adapters
for the adaptation file `tlsio_mbed.c` when compiled with `use_mbedtls` option

If the device authenticates using x509 and the owner certificate is provided, the `mbedtls_uninit `function does not release the allocated chain.

Add the following line to the function to resolve it.
` mbedtls_x509_crt_free(&tls_io_instance->owncert);`

This is the modified function

```
// Un-initialize mbedTLS
static void mbedtls_uninit(TLS_IO_INSTANCE *tls_io_instance)
{
    if (tls_io_instance->tls_status != TLS_STATE_NOT_INITIALIZED)
    {
        // mbedTLS cleanup...
        mbedtls_x509_crt_free(&tls_io_instance->trusted_certificates_parsed);
        //This is needed to clear the Ecc certificate
        mbedtls_x509_crt_free(&tls_io_instance->owncert);  //<== ADD this line or leak occurs
        mbedtls_pk_free(&tls_io_instance->pKey);
        mbedtls_ssl_free(&tls_io_instance->ssl);
        mbedtls_ssl_config_free(&tls_io_instance->config);
        mbedtls_ctr_drbg_free(&tls_io_instance->ctr_drbg);
        mbedtls_entropy_free(&tls_io_instance->entropy);


        tls_io_instance->tls_status = TLS_STATE_NOT_INITIALIZED;
    }
    else
    {
        LogError("Uninitialzing when not previously initialized");
    }
}
```

hihigupt:
Thanks for reporting this. Changes are now merged into the azure-c-shared-utility master branch.

az-iot-builder-01:
@uriel-kluk, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

Issue #376, opened 2019-09-18
Meaningless condition in tlsio_mbedtls.c


DavidJurajdaNXP:
### Environment:
mbed TLS 2.16.2 branch released 2019-06-11
azure iot c sdk 1.3.4

### Issue descripton:
tlsio_mbedtls.c line: 989
`&tls_io_instance->owncert != NULL`

Struct member owncert is struct (not pointer) => address will be always non null value, the condition above is meaningless.

same issue apply for 
tlsio_mbedtls.c line: 993
`&tls_io_instance->pKey != NULL`

### Additional note

Structs are declared in following locations:

azure_iot\c-utility\adapters\tlsio_mbedtls.c
TLS_IO_INSTANCE           struct declared in tlsio_mbedtls, line 72

mbedtls\include\mbedtls\x509_crt.h
mbedtls_x509_crt          struct declared on line 53

mbedtls\include\mbedtls\pk.h
mbedtls_pk_context        struct declared on line 130

hihigupt:
Thanks for reporting this. Changes are now merged into the azure-c-shared-utility master branch.

az-iot-builder-01:
@DavidJurajdaNXP, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #422, opened 2019-12-10
[tlsio_mbedtls] Fix memory leak

hihigupt:
Fixing memory leak in tlsio_mbedtls.c

Issues: https://github.com/Azure/azure-c-shared-utility/issues/381
https://github.com/Azure/azure-c-shared-utility/issues/376

### DIFF
--- a/adapters/tlsio_mbedtls.c
+++ b/adapters/tlsio_mbedtls.c
@@ -443,6 +443,8 @@ static void mbedtls_uninit(TLS_IO_INSTANCE *tls_io_instance)
         mbedtls_ssl_free(&tls_io_instance->ssl);
         mbedtls_ssl_config_free(&tls_io_instance->config);
         mbedtls_x509_crt_free(&tls_io_instance->trusted_certificates_parsed);
+        mbedtls_x509_crt_free(&tls_io_instance->owncert);
+        mbedtls_pk_free(&tls_io_instance->pKey);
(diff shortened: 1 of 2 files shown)


adapters/tlsio_mbedtls.c
```diff
@@ -1068,17 +1073,15 @@ OPTIONHANDLER_HANDLE tlsio_mbedtls_retrieveoptions(CONCRETE_IO_HANDLE handle)
                 OptionHandler_Destroy(result);
                 result = NULL;
             }
-            else if (&tls_io_instance->owncert != NULL && tls_io_instance->x509_certificate != NULL &&
-                    OptionHandler_AddOption(result, SU_OPTION_X509_CERT, tls_io_instance->x509_certificate) != OPTIONHANDLER_OK)
+            else if (tls_io_instance->x509_certificate != NULL &&
```
danewalton: >tls_io_instance->x509_certificate [](start = 21, length = 33)

is `own_cert` not being used in this case?

hihigupt: x509_certificate gets set when own_cert is already set. 

Also checking address of allocated structure (own_cert is memory part of tls_io_instance structure) and comparing it to NULL is always false.


---
In reply to: [356730641](https://github.com/Azure/azure-c-shared-utility/pull/422#discussion_r356730641) [](ancestors = 356730641)

danewalton: ahh i see okay

PR #440, opened 2020-03-17
Malloc to Calloc calls for 0 initialize padding bytes

ttins:


ttins:
Do not merge until [https://github.com/Azure/azure-iot-sdk-c/pull/1455] has been merged

adapters/condition_win32.c
```diff
@@ -20,12 +20,12 @@ CONDITION;
 COND_HANDLE Condition_Init(void)
 {
     // Codes_SRS_CONDITION_18_002: [ Condition_Init shall create and return a CONDITION_HANDLE ]
-
     CONDITION* cond = (CONDITION*)malloc(sizeof(CONDITION));
 
     // Codes_SRS_CONDITION_18_008: [ Condition_Init shall return NULL if it fails to allocate the CONDITION_HANDLE ]
     if (cond != NULL)
     {
+        (void)memset(cond, 0, sizeof(CONDITION));
```
danewalton: nit: I prefer doing `sizeof({var_name})`. In this case:
```C
(void)memset(cond, 0, sizeof(cond));
```
Reason being if the type of `cond` were to ever change, the `sizeof()` call gets resolved correctly and you don't get out of sync.

ttins: I prefer to have that one calloced, but based on the UT and how it is mocking the memory management functions and redefining them 3 times it was currently out of our scope as it gets passed into umock as a mock function it seems to default to using malloc where it should be calling calloc so something is amiss in how umock is mocking it

danewalton: no i meant specifically the use of `sizeof()` not memset

PR #445, opened 2020-03-31
Ttins/bug 43496

ttins:
Fixes for bugs:
4349662 & 6084598
4349643
4349642

### DIFF
--- a/adapters/x509_schannel.c
+++ b/adapters/x509_schannel.c
@@ -38,7 +38,7 @@ static const size_t pem_crlf_value_length = sizeof(pem_crlf_value) - 1;
 static unsigned char* convert_cert_to_binary(const char* crypt_value, DWORD crypt_value_in_len, DWORD* crypt_length)
 {
     unsigned char* result = NULL;
-    DWORD result_length;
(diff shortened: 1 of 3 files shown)


adapters/x509_schannel.c
```diff
@@ -45,7 +45,7 @@ static unsigned char* convert_cert_to_binary(const char* crypt_value, DWORD cryp
         LogErrorWinHTTPWithGetLastErrorAsString("Failed determine crypt value size");
         result = NULL;
     }
-    else if (result_length != 0)
```
ericwol-msft, line 48: Add a log error message like the other parts of the function

src/sha1.c
```diff
@@ -117,41 +117,51 @@ int SHA1Reset(SHA1Context *context)
 *          the message.
 *      length: [in]
 *          The length of the message in message_array
+*          Max length can be 64 bits
```
ericwol-msft, line 120: I don't get what this commit says

ttins: Removed, meant to say max value of length

PR #452, opened 2020-04-29
Add proper handling of options retrieval in Http Proxy IO

ewertons:
HTTP proxy IO tries to simplify the retrieve_options() function by returning the result of the
underlying I/O layer retrieve_options() directly, but that does not work when upper layer (TLS I/O)
tries to reapply the saved options.

### DIFF
--- a/src/http_proxy_io.c
+++ b/src/http_proxy_io.c
@@ -923,7 +923,7 @@ static int http_proxy_io_set_option(CONCRETE_IO_HANDLE http_proxy_io, const char
         else if (xio_setoption(http_proxy_io_instance->underlying_io, option_name, value) != 0)
         {
             /* Codes_SRS_HTTP_PROXY_IO_01_044: [ if xio_setoption fails, http_proxy_io_set_option shall return a non-zero value. ]*/
-            LogError("Unrecognized option");
+            LogError("Unrecognized option %s", option_name);
             result = MU_FAILURE;
         }
         else
@@ -944,7 +944,7 @@ static void* http_proxy_io_clone_option(const char* name, const void* value)
         (name == NULL) || (value == NULL)
         )
     {
-        LogError("invalid parameter detected: const char* name=%p, const void* value=%p", name, value);
+        LogError("invalid parameter detected: name=%p, value=%p", name, value);
         result = NULL;
     }
     else
--- a/src/http_proxy_io.c
+++ b/src/http_proxy_io.c
@@ -13,6 +13,8 @@
 #include "azure_c_shared_utility/http_proxy_io.h"
 #include "azure_c_shared_utility/azure_base64.h"
 
+static const char* const OPTION_UNDERLYING_IO_OPTIONS = "underlying_io_options";
+
 typedef enum HTTP_PROXY_IO_STATE_TAG
 {
     HTTP_PROXY_IO_STATE_CLOSED,
@@ -896,7 +898,7 @@ static int http_proxy_io_set_option(CONCRETE_IO_HANDLE http_proxy_io, const char
         /* Codes_SRS_HTTP_PROXY_IO_01_040: [ If any of the arguments http_proxy_io or option_name is NULL, http_proxy_io_set_option shall return a non-zero value. ]*/
         LogError("Bad arguments: http_proxy_io = %p, option_name = %p",
             http_proxy_io, option_name);
-        result = __LINE__;
+        result = MU_FAILURE;
     }
     else
     {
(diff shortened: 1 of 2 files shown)


src/http_proxy_io.c
```diff
@@ -896,21 +898,33 @@ static int http_proxy_io_set_option(CONCRETE_IO_HANDLE http_proxy_io, const char
...
     {
         HTTP_PROXY_IO_INSTANCE* http_proxy_io_instance = (HTTP_PROXY_IO_INSTANCE*)http_proxy_io;
 
         /* Codes_SRS_HTTP_PROXY_IO_01_045: [ None. ]*/
 
+        if (strcmp(option_name, OPTION_UNDERLYING_IO_OPTIONS) == 0)
+        {
+            if (OptionHandler_FeedOptions((OPTIONHANDLER_HANDLE)value, (void*)http_proxy_io_instance->underlying_io) != OPTIONHANDLER_OK)
+            {
+                LogError("failed feeding options to underlying I/O instance");
```
ericwol-msft, line 913: first work should be CAPS -> Failed

src/http_proxy_io.c
```diff
@@ -896,21 +898,33 @@ static int http_proxy_io_set_option(CONCRETE_IO_HANDLE http_proxy_io, const char
...
+                result = 0;
+            }
+        }
         /* Codes_SRS_HTTP_PROXY_IO_01_043: [ If the option_name argument indicates an option that is not handled by http_proxy_io_set_option, then xio_setoption shall be called on the underlying IO created in http_proxy_io_create, passing the option name and value to it. ]*/
         /* Codes_SRS_HTTP_PROXY_IO_01_056: [ The value argument shall be allowed to be NULL. ]*/
-        if (xio_setoption(http_proxy_io_instance->underlying_io, option_name, value) != 0)
+        else if (xio_setoption(http_proxy_io_instance->underlying_io, option_name, value) != 0)
         {
             /* Codes_SRS_HTTP_PROXY_IO_01_044: [ if xio_setoption fails, http_proxy_io_set_option shall return a non-zero value. ]*/
             LogError("Unrecognized option");
```
ericwol-msft, line 926: add option name to this log error?

src/http_proxy_io.c
```diff
@@ -922,6 +936,56 @@ static int http_proxy_io_set_option(CONCRETE_IO_HANDLE http_proxy_io, const char
...
 
+/*this function will clone an option given by name and value*/
+static void* http_proxy_io_clone_option(const char* name, const void* value)
+{
+    void* result;
+    if (
+        (name == NULL) || (value == NULL)
+        )
+    {
+        LogError("invalid parameter detected: const char* name=%p, const void* value=%p", name, value);
```
ericwol-msft, line 947: don't need to specify the consts here

src/http_proxy_io.c
```diff
@@ -922,6 +936,56 @@ static int http_proxy_io_set_option(CONCRETE_IO_HANDLE http_proxy_io, const char
     return result;
 }
 
+/*this function will clone an option given by name and value*/
+static void* http_proxy_io_clone_option(const char* name, const void* value)
```
ericwol-msft, line 940: const char* const name,

ericwol-msft: I don't see how this is actually "cloning" anything?

src/http_proxy_io.c
```diff
@@ -922,6 +936,56 @@ static int http_proxy_io_set_option(CONCRETE_IO_HANDLE http_proxy_io, const char
...
+            LogError("not handled option : %s", name);
+            result = NULL;
+        }
+    }
+
+    return result;
+}
+
+/*this function destroys an option previously created*/
+static void http_proxy_io_destroy_option(const char* name, const void* value)
```
ericwol-msft, line 967: const char* const name,

src/http_proxy_io.c
```diff
@@ -922,6 +936,56 @@ static int http_proxy_io_set_option(CONCRETE_IO_HANDLE http_proxy_io, const char
...
+    }
+    else
+    {
+        if (strcmp(name, OPTION_UNDERLYING_IO_OPTIONS) == 0)
+        {
+            OptionHandler_Destroy((OPTIONHANDLER_HANDLE)value);
+        }
+        else
+        {
+            LogError("not handled option : %s", name);
```
ericwol-msft, line 984: first word caps

ericwol-msft: extra space before the :

ericwol-msft: option not handled?

src/http_proxy_io.c
ewertons: Seems like a minor ask, but that is different than the rest of the logs throughout the SDK. LogError itself already prints an "Error" tag, so I rather keep it consistent for now.

src/http_proxy_io.c
ewertons: Yeah that's a good idea. Doing it.

src/http_proxy_io.c
ewertons: True. C&P side effect. Fixed.

src/http_proxy_io.c
ewertons: That is different than the function pointer definition for OptionHandler. Will keep as is unless we change those fps.

src/http_proxy_io.c
ewertons: Same here, that would differ from the fp defined for OptionHandler_Create.

src/http_proxy_io.c
ewertons: Keeping it consistent with the rest of the SDK.

src/http_proxy_io.c
```diff
@@ -896,21 +898,33 @@ static int http_proxy_io_set_option(CONCRETE_IO_HANDLE http_proxy_io, const char
...
+            }
+            else
+            {
+                result = 0;
+            }
+        }
         /* Codes_SRS_HTTP_PROXY_IO_01_043: [ If the option_name argument indicates an option that is not handled by http_proxy_io_set_option, then xio_setoption shall be called on the underlying IO created in http_proxy_io_create, passing the option name and value to it. ]*/
         /* Codes_SRS_HTTP_PROXY_IO_01_056: [ The value argument shall be allowed to be NULL. ]*/
-        if (xio_setoption(http_proxy_io_instance->underlying_io, option_name, value) != 0)
+        else if (xio_setoption(http_proxy_io_instance->underlying_io, option_name, value) != 0)
```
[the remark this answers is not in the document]
ewertons: No it shouldn't. The way options work in the SDK is: the current layer tries to consume it, otherwise pass it to the underlying layer.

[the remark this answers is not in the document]
ewertons: No, the way options work on the SDK is: the current layer tries to consume the option, otherwise pass it to the underlying layer.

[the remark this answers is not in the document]
danewalton: gotcha thats right. Okay thanks

PR #456, opened 2020-07-04
Fix redirection of errno header

sergioprado:
The POSIX correct header to include is <errno.h>.

The glibc silently redirects incorrect includes, but musl libc prints
a warning:
   #warning redirecting incorrect #include <sys/errno.h> to <errno.h>

Signed-off-by: Sergio Prado <sergio.prado@e-labworks.com>


danewalton:
This has already been updated in the code thank you for the contribution!

PR #447, opened 2020-04-02
fix library ref and casts in tlsio_mbedtls

alexmrtn:
Hi,

Small fixes to use mbedtls library on windows,

it fix cast errors and libraries missing `.lib` extention in cmakefile

Regards,

PR #457, opened 2020-07-08
fix library ref and casts in tlsio_mbedtls

ewertons:
This updates and replaces https://github.com/Azure/azure-c-shared-utility/pull/447

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -535,10 +535,10 @@ endif()
 
 if(${use_mbedtls})
     if (WIN32)
-        set(aziotsharedutil_target_libs ${aziotsharedutil_target_libs} 
-            $ENV{mbedtlsLibDir}/mbedtls 
-            $ENV{mbedtlsLibDir}/mbedx509 
-            $ENV{mbedtlsLibDir}/mbedcrypto
+        set(aziotsharedutil_target_libs ${aziotsharedutil_target_libs} ws2_32
+            $ENV{mbedtlsLibDir}/mbedtls.lib 
+            $ENV{mbedtlsLibDir}/mbedx509.lib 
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -535,10 +535,10 @@ endif()
 
 if(${use_mbedtls})
     if (WIN32)
-        set(aziotsharedutil_target_libs ${aziotsharedutil_target_libs} 
-            $ENV{mbedtlsLibDir}/mbedtls 
-            $ENV{mbedtlsLibDir}/mbedx509 
-            $ENV{mbedtlsLibDir}/mbedcrypto
+        set(aziotsharedutil_target_libs ${aziotsharedutil_target_libs} ws2_32
```
danewalton, line 538: what is `ws2_32`?

adapters/platform_win32.c
```diff
@@ -20,6 +20,9 @@
 #if USE_WOLFSSL
 #include "azure_c_shared_utility/tlsio_wolfssl.h"
 #endif
+#if USE_MBEDTLS
```
danewalton, line 23: ideally should these all be `#if` and `#elif`?

CMakeLists.txt
ewertons: Windows socket library.
https://docs.microsoft.com/en-us/windows/win32/winsock/creating-a-basic-winsock-application

adapters/platform_win32.c
ewertons: We could. But the effect would be similar anyway if only one define is provided. 
On the other side, if more than one is provided, we better try to include everything and let the compilation fail, so customers know right away what is wrong.

PR #460, opened 2020-08-03
Update call to get private key for compatibility with OpenSSL@1.1

makononov:
This changes updates the method used to get the PK ID to make it compatible with OpenSSL@1.1

### DIFF
--- a/adapters/x509_openssl.c
+++ b/adapters/x509_openssl.c
@@ -13,7 +13,7 @@
 
 #ifdef __APPLE__
     #ifndef EVP_PKEY_id
-        #define EVP_PKEY_id(evp_key) evp_key->type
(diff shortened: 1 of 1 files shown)


adapters/x509_openssl.c
```diff
@@ -13,7 +13,7 @@
 
 #ifdef __APPLE__
     #ifndef EVP_PKEY_id
-        #define EVP_PKEY_id(evp_key) evp_key->type
+	#define EVP_PKEY_ID(evp_key) EVP_PKEY_get1_EC_KEY(evp_key)
```
CIPop, line 16: I ran into this issue myself - at this point we are only testing with OpenSSL 1.0 for OSX.

Would this work for RSA keys as well?

danewalton:
Does this solution work as well for you?

https://github.com/Azure/azure-iot-sdk-c/issues/2026#issuecomment-897927362

PR #462, opened 2020-08-10
fix wolfssl retry bug

fanzhe98:
fix wolfssl retry bug

### DIFF
--- a/adapters/tlsio_wolfssl.c
+++ b/adapters/tlsio_wolfssl.c
@@ -56,17 +56,6 @@ typedef struct TLS_IO_INSTANCE_TAG
 STATIC_VAR_UNUSED const char* const OPTION_WOLFSSL_SET_DEVICE_ID = "SetDeviceId";
 static const size_t SOCKET_READ_LIMIT = 5;
 
-static int mallocAndIntcpy_s(int** destination, const int* src)
-{
-    *destination = (int*) malloc(sizeof(int));
-    if(*destination == NULL)
-    {
-        return ENOMEM;
-    }
-    **destination = *src;
-    return 0; 
-}
-
 /*this function will clone an option given by name and value*/
 static void* tlsio_wolfssl_CloneOption(const char* name, const void* value)
 {
@@ -114,18 +103,23 @@ static void* tlsio_wolfssl_CloneOption(const char* name, const void* value)
                 /*return as is*/
             }
         }
+        #ifdef INVALID_DEVID
         else if(strcmp(name, OPTION_WOLFSSL_SET_DEVICE_ID) == 0 )
         {
-            if (mallocAndIntcpy_s((int**)&result, value) !=0 )
-            {
-                LogError("unable to mallocAndIntcpy_s device id value");
-                result = NULL;
-            }
-            else
-            {
-                /*return as is*/
-            }
+             int* value_clone; 
+  
+             if ((value_clone = (int*)malloc(sizeof(int))) == NULL) 
+             { 
+                 LogError("unable to clone device id option"); 
+             } 
+             else 
+             { 
+                 *value_clone = *value; 
+             } 
+
+             result = value_clone; 
         }
+        #endif
         else
         {
             LogError("not handled option : %s", name);
@@ -206,6 +200,7 @@ static OPTIONHANDLER_HANDLE tlsio_wolfssl_retrieveoptions(CONCRETE_IO_HANDLE tls
                 OptionHandler_Destroy(result);
                 result = NULL;
             }
+            #ifdef INVALID_DEVID
             else if (
                 (tls_io_instance->wolfssl_device_id != INVALID_DEVID) &&
                 (OptionHandler_AddOption(result, OPTION_WOLFSSL_SET_DEVICE_ID, &tls_io_instance->wolfssl_device_id) != OPTIONHANDLER_OK)
@@ -215,6 +210,7 @@ static OPTIONHANDLER_HANDLE tlsio_wolfssl_retrieveoptions(CONCRETE_IO_HANDLE tls
                 OptionHandler_Destroy(result);
                 result = NULL;
             }
+            #endif
             else
             {
                 /*all is fine, all interesting options have been saved*/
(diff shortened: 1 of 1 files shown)


momuno:
@fanzhe98 I have the task to validate and merge this fix.  Can you tell me more about how the bug was found initially so I know what I'm testing against?  I don't see any linked issues. Thanks.

adapters/tlsio_wolfssl.c
```diff
@@ -103,6 +114,18 @@ static void* tlsio_wolfssl_CloneOption(const char* name, const void* value)
                 /*return as is*/
             }
         }
+        else if(strcmp(name, OPTION_WOLFSSL_SET_DEVICE_ID) == 0 )
```
momuno, line 117: We should be fine if these added logic sections (line 117, and line 151) that don't mention INVALID_DEVID aren't wrapped in an `#ifdef INVALID_DEVID ... #endif`.  But I will double check with my colleague what he thinks about long term maintainability and safety.
https://github.com/Azure/azure-c-shared-utility/blob/ef6ebf91a2e3530275050f0f789bab9f744c45e9/adapters/tlsio_wolfssl.c#L915-L917

momuno (changes requested):
Thanks for your work tracking down and resolving this issue!
I am still in testing.

fanzhe98:
@momuno please see the latest change

adapters/tlsio_wolfssl.c
```diff
@@ -103,6 +103,23 @@ static void* tlsio_wolfssl_CloneOption(const char* name, const void* value)
                 /*return as is*/
             }
         }
+        #ifdef INVALID_DEVID
+        else if(strcmp(name, OPTION_WOLFSSL_SET_DEVICE_ID) == 0 )
+        {
+             int* value_clone; 
+  
+             if ((value_clone = (int*)malloc(sizeof(int))) == NULL)
```
ewertons, line 111: A minor suggestion that would align with the usual coding standard we **try** to maintain: the return from master should not be cast (backed by guidance in the C99 language spec).
It would result in:
```
if ((value_clone = malloc(sizeof(int))) == NULL) 
```

ewertons (approved):

adapters/tlsio_wolfssl.c
```diff
@@ -103,6 +103,23 @@ static void* tlsio_wolfssl_CloneOption(const char* name, const void* value)
...
+        {
+             int* value_clone; 
+  
+             if ((value_clone = (int*)malloc(sizeof(int))) == NULL) 
+             { 
+                 LogError("unable to clone device id option"); 
+             } 
+             else 
+             { 
+                 *value_clone = *value;
```
momuno, line 117: This is not building because cannot dereference a void pointer.  Please cast to int pointer first:

```
*value_clone = *(int*)value; 
```

fanzhe98: resolved

momuno:
Since it is not possible for me to validate these updated changes on the Sphere (I don't have access to your build environment currently), can you please test the updated code and confirm it works?  I believe I have encountered a different bug that does not allow me to test the changes on Linux. Thank you!

fanzhe98:
> Since it is not possible for me to validate these updated changes on the Sphere (I don't have access to your build environment currently), can you please test the updated code and confirm it works? Is there a way you can forward me that OS test bundle to install on my sphere so I can validate? Thank you!

Hi Mollie, I believe this is affecting all the customers who are using the C SDK with wolfssl. Verifying on our OS test bundle is not comprehensive enough for this task since we also modify some of our code to adapt our OS components. I could do so if this is the last resort. @ewertons, Could you please help @momuno here? Thanks!

momuno:
> > Since it is not possible for me to validate these updated changes on the Sphere (I don't have access to your build environment currently), can you please test the updated code and confirm it works? Is there a way you can forward me that OS test bundle to install on my sphere so I can validate? Thank you!
> 
> Hi Mollie, I believe this is affecting all the customers who are using the C SDK with wolfssl. Verifying on our OS test bundle is not comprehensive enough for this task since we also modify some of our code to adapt our OS components. I could do so if this is the last resort. @ewertons, Could you please help @momuno here? Thanks!

Since the PR changes have been updated, it would be helpful if you could please validate those changes still work as expected for the Azure Sphere.  I am currently blocked at testing on Linux due to a newly discovered bug with wolfssl.  I will also talk to my team about what other device options we have to test on.  Thanks Fanzhe for your work on this.

fanzhe98:
> > > Since it is not possible for me to validate these updated changes on the Sphere (I don't have access to your build environment currently), can you please test the updated code and confirm it works? Is there a way you can forward me that OS test bundle to install on my sphere so I can validate? Thank you!
> > 
> > 
> > Hi Mollie, I believe this is affecting all the customers who are using the C SDK with wolfssl. Verifying on our OS test bundle is not comprehensive enough for this task since we also modify some of our code to adapt our OS components. I could do so if this is the last resort. @ewertons, Could you please help @momuno here? Thanks!
> 
> Since the PR changes have been updated, it would be helpful if you could please validate those changes still work as expected for the Azure Sphere. I am currently blocked at testing on Linux due to a newly discovered bug with wolfssl. I will also talk to my team about what other device options we have to test on. Thanks Fanzhe for your work on this.

It works. Again, even though this is a data point, you should not rely on whether the patch is working on Azure Sphere or not.

momuno (approved):

momuno:
Issue addressed:
Toggling wifi enable/disable state on Azure Sphere, the device will not reconnect.
On a PC, reconnection will happen.

Testing:
Above code changes tested on Azure Sphere and fix the issue.
On a PC, the bug cannot be reproduced, which aligns with the original issue described.

Will merge and close.

fanzhe98:
> Issue addressed:
> Toggling wifi enable/disable state on Azure Sphere, the device will not reconnect.
> On a PC, reconnection will happen.
> 
> Testing:
> Above code changes tested on Azure Sphere and fix the issue.
> On a PC, the bug cannot be reproduced, which aligns with the original issue described.
> 
> Will merge and close.

I believe this bug is targeting a general usage of wolfssl instead of only problems on Azure Sphere. Even though verifying on Azure Sphere provide a significant data point, I think it is better for you to test on your build environment. Anyway, this should be your group's call.

PR #466, opened 2020-08-11
Fix for bug-7579476

momuno:
While memory was allocated for buffer_size + size + 1, memcpy only copied up to buffer_size + size.

I would think writing 0 to the last position of the reallocated memory would be fine, but based on the reported details of the bug there was still a complaint:
"Buffer overrun while writing to 'http_proxy_io_instance->receive_buffer': **the writable size is 'http_proxy_io_instance->receive_buffer_size+size+1' bytes**, but 'http_proxy_io_instance->receive_buffer_size' bytes might be written."


momuno:
> I believe that will break the functionality.
Ignore.  Change reverted.

PR #467, opened 2020-08-11
add check for overflow for BUFFER_prepend

danewalton:


src/buffer.c
```diff
@@ -540,32 +540,40 @@ int BUFFER_prepend(BUFFER_HANDLE handle1, BUFFER_HANDLE handle2)
...
             {
                 // do nothing
                 result = 0;
             }
             else
             {
-                // b2->size != 0
-                unsigned char* temp = (unsigned char*)malloc(b1->size + b2->size);
-                if (temp == NULL)
+                if (b1->size + b2->size < b2->size)
```
ericwol-msft, line 550: move this out one level...else if

PR #472, opened 2020-10-20
fixed openssl include for protocol fuzzing

ericwol-msft:
This is needed so that we can build the fuzzer under linux, without openssl and socketio enabled

### DIFF
--- a/adapters/platform_linux.c
+++ b/adapters/platform_linux.c
@@ -7,6 +7,8 @@
 #include "azure_c_shared_utility/httpapiex.h"
 #ifdef USE_OPENSSL
 #include "azure_c_shared_utility/tlsio_openssl.h"
+#else
+const IO_INTERFACE_DESCRIPTION* tlsio_openssl_get_interface_description();
 #endif
 #if USE_CYCLONESSL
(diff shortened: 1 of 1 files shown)


adapters/platform_linux.c
```diff
@@ -7,6 +7,8 @@
 #include "azure_c_shared_utility/httpapiex.h"
 #ifdef USE_OPENSSL
 #include "azure_c_shared_utility/tlsio_openssl.h"
+#else
+const IO_INTERFACE_DESCRIPTION* tlsio_openssl_get_interface_description();
```
danewalton, line 11: wouldn't this need the definition which is in the header?

ericwol-msft: only if USE_OPENSSL is used, which will include it in tlsio_openssl.h"

danewalton: If this is used for fuzzing, should it be called something other than openssl? Or is it actually somehow using openssl?

ericwol-msft: note this is needed because of line 61

danewalton (approved):
looks good to me but maybe @ewertons weigh in here?

momuno (approved):

PR #473, opened 2020-10-21
Fix for wolfssl failure to reconnect

momuno:
Address Issue [#1704](https://github.com/Azure/azure-iot-sdk-c/issues/1704) of the azure-iot-sdk-c.

`on_io_send` is registered to be the function called for `wolfssl_write()`. (See `tlsio_wolfssl.c`, `tlsio_wolfssl_create()`: `wolfSSL_SetIOSend(result->ssl_context, on_io_send);`)

From the wolfssl [sample](https://github.com/wolfSSL/wolfssl-examples/blob/master/embedded/tls-client-server.c) implementing a similar pattern, the correct return value is `WOLFSSL_CBIO_ERR_GENERAL` (which is -1); From this [sample](https://github.com/wolfSSL/wolfssl-examples/blob/master/custom-io-callbacks/file-server/file-server.c), we see the return value is based on the result of `write`, so on error the value returned needs to be negative, and not 0.  A return of 0 means nothing was written, and not that an actual error occurred. 

Please see issue [#1704](https://github.com/Azure/azure-iot-sdk-c/issues/1704) for instructions on how to reproduce the bug and test this fix.

### DIFF
--- a/adapters/tlsio_wolfssl.c
+++ b/adapters/tlsio_wolfssl.c
@@ -422,7 +422,7 @@ static int on_io_send(WOLFSSL *ssl, char *buf, int sz, void *context)
         LogError("Failed sending bytes through underlying IO");
         tls_io_instance->tlsio_state = TLSIO_STATE_ERROR;
         indicate_error(tls_io_instance);
-        result = 0;
+        result = -1;
     }
     else
     {
--- a/adapters/tlsio_wolfssl.c
+++ b/adapters/tlsio_wolfssl.c
@@ -422,7 +422,7 @@ static int on_io_send(WOLFSSL *ssl, char *buf, int sz, void *context)
         LogError("Failed sending bytes through underlying IO");
         tls_io_instance->tlsio_state = TLSIO_STATE_ERROR;
         indicate_error(tls_io_instance);
-        result = -1;
+        result = WOLFSSL_CBIO_ERR_WANT_WRITE;
     }
     else
     {
--- a/adapters/tlsio_wolfssl.c
+++ b/adapters/tlsio_wolfssl.c
@@ -422,7 +422,7 @@ static int on_io_send(WOLFSSL *ssl, char *buf, int sz, void *context)
         LogError("Failed sending bytes through underlying IO");
         tls_io_instance->tlsio_state = TLSIO_STATE_ERROR;
         indicate_error(tls_io_instance);
-        result = WOLFSSL_CBIO_ERR_WANT_WRITE;
+        result = WOLFSSL_CBIO_ERR_GENERAL;
     }
     else
     {


adapters/tlsio_wolfssl.c
```diff
@@ -422,7 +422,7 @@ static int on_io_send(WOLFSSL *ssl, char *buf, int sz, void *context)
         LogError("Failed sending bytes through underlying IO");
         tls_io_instance->tlsio_state = TLSIO_STATE_ERROR;
         indicate_error(tls_io_instance);
-        result = 0;
+        result = WOLFSSL_CBIO_ERR_WANT_WRITE;
```
CIPop, line 425: Shouldn't this be `sz` (assuming it is negative).

According to https://www.wolfssl.com/wolfssl-alternative-io-support/ the error code might have a different meaning: 
> If the “cipher-send” buffer isn’t big enough, the callback should return -2 (WOLFSSL_CBIO_ERR_WANT_WRITE).

momuno: I noticed it was used in that sample for buffer not being the correct size, but did not find a description for the usage of WOLFSSL_CBIO_ERR_WANT_WRITE specifically.  I interpreted it being used there as an example of its usage, but is not limited to that scenario. 

We cannot assume on `xio_send` failure that the size returned is negative.  Size could still be 0 or positive based on `socketio_send`

momuno: I think the options are either to return a negative number explicitly, like '-1' per the CBIOSend sample (https://github.com/wolfSSL/wolfssl-examples/blob/master/custom-io-callbacks/file-server/file-server.c), or use the WOLFSSL error macro per the send_client sample (https://github.com/wolfSSL/wolfssl-examples/blob/master/custom-io-callbacks/file-server/file-server.c).

momuno: I think the options are either to return a negative number explicitly, like '-1' per the CBIOSend sample (https://github.com/wolfSSL/wolfssl-examples/blob/master/custom-io-callbacks/file-server/file-server.c), or use the WOLFSSL error macro per the send_client sample (https://github.com/wolfSSL/wolfssl-examples/blob/master/custom-io-callbacks/file-server/file-server.c).

momuno: https://github.com/wolfSSL/wolfssl-examples/blob/2f048ad8dc0c556f021b614bc48e2dce2596fe18/custom-io-callbacks/file-server/file-server.c#L87-L102

https://github.com/wolfSSL/wolfssl-examples/blob/2f048ad8dc0c556f021b614bc48e2dce2596fe18/embedded/tls-client-server.c#L84-L99

momuno: WOLFSSL_CBIO_ERR_GENERAL // -1

CIPop (approved):
Thanks @momuno !

PR #476, opened 2020-10-30
add ecc support in httpapi_compact

YannGirard:
Hi,

This PR is designed to add ECC support for httpapi_compact. It is similar to the one made for curl ([#120](https://github.com/Azure/azure-c-shared-utility/pull/120))

The underlying function in X509_schannel.c handle the RSA and ECC certificates but the options were missing in the SetOption and CloneOption functions of httpapi_compact implementation of HTTPAPI.


ewertons (approved):
This is good. For reference, the other http api adapters do exactly the same.

PR #479, opened 2020-11-30
patch to suppress sigpipe

fanzhe98:
patch to suppress sigpipe

https://github.com/Azure/azure-iot-sdk-c/issues/1423


fanzhe98:
for patch#11 discussed in the last meeting

cc @suhuruli @momuno

ewertons:
This change is good to go, with it just SIGPIPE is suppressed.
Source: https://www.man7.org/linux/man-pages/man2/send.2.html 

Since we have a function in the tlsio for suppressing SIGPIPE, this commit wouldn't change the current expected behavior.

ewertons (approved):

ewertons:
@fanzhe98 , this change fails to build on OS-X (it uses the same socket adapter).
For addressing that, one way is to add a define at the top of socketio_berkeley.c:

```c
// MSG_NOSIGNAL is not defined on all platforms where socketio_berkeley.c is used (e.g., OS-X).
// Setting MSG_NOSIGNAL to 0 results in not suppressing SIGPIPE.
#ifndef MSG_NOSIGNAL
#define MSG_NOSIGNAL 0 
#endif
```

If you do need to have the signal suppressed on OS-X as well, this would be a recourse to be added:
https://stackoverflow.com/questions/19509348/sigpipe-osx-and-disconnected-sockets

fanzhe98:
@ewertons we do not support development on Os-X yet. 

I will go ahead and add the lines of ifdef later today

fanzhe98:
code updated @ewertons

Issue #480, opened 2020-12-03
UNUSED parameter in process_option under tlsio_wolfssl.c


fanzhe98:
UNUSED parameter in process_option under tlsio_wolfssl.c

fanzhe98:
Moved from https://github.com/Azure/azure-iot-sdk-c/issues/1777
cc. @suhuruli @momuno

PR #481, opened 2020-12-03
decorate unused parameter

fanzhe98:
decorate unused parameter in tlsio_wolfssl.c #480


momuno (approved):

momuno:
@fanzhe98 this is causing the build to fail.  Why is it passing in the build Azure Sphere is using?

fanzhe98:
@momuno yeah, possibly the depedencies in azure sphere that made this work though in 
xlogging.h there is ``` #define UNUSED(x) (void)(x)```

changing it to```void(name)``` should be the same

eddited.

momuno:
> @momuno yeah, possibly the depedencies in azure sphere that made this work though in
> xlogging.h there is ` #define UNUSED(x) (void)(x)`
> 
> changing it to`void(name)` should be the same
> 
> eddited.

I see tlsio_wolfssl.c has
```
#include "azure_c_shared_utility/xlogging.h"
```
peculiar why it didn't compile.  Thanks for making the update.

Issue #482, opened 2020-12-03
[Azure Sphere] removing dependency on libuuid


fanzhe98:
To save flash space on our OS. we replace uuid generation with similar logics.

fanzhe98:
issue for patch#8 as dicussed in the last meeting. 
cc @suhuruli @momuno

momuno:
Closing due to PR not being merged.

PR #483, opened 2020-12-03
remove libuuid dependency

fanzhe98:
remove libuuid dependency #482


fanzhe98:
cc @suhuruli @momuno

fanzhe98:
Waiting for an agreement on ifdef

adapters/uniqueid_linux.c
```diff
@@ -25,13 +39,36 @@ UNIQUEID_RESULT UniqueId_Generate(char* uid, size_t len)
...
+            /* Codes_SRS_UNIQUEID_07_001: [UniqueId_Generate shall create a unique Id 36 character long string.] */
+            // write out unique_id as string
+            memset(uid, 0, len);
+            int chars_written = snprintf(uid, len, uuid_format,
+                unique_id.time_low, unique_id.time_mid, unique_id.time_hi_and_version,
+                unique_id.clock_seq_and_variant >> 8, unique_id.clock_seq_and_variant & 0xFF,
+                unique_id.node[0], unique_id.node[1], unique_id.node[2],
+                unique_id.node[3], unique_id.node[4], unique_id.node[5]);
+
+            if (chars_written != 36) {
```
hihigupt, line 65: what this magic number represent ?

adapters/uniqueid_linux.c
```diff
@@ -4,14 +4,28 @@
...
 
 MU_DEFINE_ENUM_STRINGS(UNIQUEID_RESULT, UNIQUEID_RESULT_VALUES);
 
+// UUID fields as specified in RFC 4122
+struct uuid {
+    uint32_t    time_low;
+    uint16_t    time_mid;
+    uint16_t    time_hi_and_version;
+    uint16_t    clock_seq_and_variant;
+    uint8_t node[6];
```
hihigupt, line 21: nit: align to other member in this struct.

adapters/uniqueid_linux.c
```diff
@@ -4,14 +4,28 @@
...
+#include <sys/random.h>
 
 #include "azure_macro_utils/macro_utils.h"
 #include "azure_c_shared_utility/uniqueid.h"
 #include "azure_c_shared_utility/xlogging.h"
 
 MU_DEFINE_ENUM_STRINGS(UNIQUEID_RESULT, UNIQUEID_RESULT_VALUES);
 
+// UUID fields as specified in RFC 4122
+struct uuid {
```
hihigupt, line 16: Can we define struct in sync with other code in sdk?

Something like this:

typedef struct UUID_TAG
{
    uint32_t    time_low;
    uint16_t    time_mid;
    uint16_t    time_hi_and_version;
    uint16_t    clock_seq_and_variant;
    uint8_t      node[6];
} UUID;

https://github.com/Azure/azure-c-shared-utility/blob/b38adefd7d0f94d4118a1dc80b285f571a6bb8e1/adapters/socketio_mbed.c#L26

adapters/uniqueid_linux.c
```diff
@@ -4,14 +4,28 @@
...
+    uint32_t    time_low;
+    uint16_t    time_mid;
+    uint16_t    time_hi_and_version;
+    uint16_t    clock_seq_and_variant;
+    uint8_t node[6];
+};
+
+// lowercase uuid format expected by azure iot sdk
+static const char *uuid_format =
```
hihigupt, line 25: Nit: * with type
static const char* uuid_format

adapters/uniqueid_linux.c
```diff
@@ -25,13 +39,36 @@ UNIQUEID_RESULT UniqueId_Generate(char* uid, size_t len)
...
 
-        /* Codes_SRS_UNIQUEID_07_001: [UniqueId_Generate shall create a unique Id 36 character long string.] */
-        memset(uid, 0, len);
-        uuid_unparse(uuidVal, uid);
-        result = UNIQUEID_OK;
+        // generate version 4 uuid
+        struct uuid unique_id;
+
+        // get random bytes
+        ssize_t bytes_got = getrandom(&unique_id, sizeof(unique_id), 0);
```
hihigupt, line 47: earlier behavior of uuid_generate was to never block even if the high-quality randomness from /dev/urandom is not available or initialized, because it use to fallback on alternative algorithm.
"The uuid will be generated based on high-quality randomness from /dev/urandom, if available. If it is not available, then uuid_generate will use an alternative algorithm which uses the current time, the local ethernet MAC address (if available), and random data generated using a pseudo-random generator. "

But with flag zero to function from sys/random.h, it will block if the high quality random is not initialized. Should we fail fast by adding flag GRND_NONBLOCK?

momuno:
Thanks @hihigupt for taking an initial look at this PR

adapters/uniqueid_linux.c
```diff
@@ -25,13 +39,36 @@ UNIQUEID_RESULT UniqueId_Generate(char* uid, size_t len)
...
+        if (bytes_got != sizeof(unique_id)) {
+            LogError("Failed to obtain random numbers from getrandom.");
+            result = UNIQUEID_ERROR;
+        } else {
```
ericwol-msft, line 51:
```suggestion
        } 
        else 
        {
```

adapters/uniqueid_linux.c
```diff
@@ -25,13 +39,36 @@ UNIQUEID_RESULT UniqueId_Generate(char* uid, size_t len)
...
+            // format as version 4 (random) uuid variant 2
+            unique_id.clock_seq_and_variant = (unique_id.clock_seq_and_variant & 0x3FFF) | 0x8000;
+            unique_id.time_hi_and_version = (unique_id.time_hi_and_version & 0x0FFF) | 0x4000;
+
+            /* Codes_SRS_UNIQUEID_07_001: [UniqueId_Generate shall create a unique Id 36 character long string.] */
+            // write out unique_id as string
+            memset(uid, 0, len);
```
ericwol-msft, line 58: why memset?

fanzhe98:
PR Comments addressed

momuno:
What unit tests do you have developed for this on the Azure Sphere side?  Do you have anything that could be ported to the C SDK unit tests?

adapters/uniqueid_linux.c
```diff
@@ -25,13 +39,38 @@ UNIQUEID_RESULT UniqueId_Generate(char* uid, size_t len)
...
-        /* Codes_SRS_UNIQUEID_07_001: [UniqueId_Generate shall create a unique Id 36 character long string.] */
-        memset(uid, 0, len);
-        uuid_unparse(uuidVal, uid);
-        result = UNIQUEID_OK;
+        // generate version 4 uuid
+        UUID unique_id;
+
+        // get random bytes
+        ssize_t bytes_got = getrandom(&unique_id, sizeof(unique_id), GRND_NONBLOCK);
+        if (bytes_got != sizeof(unique_id)) {
```
hihigupt, line 48: Add newline
if ()
{
}

adapters/uniqueid_linux.c
```diff
@@ -25,13 +39,38 @@ UNIQUEID_RESULT UniqueId_Generate(char* uid, size_t len)
...
+            int chars_written = snprintf(uid, len, uuid_format,
+                unique_id.time_low, unique_id.time_mid, unique_id.time_hi_and_version,
+                unique_id.clock_seq_and_variant >> 8, unique_id.clock_seq_and_variant & 0xFF,
+                unique_id.node[0], unique_id.node[1], unique_id.node[2],
+                unique_id.node[3], unique_id.node[4], unique_id.node[5]);
+
+            if (chars_written != 36) {
+                LogError("Failed to convert binary uuid to string format.");
+                result = UNIQUEID_ERROR;
+            } else {
```
hihigupt, line 70: nit: all if/else, blocks start from new line

adapters/uniqueid_linux.c
```diff
@@ -25,13 +39,38 @@ UNIQUEID_RESULT UniqueId_Generate(char* uid, size_t len)
     }
     else
     {
-        uuid_t uuidVal;
-        uuid_generate(uuidVal);
```
momuno, line 42: There is a UUID_generate function in src/uuid.h.  Have you tried using this function instead?  Also, a custom UUID_unparse may be beneficial there.  That way others can use the functionality, and this file has very little change.  Those functions would then need unit tests to accompany them.

adapters/uniqueid_linux.c
momuno: I see your note above now and in man uuid_unparse that the final string length is 36 chars + trailing '\0'.  I would use a MACRO here to give context. And replace the above 37 on line 35 too.
```
#define UNIQUE_ID_REQ_LENGTH 37
...
if (uid == NULL || len <  UNIQUE_ID_REQ_LENGTH)
...
if (chars_written != (UNIQUE_ID_REQ_LENGTH - 1)) 
{

}
```

adapters/uniqueid_linux.c
```diff
@@ -25,13 +39,38 @@ UNIQUEID_RESULT UniqueId_Generate(char* uid, size_t len)
...
+
+        // get random bytes
+        ssize_t bytes_got = getrandom(&unique_id, sizeof(unique_id), GRND_NONBLOCK);
+        if (bytes_got != sizeof(unique_id)) {
+            LogError("Failed to obtain random numbers from getrandom.");
+            result = UNIQUEID_ERROR;
+        }
+        else
+        {
+            // format as version 4 (random) uuid variant 2
```
momuno, line 54: I would like to see this "unparse" section as a separate function.  It will also need a unit test.
Our unit tests for UniqueId_Generate only check for expected final string length or null input. 
Unit tests for this are found in tests/uniqueid_ut/uniqueid_ut_linux.c

adapters/uniqueid_linux.c
momuno, line 48: Can you also specify the case where -1 could be returned?  (I get that size of unique_id won't be this, but I think it is more clear)

adapters/uniqueid_linux.c
```diff
@@ -4,14 +4,28 @@
...
 
 MU_DEFINE_ENUM_STRINGS(UNIQUEID_RESULT, UNIQUEID_RESULT_VALUES);
 
+// UUID fields as specified in RFC 4122
+typedef struct UUID_TAG {
+    uint32_t    time_low;
+    uint16_t    time_mid;
+    uint16_t    time_hi_and_version;
+    uint16_t    clock_seq_and_variant;
+    uint8_t     node[6];
```
momuno, line 21: i see two fields in place of `clock_seq_and_variant` when I view the RFC:
```
    uint8_t clock_seq_hi_and_reserved;
    uint8_t clock_seq_low;
```
    
```
typedef struct {
    unsigned32  time_low;
    unsigned16  time_mid;
    unsigned16  time_hi_and_version;
    unsigned8   clock_seq_hi_and_reserved;
    unsigned8   clock_seq_low;
    byte        node[6];
} uuid_t;
```

momuno:
This will not be absorbed into C SDK for backwards compatibility assurance.  Looking at alternative approach for Azure Sphere to maintain their required changes.

PR #486, opened 2020-12-07
check kCFStreamStatusError on macos tlsio

yunhaoling:
See issue: https://github.com/Azure/azure-c-shared-utility/issues/487


danewalton (approved):
thanks for the addition @yunhaoling

PR #494, opened 2020-12-22
Addressing the Y2k38 issue for SAS tokens

CIPop:
Changing from `size_t` (32 bit on 32bit systems) to `uint64_t` for SAS time related operations to avoid the https://en.wikipedia.org/wiki/Year_2038_problem .


inc/azure_c_shared_utility/crt_abstractions.h
```diff
@@ -95,6 +97,7 @@ long double strtold_s(const char* nptr, char** endPtr);
 MOCKABLE_FUNCTION(, int, mallocAndStrcpy_s, char**, destination, const char*, source);
 MOCKABLE_FUNCTION(, int, unsignedIntToString, char*, destination, size_t, destinationSize, unsigned int, value);
 MOCKABLE_FUNCTION(, int, size_tToString, char*, destination, size_t, destinationSize, size_t, value);
+MOCKABLE_FUNCTION(, int, uint64_tToString, char*, destination, size_t, destinationSize, uint64_t, value);
```
ewertons, line 100: What if the platform does not support uint64_t? 
Maybe it would be safer to define a uint64_t as size_t if it is not defined.

src/crt_abstractions.c
```diff
@@ -835,3 +836,54 @@ int size_tToString(char* destination, size_t destinationSize, size_t value)
...
+            destination[pos] = '\0';
+            /*all converted and they fit*/
+            for (w = 0; w <= (pos - 1) >> 1; w++)
+            {
+                char temp;
+                temp = destination[w];
+                destination[w] = destination[pos - 1 - w];
+                destination[pos - 1 - w] = temp;
+            }
+            /*Codes_SRS_CRT_ABSTRACTIONS_02_004: [If the conversion has been successfull then unsignedIntToString shall return 0.] */
```
ewertons, line 879: These "Codes" statements do not belong to this function implementation. Should be removed.

src/crt_abstractions.c
```diff
@@ -835,3 +836,54 @@ int size_tToString(char* destination, size_t destinationSize, size_t value)
...
+            result = 0;
+        }
+        else
+        {
+            /*Codes_SRS_CRT_ABSTRACTIONS_02_002: [If the conversion fails for any reason (for example, insufficient buffer space), a non-zero return value shall be supplied and unsignedIntToString shall fail.] */
```
ewertons, line 884: This requirement statement does not apply to this code, it should be removed.

src/crt_abstractions.c
```diff
@@ -835,3 +836,54 @@ int size_tToString(char* destination, size_t destinationSize, size_t value)
...
+            destination[pos++] = '0' + (value % 10);
+            value /= 10;
+        } while ((value > 0) && (pos < (destinationSize - 1)));
+
+        if (value == 0)
+        {
+            size_t w;
+            destination[pos] = '\0';
+            /*all converted and they fit*/
+            for (w = 0; w <= (pos - 1) >> 1; w++)
```
ewertons, line 872: Suggestion: If this code is used by other functions, maybe it could be extracted as a reverseArray function.

inc/azure_c_shared_utility/crt_abstractions.h
ewertons: However, we use uint64_t already throughout the C SDK. So I think it would not be an issue.

tests/sastoken_ut/sastoken_ut.c
```diff
@@ -1793,4 +1824,51 @@ TEST_FUNCTION(SASToken_CreateString_succeeds)
     ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
 }
 
+TEST_FUNCTION(SASToken_CreateString_far_future_succeeds)
```
ewertons, line 1827: "far_future" alone does not give the full idea of what we are testing here.
Could you add a comment on top: // Test added to verify code against Y2k38 issue.

ewertons (approved):
Leaving comments but approving to unblock.

src/crt_abstractions.c
CIPop: The same code is used several times already in this file. I agree that we may want to optimize crt_abstractions.c later on but I'd like to separate that change from this PR.

src/crt_abstractions.c
CIPop: The same code is used several times already in this file. I agree that we may want to optimize crt_abstractions.c later on but I'd like to separate that change from this PR.

tests/sastoken_ut/sastoken_ut.c
CIPop: added comment and changed to long_int instead of far_future.

PR #496, opened 2021-01-06
Fix for C SDK Issue segfault when using wolfSSL and MQTT/WS

momuno:
This is a fix for C SDK Issue #1759 [https://github.com/Azure/azure-iot-sdk-c/issues/1759].

It involves two fixes:
1) Addresses the segfault using wolfSSL and MQTT/WS.  A callback was not reset to NULL and was mistakenly being called resulting in the segfault. (It should only be called on application data).
2) Updates logic to not return WANT_READ during renegotiation when waiting to receive CONNACK.

This change requires https://github.com/wolfSSL/wolfssl/pull/3594 to be merged, and the following wolfSSL release to be used. (v4.7.0 to be released mid-February)
Assuming this has occurred, wolfSSL must be configured with --enable-secure-renegotiation.

Enabling secure renegotiation does not obstruct using MQTT or AMQP.  It DOES enable MQTT/WS and AMQP/WS.

### DIFF
--- a/adapters/tlsio_wolfssl.c
+++ b/adapters/tlsio_wolfssl.c
@@ -429,11 +429,15 @@ static int on_io_recv(WOLFSSL *ssl, char *buf, int sz, void *context)
         }
         else if ( (result == 0) && (tls_io_instance->tlsio_state == TLSIO_STATE_OPEN))
         {
-            if ( wolfSSL_get_state(tls_io_instance->ssl) >= 8) // SERVER_HELLODONE_COMPLETE
+#ifdef HAVE_SECURE_RENEGOTIATION
+            if (wolfSSL_SSL_renegotiate_pending(tls_io_instance->ssl) == 0) // SERVER_HELLODONE_COMPLETE
             {
                 result = WOLFSSL_CBIO_ERR_WANT_READ;
             }
-            // Otherwise if Server Hello not complete during renegotiation, do not return error.
+            // If Server Hello not complete during renegotiation, do not return error.
+#else
+            result = WOLFSSL_CBIO_ERR_WANT_READ;
+#endif
         }
         else if ((result == 0) && (tls_io_instance->tlsio_state == TLSIO_STATE_CLOSING || tls_io_instance->tlsio_state == TLSIO_STATE_NOT_OPEN))
         {
@@ -933,7 +937,7 @@ static int process_option(char** destination, const char* name, const char* valu
 {
 
     (void) name;
-    
+
     int result;
     if (*destination != NULL)
     {
--- a/adapters/tlsio_wolfssl.c
+++ b/adapters/tlsio_wolfssl.c
@@ -108,18 +108,18 @@ static void* tlsio_wolfssl_CloneOption(const char* name, const void* value)
         #ifdef INVALID_DEVID
         else if(strcmp(name, OPTION_WOLFSSL_SET_DEVICE_ID) == 0 )
         {
-             int* value_clone; 
-  
-             if ((value_clone = malloc(sizeof(int))) == NULL) 
-             { 
-                 LogError("unable to clone device id option"); 
-             } 
-             else 
-             { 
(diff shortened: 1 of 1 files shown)


dgarske (approved):
@momuno  These changes look good.

adapters/tlsio_wolfssl.c
```diff
@@ -389,7 +389,8 @@ static int on_io_recv(WOLFSSL *ssl, char *buf, int sz, void *context)
         while (tls_io_instance->socket_io_read_byte_count == 0 && socket_reads < SOCKET_READ_LIMIT)
         {
             xio_dowork(tls_io_instance->socket_io);
-            if (tls_io_instance->tlsio_state != TLSIO_STATE_IN_HANDSHAKE)
+            if (tls_io_instance->tlsio_state != TLSIO_STATE_IN_HANDSHAKE &&
+                tls_io_instance->tlsio_state != TLSIO_STATE_OPEN) // Renegotiation
             {
                 break;
```
danewalton, line 395: `LogInfo` here?

adapters/tlsio_wolfssl.c
```diff
@@ -562,6 +565,14 @@ static int create_wolfssl_instance(TLS_IO_INSTANCE* tls_io_instance)
         wolfSSL_SetIOWriteCtx(tls_io_instance->ssl, tls_io_instance);
         wolfSSL_SetIOReadCtx(tls_io_instance->ssl, tls_io_instance);
 
+#ifdef HAVE_SECURE_RENEGOTIATION
+        if(wolfSSL_UseSecureRenegotiation(tls_io_instance->ssl) != SSL_SUCCESS)
+        {
+            LogError("unable to enable secure renegotiation");
+            return MU_FAILURE;
```
danewalton, line 572: I believe the code style in this repo is to have one return point except parameter checking. This should likely be
```suggestion
            result = MU_FAILURE;
```
and may have to move around to work with the other `if/else if's`

danewalton (approved):
Please update that result to not return in the middle of the function. Other than that lgtm

adapters/tlsio_wolfssl.c
momuno: In this case, if result = MU_FAILURE, result will get reset to 0 at line 577.  So it should be returned here.

adapters/tlsio_wolfssl.c
```diff
@@ -564,6 +565,14 @@ static int create_wolfssl_instance(TLS_IO_INSTANCE* tls_io_instance)
 
         tls_io_instance->tlsio_state = TLSIO_STATE_NOT_OPEN;
         result = 0;
+
+#ifdef HAVE_SECURE_RENEGOTIATION
+        if(wolfSSL_UseSecureRenegotiation(tls_io_instance->ssl) != SSL_SUCCESS)
```
ericwol-msft, line 570:
```suggestion
        if (wolfSSL_UseSecureRenegotiation(tls_io_instance->ssl) != SSL_SUCCESS)
```

adapters/tlsio_wolfssl.c
```diff
@@ -892,6 +901,10 @@ int tlsio_wolfssl_send(CONCRETE_IO_HANDLE tls_io, const void* buffer, size_t siz
             {
                 result = 0;
             }
+
+            // remove on send complete and callback context
+            memset((void*)&tls_io_instance->on_send_complete, 0, sizeof(tls_io_instance->on_send_complete));
```
ericwol-msft, line 906: isn't this just setting the pointer to NULL?

ericwol-msft (approved):

adapters/tlsio_wolfssl.c
momuno: It is, but I wanted to mimic the code that was already in tlsio_mbedtls.c doing its own reset (albeit of a struct containing the function pointer and the callback context).  Maybe it's not worth it.

Issue #497, opened 2021-01-13
[Azure Sphere] retry fails w/ wolfssl


fanzhe98:
Recently we updated all of other system components as well, including wolfssl, libc, curl and etc. this results the retry/reauthentication path after connection loss unusable again.

momuno:
Thank you for creating this.  I've linked the PR.

PR #498, opened 2021-01-13
fix wolfssl retry path

fanzhe98:
Fix wolfssl retry path with wolfssl version 4.6.0
note that ssl had chance to be null, and device id for ctx must also be set
#497


fanzhe98:
@ewertons @momuno for visibility

momuno:
Why is wolfSSL_CTX_SetDevId() now required?

fanzhe98:
Not entirely sure - theoretically CTX SetDevId is not needed per connection. Just get several trials and this will let it work.

momuno:
> Not entirely sure - theoretically CTX SetDevId is not needed per connection. Just get several trials and this will let it work.

I want to check this out a bit more and understand why before I add it in.  Can you please remind me, to reproduce the issue and test this fix, did you run the Azure IoT sample, and then disconnect from the network and then reconnect?

fanzhe98:
> > Not entirely sure - theoretically CTX SetDevId is not needed per connection. Just get several trials and this will let it work.
> 
> I want to check this out a bit more and understand why before I add it in. Can you please remind me, to reproduce the issue and test this fix, did you run the Azure IoT sample, and then disconnect from the network and then reconnect?

Yes, that is what I did.

momuno:
After reviewing the code, I suspect there may have been a bug from before the Azure Sphere updates.
tlsio_wolfssl_retrieveoptions() will add the needed SetDeviceId option ONLY if tls_io_instance->wolfssl_device_id has been set to something other than INVALID_DEVID.  But this never gets set unless the ssl object has not been created, and this scenario cannot happen (the ssl object appears to always be set before the options can be set).

Your code change includes a fix for this, but I do not think the additional wolfSSL_CTX_SetDevId() function calls that matter in this case.

In short, can you please try these changes and see if it still fixes the retry path? Thanks!

https://github.com/Azure/azure-c-shared-utility/compare/momuno/azure-sphere-retry-wolfssl#diff-5bf189cb1123964e6673fed51a851c5d7625626b2258e00f5317c067e769e1a3L979-L999

momuno (approved):

momuno:
Hi @fanzhe98.  One of the unit tests was not passing due to the change and the following fixes that.  Can you please include this in your change?  Thanks.

https://github.com/Azure/azure-c-shared-utility/compare/momuno/azure-sphere-retry-wolfssl#diff-5bf189cb1123964e6673fed51a851c5d7625626b2258e00f5317c067e769e1a3L609-L615

momuno:
Replacing with #500

PR #499, opened 2021-01-16
Fix finding apple libraries

SuperSandro2000:
I wrote this while fixing the library in nixpkgs for darwin and thought upstreaming is a good idea.


SuperSandro2000:
Is the other pipeline supposed to be stuck like this?

danewalton:
@SuperSandro2000 thanks for the contribution! Can you explain what behavior you were seeing and what this fixes? We have builds at the gate for nativexcode and OSX that don't seem to be having issues.

SuperSandro2000:
> Can you explain what behavior you were seeing and what this fixes?

I can't remember but nixpkgs has isolation for all frameworks so they are not just all available all the time and they are available under different paths. Without those changes it didn't build.

danewalton:
Thanks for the contribution @SuperSandro2000. At this time we don't have any issues with our gate checkins or our iOS samples. I'm going to close this for now since we don't currently see anything broken.

PR #504, opened 2021-01-30
Proxy use-case fixed for wolfSSL

momuno:
This fixes the issue reported in https://github.com/Azure/azure-iot-sdk-c/issues/1842

I have tested this with and without a proxy, for all 4 working protocols. MQTT, MQTT/WS, AMQP, AMQP/WS.  HTTP is not working currently and already has an issue filed for it.

### DIFF
--- a/adapters/tlsio_wolfssl.c
+++ b/adapters/tlsio_wolfssl.c
@@ -768,7 +768,7 @@ void tlsio_wolfssl_destroy(CONCRETE_IO_HANDLE tls_io)
 
 int tlsio_wolfssl_open(CONCRETE_IO_HANDLE tls_io, ON_IO_OPEN_COMPLETE on_io_open_complete, void* on_io_open_complete_context, ON_BYTES_RECEIVED on_bytes_received, void* on_bytes_received_context, ON_IO_ERROR on_io_error, void* on_io_error_context)
 {
-    int result = 0;
+    int result;
 
     if (tls_io == NULL)
     {
@@ -808,6 +808,10 @@ int tlsio_wolfssl_open(CONCRETE_IO_HANDLE tls_io, ON_IO_OPEN_COMPLETE on_io_open
                 LogError("Cannot open the underlying IO.");
                 tls_io_instance->tlsio_state = TLSIO_STATE_NOT_OPEN;
                 result = MU_FAILURE;
+            }
+            else
+            {
+                result = 0;
             }
         }
     }
--- a/adapters/tlsio_wolfssl.c
+++ b/adapters/tlsio_wolfssl.c
@@ -768,7 +768,7 @@ void tlsio_wolfssl_destroy(CONCRETE_IO_HANDLE tls_io)
 
 int tlsio_wolfssl_open(CONCRETE_IO_HANDLE tls_io, ON_IO_OPEN_COMPLETE on_io_open_complete, void* on_io_open_complete_context, ON_BYTES_RECEIVED on_bytes_received, void* on_bytes_received_context, ON_IO_ERROR on_io_error, void* on_io_error_context)
 {
-    int result;
+    int result = 0;
 
     if (tls_io == NULL)
     {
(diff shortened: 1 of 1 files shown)


adapters/tlsio_wolfssl.c
```diff
@@ -768,7 +768,7 @@ void tlsio_wolfssl_destroy(CONCRETE_IO_HANDLE tls_io)
 
 int tlsio_wolfssl_open(CONCRETE_IO_HANDLE tls_io, ON_IO_OPEN_COMPLETE on_io_open_complete, void* on_io_open_complete_context, ON_BYTES_RECEIVED on_bytes_received, void* on_bytes_received_context, ON_IO_ERROR on_io_error, void* on_io_error_context)
 {
-    int result;
+    int result = 0;
```
hihigupt, line 771: Lets not give default value and let each branch set this value.

adapters/tlsio_wolfssl.c
```diff
@@ -924,7 +911,10 @@ void tlsio_wolfssl_dowork(CONCRETE_IO_HANDLE tls_io)
         if ((tls_io_instance->tlsio_state != TLSIO_STATE_NOT_OPEN) &&
             (tls_io_instance->tlsio_state != TLSIO_STATE_ERROR))
         {
-            decode_ssl_received_bytes(tls_io_instance);
+            if (tls_io_instance->tlsio_state != TLSIO_STATE_OPENING_UNDERLYING_IO) //Proxy used
```
hihigupt, line 914: In general flow of state is 

TLSIO_STATE_NOT_OPEN -> TLSIO_STATE_OPENING_UNDERLYING_IO -> TLSIO_STATE_IN_HANDSHAKE -> TLSIO_STATE_OPEN

and when in TLSIO_STATE_OPENING_UNDERLYING_IO we do xio_open (line : 798), and when xio_open completes we move state to TLSIO_STATE_IN_HANDSHAKE. 

Curious to know what is the state transition in proxy, that is causing do work to call in TLSIO_STATE_OPENING_UNDERLYING_IO state.

adapters/tlsio_wolfssl.c
momuno: FYI, these files are not consistent in their approach.  mbedtls does start with int result = 0 at this point. I understand the value of not starting with result = 0, but I also think we need to be consistent in our SDK.

adapters/tlsio_wolfssl.c
momuno: When NOT using a proxy, the calls run `tlsio_wolfssl_open -> xio_open -> socketio_open -> (tlsio_wolfssl.c)on_underlying_io_open_complete`.  Here the tlsio state changes to TLSIO_STATE_IN_HANDSHAKE, followed by the callback `on_handshake_done` which changes the state to TLSIO_STATE_OPEN.  This is what you describe above and is what we are used to.

When using a proxy, the calls run `tlsio_wolfssl_open -> xio_open -> http_proxy_io_open -> xio_open -> socketio_open -> (http_proxy_io.c)on_underlying_io_open_complete -> xio_send ->socketio_send -> (http_proxy_io.c)unchecked_on_send_complete`.  The tlsio state remains in TLSIO_STATE_OPENING_UNDERLYING_IO. 

Following the mbedtls implementation, it is during the do_work section (not the open) that the callback `(tlsio_wolfssl.c)on_underlying_io_open_complete` is called.  `tlsio_wolfssl_dowork -> xio_dowork -> http_proxy_io_dowork -> xio_dowork -> socketio_dowork -> (http_proxy_io.c)on_underlying_io_bytes_received -> (tlsio_wolfssl.c)on_underlying_io_open_complete`.  

IF the state is still in TLSIO_STATE_OPENING_UNDERLYING_IO and the client tries to decode received bytes, the connection fails.  Dowork must occur first.

momuno:
It also looks like the Proxy state is `HTTP_PROXY_IO_STATE_WAITING_FOR_CONNECT_RESPONSE` when `tlsio_wolfSSL_dowork` is first called.  When `decode_ssl_received_bytes` is then called, wolfSSL will end up internally calling wolfSSL_connect, but since the Proxy state is not yet `HTTP_PROXY_IO_STATE_OPEN`, the failure occurs on line 850 of http_proxy_io.c

adapters/tlsio_wolfssl.c
momuno: I agree that would be a cleaner solution.  Since it is different than how wolfssl is written, my concern was the possibility of introducing a bug to wolfssl with it, but that would only pertain to the state `TLSIO_STATE_CLOSING`.  Does this state need to run `decode_ssl_received_bytes`?

hihigupt (approved):

PR #512, opened 2021-02-08
tlsio support for http proxy io

yunhaoling:


ericwol-msft:
please add unit tests to https://github.com/Azure/azure-c-shared-utility/blob/master/tests/http_proxy_io_ut/http_proxy_io_ut.c

momuno:
> please add unit tests to https://github.com/Azure/azure-c-shared-utility/blob/master/tests/http_proxy_io_ut/http_proxy_io_ut.c

Also, search for `http_proxy_io_create`.  The unit tests will show an example of how to create proxy.

inc/azure_c_shared_utility/http_proxy_io.h
```diff
@@ -19,6 +19,7 @@ typedef struct HTTP_PROXY_IO_CONFIG_TAG
     int proxy_port;
     const char* username;
     const char* password;
+    bool is_tls_proxy;
```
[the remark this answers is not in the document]
yunhaoling: naming updated

yunhaoling:
hey @ericwol-msft, I tried to add two tests.

however, I don't know how to give a default value to the bool variable (`true/false`) in the struct -- I guess it's not supported in C.
I find that during runtime a random `true/false` could be assigned to the variable, do you know how to resolve the issue in a non-breaking way?

ericwol-msft:
yes, this is an issue. You should use [http_proxy_io_set_option()](https://github.com/Azure/azure-c-shared-utility/blob/95ffad454d63ae5c4316909cdd3b5890091b9714/src/http_proxy_io.c#L913) to extend the http functionality and not change the io_create_parameters. Move use_tls_proxy to HTTP_PROXY_IO_INSTANCE and add a new option name/value that can be passed in to see this value.

@ewertons

src/http_proxy_io.c
```diff
@@ -9,11 +9,14 @@
...
 #include "azure_c_shared_utility/xio.h"
 #include "azure_c_shared_utility/socketio.h"
+#include "azure_c_shared_utility/tlsio.h"
+#include "azure_c_shared_utility/platform.h"
 #include "azure_c_shared_utility/crt_abstractions.h"
 #include "azure_c_shared_utility/http_proxy_io.h"
 #include "azure_c_shared_utility/azure_base64.h"
 
 static const char* const OPTION_UNDERLYING_IO_OPTIONS = "underlying_io_options";
+static const char* const OPTION_USE_TLS_IO = "use_tls_io";
```
ericwol-msft, line 19:
```suggestion
static const char* const OPTION_USE_TLS_PROXY = "use_tls_proxy";
```

src/http_proxy_io.c
```diff
@@ -45,6 +48,7 @@ typedef struct HTTP_PROXY_IO_INSTANCE_TAG
     XIO_HANDLE underlying_io;
     unsigned char* receive_buffer;
     size_t receive_buffer_size;
+    bool use_tls_io;
```
ericwol-msft, line 51: need to also add the copy code to http_proxy_io_clone_option()

src/http_proxy_io.c
```diff
@@ -918,6 +922,30 @@ static int http_proxy_io_set_option(CONCRETE_IO_HANDLE http_proxy_io, const char
...
+            (void)xio_close(http_proxy_io_instance->underlying_io, NULL, NULL);
+            xio_destroy(http_proxy_io_instance->underlying_io);
+            http_proxy_io_instance->use_tls_io = true;
+            const IO_INTERFACE_DESCRIPTION* underlying_io_interface;
+            underlying_io_interface = platform_get_default_tlsio();
+
+            TLSIO_CONFIG tls_io_config;
+            tls_io_config.hostname = http_proxy_io_instance->proxy_hostname;
+            tls_io_config.port = http_proxy_io_instance->proxy_port;
+            http_proxy_io_instance->underlying_io = xio_create(underlying_io_interface, &tls_io_config);
```
ericwol-msft, line 936: is the previous call xio_create in http_proxy_io_create going to fail on proxy with tls mutual auth failure?

src/http_proxy_io.c
```diff
@@ -918,6 +922,30 @@ static int http_proxy_io_set_option(CONCRETE_IO_HANDLE http_proxy_io, const char
...
+
+            if (http_proxy_io_instance->underlying_io == NULL)
+            {
+                http_proxy_io_destroy(http_proxy_io);
```
ericwol-msft, line 940: I don't think we should be destroying the handle. I don't think we own this here.

yunhaoling: I'm checking the socket_io creation logic in `http_proxy_io_create`, I find it would do some check and clean-up work.
do you think we also need to do the check -- tls interface and instance -- and do clean-up work in case of failure?

```c
if (result->underlying_io == NULL)
{
    /* Codes_SRS_HTTP_PROXY_IO_01_012: [ If xio_create fails, http_proxy_io_create shall fail and return NULL. ]*/
    LogError("Unable to create the underlying IO.");
    /* Codes_SRS_HTTP_PROXY_IO_01_008: [ When http_proxy_io_create fails, all allocated resources up to that point shall be freed. ]*/
    free(result->password);
    free(result->username);
    free(result->proxy_hostname);
    free(result->hostname);
    free(result);
    result = NULL;
}
```

src/http_proxy_io.c
yunhaoling: `http_proxy_io_create` won't fail as it's creating the socket_io instance.

it's the `socket_dowork` that will fail, for example on windows we will get error:
> Socketio_Failure: Receiving data from endpoint: 10054.

which is raised from the [socketio_win32 do_work code](https://github.com/Azure/azure-c-shared-utility/blob/fe841e6e673c6ec0c91cb45c634b3af8dab9bdd5/adapters/socketio_win32.c#L690-L695):
```c
int last_error = WSAGetLastError();
if (last_error != WSAEWOULDBLOCK && last_error != ERROR_SUCCESS)
{
    LogError("Socketio_Failure: Receiving data from endpoint: %d.", last_error);
    indicate_error(socket_io_instance);
}
```

src/http_proxy_io.c
yunhaoling: I'm doing it as `if ((strcmp(name, OPTION_UNDERLYING_IO_OPTIONS) == 0) || (strcmp(name, OPTION_USE_TLS_HTTP_PROXY) == 0))`

not sure if this is the correct/preferred way

samples/http_proxy_use_tlsio_connect/main.c
```diff
@@ -0,0 +1,133 @@
...
+        const IO_INTERFACE_DESCRIPTION* proxyio_interface = http_proxy_io_get_interface_description();
+        if (tlsio_interface == NULL)
+        {
+            (void)printf("Error getting tlsio interface description.");
+            result = MU_FAILURE;
+        }
+        else
+        {
+            HTTP_PROXY_IO_CONFIG proxy_io_config = { "www.google.com", 443, PROXY_HOSTNAME, PROXY_PORT, NULL, NULL };
+            TLSIO_CONFIG tlsio_config = { "www.google.com", 443, NULL, NULL};
```
ericwol-msft, line 85: please don't use google endpoints in your examples.

samples/CMakeLists.txt
```diff
@@ -16,4 +16,5 @@ add_sample_directory(iot_c_utility)
 if (NOT ("${ARCHITECTURE}" STREQUAL "ARM"))
     add_sample_directory(socketio_connect)
     add_sample_directory(tlsio_connect)
+    add_sample_directory(http_proxy_use_tlsio_connect)
```
ericwol-msft, line 19:
```suggestion
    add_sample_directory(http_proxy_tlsio_connect)
```

samples/http_proxy_use_tlsio_connect/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+#Copyright (c) Microsoft. All rights reserved.
+#Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+set(http_proxy_use_tlsio_connect_c_files
```
ericwol-msft, line 4:
```suggestion
set(http_proxy_tlsio_connect_c_files
```

samples/http_proxy_use_tlsio_connect/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+    main.c
+)
+
+IF(WIN32)
+    #windows needs this define
+    add_definitions(-D_CRT_SECURE_NO_WARNINGS)
+    add_definitions(-DGB_MEASURE_MEMORY_FOR_THIS -DGB_DEBUG_ALLOC)
+ENDIF(WIN32)
+
+add_executable(http_proxy_use_tlsio_connect ${http_proxy_use_tlsio_connect_c_files})
```
ericwol-msft, line 14:
```suggestion
add_executable(http_proxy_tlsio_connect ${http_proxy_tlsio_connect_c_files})
```

samples/http_proxy_use_tlsio_connect/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+target_link_libraries(http_proxy_use_tlsio_connect
+    aziotsharedutil
+)
+
+if(${use_openssl} AND WIN32)
+	file(COPY ${SSL_DLL} DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+	file(COPY ${CRYPTO_DLL} DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+endif()
+
+set_target_properties(http_proxy_use_tlsio_connect
```
ericwol-msft, line 25:
```suggestion
set_target_properties(http_proxy_tlsio_connect
```

samples/http_proxy_use_tlsio_connect/main.c
```diff
@@ -0,0 +1,133 @@
...
+#include "azure_c_shared_utility/optimize_size.h"
+#include "azure_c_shared_utility/xio.h"
+#include "azure_c_shared_utility/tlsio.h"
+#include "azure_c_shared_utility/http_proxy_io.h"
+#include "azure_c_shared_utility/platform.h"
+
+
+#define PROXY_HOSTNAME "127.0.0.1"
+#define PROXY_PORT 8899
+#define PROXY_SERVER_CERTIFICATE_DATA "<<<server ssl certificate data which is used by the client to authentcate the identity of the proxy server>>>"
```
ericwol-msft, line 15: is this in PEM format?

samples/http_proxy_use_tlsio_connect/main.c
```diff
@@ -0,0 +1,133 @@
...
+static void on_io_open_complete(void* context, IO_OPEN_RESULT open_result)
+{
+    (void)context, (void)open_result;
+    (void)printf("Open complete called\r\n");
+
+    if (open_result == IO_OPEN_OK)
+    {
+        XIO_HANDLE httpproxyio = (XIO_HANDLE)context;
+        const char to_send[] = "GET / HTTP/1.1\r\n"
+            "Host: www.google.com\r\n"
```
ericwol-msft, line 36: don't use google

jessek: Would `www.contoso.com` be ok?

jessek:
@yunhaoling: Looks like there is some feedback and suggestions on the code. Can you please update the PR to reflect the requested changes? We are excited to have this capability!

yunhaoling:
hey @jessek , thanks for checking! I was tied up with some python sdk issues last month, but I should have some free cycles this month and will continue the implantation on this one.

yunhaoling:
closing the PR for now, but we could reactive this in the future if there's new customer wanting it

PR #515, opened 2021-03-18
fix memory leak in win32 socketio

ericwol-msft:


ewertons (approved):
This seems ok to be removed.
It might be left over from before we introduced the dns resolver.

result->addrInfo gets set by 
```c
            (void)memcpy((socket_io_instance->addrInfo), addr, sizeof(*(socket_io_instance->addrInfo)));
```
in 
```c
static int initiate_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
```

Which overwrite the instance that we create in the block you are removing.

PR #516, opened 2021-03-22
fileupload through specific network interface

rajaggrawal:
raising a separate PR for c-utility.

This PR address the following feature requirement
https://github.com/Azure/azure-iot-sdk-c/pull/1884

### DIFF
--- a/adapters/httpapi_curl.c
+++ b/adapters/httpapi_curl.c
@@ -904,6 +904,26 @@ HTTPAPI_RESULT HTTPAPI_SetOption(HTTP_HANDLE handle, const char* optionName, con
                 }
             }
         }
+        else if (strcmp(OPTION_NETWORK_INTERFACE_UPLOAD_TO_BLOB, optionName) == 0)
+        {
+            const char *interfaceName = (const char*)value;
+            if (interfaceName != NULL && interfaceName[0] != '\0')
+            {
+                if (curl_easy_setopt(httpHandleData->curl, CURLOPT_INTERFACE, interfaceName) != CURLE_OK)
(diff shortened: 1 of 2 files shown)


adapters/httpapi_curl.c
```diff
@@ -904,6 +904,26 @@ HTTPAPI_RESULT HTTPAPI_SetOption(HTTP_HANDLE handle, const char* optionName, con
...
+                    result = HTTPAPI_ERROR;
+                }
+                else
+                {
+                    result = HTTPAPI_OK;
+                }
+            }
+            else
+            {
+                result = HTTPAPI_OK;
```
jspaith, line 924: We should return `HTTPAPI_ERROR` here along with a LogError().  

For some more random context here -- it's true of the code above around trusted certs was returning OK if values were not NULL I know, but it's tricky because in one of cases the x509certificate is set but then the x509privateKey is checked against NULL and return OK if it's NULL.  This does this because both values need to be set.  (We really should've made a single option for cert/private key in a struct{} so one OPTION set this, instead of the tearing and needing multiple SetOptions.  But that's a digression).

```c
            httpHandleData->x509certificate = value;
            if (httpHandleData->x509privatekey != NULL)
```

jspaith (changes requested):
Only one small change from me requested here and then after that this repo's change looks good.

ewertons (approved):

adapters/httpapi_curl.c
rajaggrawal: done

jspaith (approved):

PR #518, opened 2021-03-24
Log the results of pthread Lock and Unlock in case of error

ewertons:


momuno (approved):

ericwol-msft (approved):

ericwol-msft:
if you do the submodule dance, can you work with john on #516 getting it in so that it's only done once.

PR #522, opened 2021-04-07
OpenSSL IP Address SAN validation support

CIPop:
- Adding IP Address SAN list support.
- Properly initializing WolfSSL.
- Adding WolfSSL debug logging set-option (requires WolfSSL `--enable-debug` configured binary)

### DIFF
--- a/adapters/platform_linux.c
+++ b/adapters/platform_linux.c
@@ -42,6 +42,11 @@ int platform_init(void)
     {
         result = tlsio_openssl_init();
     }
+#elif USE_WOLFSSL
+    if (result == 0)
+    {
+        result = tlsio_wolfssl_init();
+    }
 #endif
     return result;
 }
@@ -91,5 +96,7 @@ void platform_deinit(void)
 #endif /* DONT_USE_UPLOADTOBLOB */
 #ifdef USE_OPENSSL
     tlsio_openssl_deinit();
+#elif USE_WOLFSSL
+    tlsio_wolfssl_deinit();
 #endif
 }
--- a/adapters/tlsio_openssl.c
+++ b/adapters/tlsio_openssl.c
@@ -1022,8 +1022,8 @@ static int enable_domain_check(TLS_IO_INSTANCE* tlsInstance)
         X509_VERIFY_PARAM *param = SSL_get0_param(tlsInstance->ssl);
 
         X509_VERIFY_PARAM_set_hostflags(param, 0);
-        if ((!X509_VERIFY_PARAM_set1_ip_asc(param, tlsInstance->hostname) ||
-            (!X509_VERIFY_PARAM_set1_host(param, tlsInstance->hostname, strlen(tlsInstance->hostname)))))
+        if (!(X509_VERIFY_PARAM_set1_ip_asc(param, tlsInstance->hostname) ||
+              X509_VERIFY_PARAM_set1_host(param, tlsInstance->hostname, strlen(tlsInstance->hostname))))
         {
             result = MU_FAILURE;
         }
--- a/adapters/tlsio_openssl.c
+++ b/adapters/tlsio_openssl.c
@@ -1022,8 +1022,8 @@ static int enable_domain_check(TLS_IO_INSTANCE* tlsInstance)
         X509_VERIFY_PARAM *param = SSL_get0_param(tlsInstance->ssl);
 
         X509_VERIFY_PARAM_set_hostflags(param, 0);
-        if (!(X509_VERIFY_PARAM_set1_ip_asc(param, tlsInstance->hostname) ||
-              X509_VERIFY_PARAM_set1_host(param, tlsInstance->hostname, strlen(tlsInstance->hostname))))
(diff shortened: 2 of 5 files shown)


adapters/tlsio_wolfssl.c
```diff
@@ -999,6 +1010,12 @@ int tlsio_wolfssl_setoption(CONCRETE_IO_HANDLE tls_io, const char* optionName, c
             tls_io_instance->ignore_host_name_check = *server_name_check;
             result = 0;
         }
+        else if (strcmp("debugLog", optionName) == 0)
+        {
+            (void)wolfSSL_Debugging_ON();
+            (void)wolfSSL_SetLoggingCb(&logging_callback);
```
danewalton, line 1016: not sure what the return values are here but do we care at all about them?

adapters/tlsio_wolfssl.c
```diff
@@ -999,6 +1010,12 @@ int tlsio_wolfssl_setoption(CONCRETE_IO_HANDLE tls_io, const char* optionName, c
             tls_io_instance->ignore_host_name_check = *server_name_check;
             result = 0;
         }
+        else if (strcmp("debugLog", optionName) == 0)
+        {
```
danewalton, line 1014: is this not checking the `value`. What if it's `false`?

CIPop: Good point: we need to check non-zero. Will add.

adapters/tlsio_wolfssl.c
CIPop: Initially I thought we don't but in the light of this failing it seems like we might.
I also need to add a `LIBWOLFSSL_VERSION_HEX` check and conditionally compile the log routing only if Wolf supports it.

adapters/tlsio_wolfssl.c
```diff
@@ -999,6 +1010,12 @@ int tlsio_wolfssl_setoption(CONCRETE_IO_HANDLE tls_io, const char* optionName, c
...
+            (void)wolfSSL_Debugging_ON();
+            (void)wolfSSL_SetLoggingCb(&logging_callback);
+            result = 0;
```
hihigupt: how about doing this in init and wrap around macro DEBUG_WOLFSSL ?

marco comes from wolfssl/options.h

adapters/platform_linux.c
```diff
@@ -91,5 +96,7 @@ void platform_deinit(void)
 #endif /* DONT_USE_UPLOADTOBLOB */
 #ifdef USE_OPENSSL
     tlsio_openssl_deinit();
+#elif USE_WOLFSSL
+    tlsio_wolfssl_deinit();
 #endif
```
hihigupt, line 101: >endif [](start = 1, length = 5)

windows platform also requires this?
[https://github.com/Azure/azure-c-shared-utility/blob/51555845f446a7f0be4e3442a321c558af04910b/adapters/platform_win32.c#L235](https://github.com/Azure/azure-c-shared-utility/blob/51555845f446a7f0be4e3442a321c558af04910b/adapters/platform_win32.c#L235)

adapters/tlsio_wolfssl.c
CIPop: I thought about having it enabled when WolfSSL is built with the debug flag but it's very verbose. 

This way we are backward compatible even in cases where WolfSSL was built with debug but we can enable it using the option if we need to investigate more.

adapters/tlsio_openssl.c
```diff
@@ -1022,7 +1022,8 @@ static int enable_domain_check(TLS_IO_INSTANCE* tlsInstance)
         X509_VERIFY_PARAM *param = SSL_get0_param(tlsInstance->ssl);
 
         X509_VERIFY_PARAM_set_hostflags(param, 0);
-        if (!X509_VERIFY_PARAM_set1_host(param, tlsInstance->hostname, strlen(tlsInstance->hostname)))
+        if (!(X509_VERIFY_PARAM_set1_ip_asc(param, tlsInstance->hostname) ||
```
hihigupt, line 1025: >X509_VERIFY_PARAM_set1_ip_asc [](start = 14, length = 29)

can we do host first and then ip. Because most people will use hostname ?

adapters/platform_linux.c
CIPop: On Windows only Schannel is recommended and officially supported as is Secure Transport on OSX. 
I won't add these to Windows.

adapters/tlsio_openssl.c
CIPop: `set1_host` will always work, even when an ASCII IP Address is used within the `hostname` variable. 
We don't want that because OpenSSL, while not allowed by RFC, will check that ASCII-encoded IPv4/v6 address within a DNS Name-typed SAN entry, something we don't want.

The code will only check IP Address-typed SAN entries when IPs are entered. If we fail to parse the IP (first function will fail) then we fallback to consider the hostname a non-IP entity and the process described here: https://wiki.openssl.org/index.php/Hostname_validation

danewalton (approved):

PR #524, opened 2021-04-20
Added framework for running longhaul tests under valgrind

ericwol-msft:


adapters/tlsio_openssl.c
```diff
@@ -592,8 +592,8 @@ static int openssl_static_locks_install(void)
 
     if (openssl_locks != NULL)
     {
-        LogError("Locks already initialized");
-        result = MU_FAILURE;
+        LogInfo("Locks already initialized");
+        result = 0;
```
ewertons, line 596: This is concerning. It is a significant change from the original behavior. An we just ignore these errors on valgrind when it runs the long haul tests?

ericwol-msft: reverting

PR #532, opened 2021-05-10
Fix uuid memcpy length

ericwol-msft:


ewertons (approved):
Goodness. That's bad and it has been there forever.

PR #542, opened 2021-08-17
Fix for vcpkg build

momuno:


CMakeLists.txt
```diff
@@ -683,15 +683,21 @@ else(${build_as_dynamic})
     set(targets aziotsharedutil)
 endif(${build_as_dynamic})
 
+set(cmake_install_includes_dir ${CMAKE_INSTALL_INCLUDEDIR})
```
ewertons, line 686: I suggest making this var name uppercase. It seems to match the case of other "install" vars. 
But it's just a suggestion.

ewertons (approved):

CMakeLists.txt
momuno: I was thinking lowercase so it does not get confused with official CMAKE variables.

PR #545, opened 2021-08-24
crt_abstractions.c: fix error management in mallocAndStrcpy

ODugas:
It can be demonstrated that the errno_t returned by strcpy_s
(which is described as a regular int by the signature) can
and will contain a positive error code. We therefore need to
check for non-zero returned value instead of strictly
negative integers.


danewalton:
From the docs here: https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2013/td1esda9(v=vs.120)

> Zero if successful; otherwise, an error.

Do you mind updating and rebasing your changes?

ericwol-msft (approved):

ODugas:
@ericwol-msft  @danewalton  I updated and rebased my changes. I'll stay nearby, but I believe this is all I need to do for you to merge this, right?
Thanks a lot.

danewalton:
@ODugas sorry about this are you able to do it once more (rebase master)? It says it's still out of date.

ericwol-msft (approved):

PR #546, opened 2021-08-25
fix memory leak mbedtls + curl

maallahatem:
This pull request fixes a potential memory leak when using curl with mbedlts

````
==32503== HEAP SUMMARY:
==32503==     in use at exit: 1,014 bytes in 3 blocks
==32503==   total heap usage: 33,097 allocs, 33,094 frees, 3,034,693 bytes allocated
==32503== 
==32503== 24 bytes in 1 blocks are definitely lost in loss record 1 of 3
==32503==    at 0x483DD99: calloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==32503==    by 0x4912B4B: ??? (in /usr/lib/x86_64-linux-gnu/libmbedtls.so.2.16.3)
==32503==    by 0x1BDB58: ssl_ctx_callback (httpapi_curl.c:303)
==32503==    by 0x48D5B89: ??? (in /usr/lib/x86_64-linux-gnu/libcurl.so.4.6.0)
==32503==    by 0x48D713E: ??? (in /usr/lib/x86_64-linux-gnu/libcurl.so.4.6.0)
==32503==    by 0x48D7FAE: ??? (in /usr/lib/x86_64-linux-gnu/libcurl.so.4.6.0)
==32503==    by 0x4883295: ??? (in /usr/lib/x86_64-linux-gnu/libcurl.so.4.6.0)
==32503==    by 0x4884D12: ??? (in /usr/lib/x86_64-linux-gnu/libcurl.so.4.6.0)
==32503==    by 0x48A58EC: ??? (in /usr/lib/x86_64-linux-gnu/libcurl.so.4.6.0)
==32503==    by 0x48A6980: curl_multi_perform (in /usr/lib/x86_64-linux-gnu/libcurl.so.4.6.0)
==32503==    by 0x489CDFA: curl_easy_perform (in /usr/lib/x86_64-linux-gnu/libcurl.so.4.6.0)
==32503==    by 0x1C085D: HTTPAPI_ExecuteRequest (httpapi_curl.c:645)
==32503== 
==32503== LEAK SUMMARY:
==32503==    definitely lost: 24 bytes in 1 blocks
==32503==    indirectly lost: 0 bytes in 0 blocks
==32503==      possibly lost: 0 bytes in 0 blocks
==32503==    still reachable: 990 bytes in 2 blocks
==32503==         suppressed: 0 bytes in 0 blocks
````

### DIFF
--- a/adapters/httpapi_curl.c
+++ b/adapters/httpapi_curl.c
@@ -297,6 +297,10 @@ static CURLcode ssl_ctx_callback(CURL *curl, void *ssl_ctx, void *userptr)
             result = CURLE_SSL_CERTPROBLEM;
         }
 #elif USE_MBEDTLS
+        
+        // save ssl context
+        httpHandleData->config = *((mbedtls_ssl_config*)ssl_ctx);
+         
         // set device cert and key
         if (
             (httpHandleData->x509certificate != NULL) && (httpHandleData->x509privatekey != NULL) &&
@@ -333,10 +337,6 @@ static CURLcode ssl_ctx_callback(CURL *curl, void *ssl_ctx, void *userptr)
 #endif
         else
         {
-
-#if USE_MBEDTLS
-            httpHandleData->config = *((mbedtls_ssl_config*)ssl_ctx);
-#endif
             result = CURLE_OK;
         }
     }
--- a/adapters/httpapi_curl.c
+++ b/adapters/httpapi_curl.c
@@ -333,7 +333,10 @@ static CURLcode ssl_ctx_callback(CURL *curl, void *ssl_ctx, void *userptr)
 #endif
         else
         {
-        	httpHandleData->config = *((mbedtls_ssl_config*)ssl_ctx);
+
+#if USE_MBEDTLS
+            httpHandleData->config = *((mbedtls_ssl_config*)ssl_ctx);
+#endif
             result = CURLE_OK;
         }
     }
--- a/adapters/httpapi_curl.c
+++ b/adapters/httpapi_curl.c
@@ -48,6 +48,7 @@ typedef struct HTTP_HANDLE_DATA_TAG
     mbedtls_x509_crt cert;
     mbedtls_pk_context key;
     mbedtls_x509_crt trusted_certificates;
+    mbedtls_ssl_config config;
 #endif
 } HTTP_HANDLE_DATA;
 
@@ -148,6 +149,7 @@ HTTP_HANDLE HTTPAPI_CreateConnection(const char* hostName)
                         mbedtls_x509_crt_init(&httpHandleData->cert);
                         mbedtls_pk_init(&httpHandleData->key);
                         mbedtls_x509_crt_init(&httpHandleData->trusted_certificates);
+                        mbedtls_ssl_config_init(&httpHandleData->config);
 #endif
                     }
                 }
@@ -175,6 +177,7 @@ void HTTPAPI_CloseConnection(HTTP_HANDLE handle)
         mbedtls_x509_crt_free(&httpHandleData->cert);
         mbedtls_pk_free(&httpHandleData->key);
         mbedtls_x509_crt_free(&httpHandleData->trusted_certificates);
+        mbedtls_ssl_config_free( &httpHandleData->config);
 #endif
         free(httpHandleData);
     }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by momuno — CHANGES_REQUESTED** [review]
This is a good find.  We are not currently storing the mbedtls_ssl_config which contains pointers to potential dynamic memory allocation.  Your change proposes to store this and later free it.

My only question pertains to the code location of the mbedtls_ssl_config assignment.  Please recheck.

File: adapters/httpapi_curl.c [hunk 1]
```diff
@@ -330,6 +333,7 @@ static CURLcode ssl_ctx_callback(CURL *curl, void *ssl_ctx, void *userptr)
 #endif
         else
         {
+        	httpHandleData->config = *((mbedtls_ssl_config*)ssl_ctx);
```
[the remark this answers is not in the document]
maallahatem: Hi @momuno , yes I totally agree with you, the mbedtls_ssl_config shall be managed only under USE_MBEDTLS flags
I did the modification in another commit

momuno:
@maallahatem There are a few scenarios where the final `else` will never be reached and the `ssl_ctx` will not be stored as intended.  I propose it be stored first thing when entering the `USE_MBEDTLS` block, since we don't want to leak that memory regardless of what else happens.

momuno:
Hi @maallahatem.  Have you been able to review my proposal for the change?  Thanks.

maallahatem:
Hi @momuno , sorry to better late in my answer
Yes, you can go ahead with your proposal :)

CIPop (approved):

adapters/httpapi_curl.c
```diff
@@ -294,6 +297,10 @@ static CURLcode ssl_ctx_callback(CURL *curl, void *ssl_ctx, void *userptr)
             result = CURLE_SSL_CERTPROBLEM;
         }
 #elif USE_MBEDTLS
+        
+        // save ssl context
+        httpHandleData->config = *((mbedtls_ssl_config*)ssl_ctx);
```
CIPop, line 302: I don't have much experience with cURL. Are all fields of the config initialized by cURL? 
mbedTLS API documentation states:

> You need to call mbedtls_ssl_config_defaults() unless you manually set all of the relevent fields yourself.

momuno:
> Hi @momuno , sorry to better late in my answer Yes, you can go ahead with your proposal :)

Thanks @maallahatem, appreciate your work on this.

adapters/httpapi_curl.c
momuno: Doing some digging, so far it is unclear if cURL calls mbedtls_ssl_config_defaults().  However it does look like this code manually sets the relevant fields(e.g., mbedtls_ssl_conf_own_cert(), mbedtls_ssl_conf_ca_chain()).  To be safe, we could call it ourselves with the same values does in tlsio_mbedtls.c.  

The memory for mbedtls_ssl_config (aka ssl_ctx) is allocated by cURL prior to sending in this callback. It looks like there is a 'curl_easy_cleanup' that is supposde to clean up memory cURL allocates, but it could be safer for us to still deallocate ourselves.

On second look, it doesn't make sense to dereference and store the ssl_ctx before these mbedtls_ssl_config (aka ssl_ctx) fields are set. But it also doesn't work to store the ssl_ctx in an 'else' clause that may not be called.  I am going to re-work the logic to reach a solution.

momuno:
Closing to create a separate PR.

momuno:
Hi @maallahatem.  I did further investigation and was able to reproduce the memory leak.  However, I suspect the issue was a mismatch of TLS platform configurations between cURL and the C SDK.  When using cURL with OpenSSL (this can occur by directly using the curl installed via apt-get) and the C SDK with mbedTLS, the memory leak occurred as you describe above.  When I downloaded and configured cURL from source to use mbedTLS, there was no leak.

I have updated the httpapi_curl.c file to check for valid and matching TLS platforms between cURL and the CSDK.
https://github.com/Azure/azure-c-shared-utility/pull/595

Thank you for your contribution and bringing this issue to our attention!

hatem-maalla:
Hi @momuno , Yes I see now the cause of this problem.
I think the fix will improve the robustness of the code against unexpected client side behavior.
Thanks a lot for your help.

Issue #547, opened 2021-08-30
[Azure Sphere] tlsio_openssl.c does not support setting signature algorithms as an option


fanzhe98:

fanzhe98:
cc @thalesx @suhuruli

CIPop:
We __strongly recommend__ using OpenSSL (or any other TLS stack) with secure defaults. 
This way, the OS is able to update the TLS stack and potentially change the secure defaults when algorithms are compromised.

That said, we do already have support for this feature: https://github.com/Azure/azure-iot-sdk-c/blob/master/iothub_client/devdoc/iothubclient_c_library.md 

Note that using this option runs the risk of either exposing the device to future security issues (if the selected algorithms are compromised) or unable to connect (if the OS updates the TLS stack and removes the selected algorithms).


```
"CipherSuite" - only available when OpenSSL is used. value is a pointer to a null terminated string that contains a list in the format specified by SSL_set_cipher_list. Example:
// IoT Hub using ECC server certificate chain:
IoTHubDeviceClient_LL_SetOption(device_ll_handle, OPTION_OPENSSL_CIPHER_SUITE, "ECDH+ECDSA+HIGH");
```

fanzhe98:
How is it that the TLS ciphersuite would determine what hash function must be used with ECDSA signatures? TLS ciphersuites list the function for KDF, independent of the key pair and signature type of the TLS client certificate.

fanzhe98:
1. the signature in the context is the digital signature in tls mutual authentication
2. openssl defaults is *NOT* with the strongest security it could have but the most accepted algorithm w/ the server at the time (and this is even not true, as their defaults is not changing within the version), and it can always be overwritten via the config file

PR #548, opened 2021-08-30
[Azure Sphere] add openssl options for setting client signature algorithms

fanzhe98:
#547


CIPop:
As discussed offline, we don't recommend changing OpenSSL defaults and this change is not meeting our feature bar at this time.
We support full TLS stack customization within our new [Embedded C SDK](https://github.com/Azure/azure-sdk-for-c).

PR #555, opened 2021-09-04
Vcpkg CURL fix for build

momuno:
This change updates how CURL is found for vcpkg.  It now checks CMake's find_package() before trying pkg-config.  This change is needed for vcpkg build.
As a result, the Config.cmake file must forward its dependency on CURL downstream.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -503,8 +503,7 @@ if(${use_http})
         if (NOT use_builtin_httpapi)
 
             # try CMake find_package first
-            set(CURL_DIR CMake)
-            find_package(CURL CONFIG)
+            find_package(CURL)
 
             # As mentioned at https://cmake.org/Wiki/CMake_Cross_Compiling the
             # pkg-config tool can not be used by cmake while cross compiling.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -501,21 +501,20 @@ if(${use_http})
         set(aziotsharedutil_target_libs ${aziotsharedutil_target_libs} crypt32 winhttp)
     else()
         if (NOT use_builtin_httpapi)
-            if (CMAKE_CROSSCOMPILING)
-                # As mentioned at https://cmake.org/Wiki/CMake_Cross_Compiling the
-                # pkg-config tool can not be used by cmake while cross compiling.
-                message(STATUS "Cross compiling not using pkg-config")
-            else()
-                # try pkg-config first
-                find_package(PkgConfig)
-                if(PKG_CONFIG_FOUND)
-                    pkg_check_modules(CURL libcurl)
-                endif()
-            endif()
 
-            # if that didn't work, try CMake find_package
-            if(NOT CURL_FOUND)
-                find_package(CURL)
+            # try CMake find_package first
+            find_package(CURL)
+
+            # As mentioned at https://cmake.org/Wiki/CMake_Cross_Compiling the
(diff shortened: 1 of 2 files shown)


configs/azure_c_shared_utilityConfig.cmake
```diff
@@ -3,10 +3,13 @@
 
 include("${CMAKE_CURRENT_LIST_DIR}/azure_c_shared_utilityTargets.cmake")
 
+include(CMakeFindDependencyMacro)
+set(CURL_DIR CMake)
+find_dependency(CURL CONFIG)
+
```
dg0yt, line 9: IMO you don't need changes to azure_c_shared_utilityConfig.cmake as long as you are linking to `${CURL_LIBRARIES}`. vcpkg needs to fix the variable content.

If you would link to `CURL::libcurl`, you would only need:
~~~cmake
include(CMakeFindDependencyMacro)
find_dependency(CURL)
~~~

CMakeLists.txt
```diff
@@ -501,21 +501,21 @@ if(${use_http})
...
-                    pkg_check_modules(CURL libcurl)
-                endif()
-            endif()
 
-            # if that didn't work, try CMake find_package
-            if(NOT CURL_FOUND)
-                find_package(CURL)
+            # try CMake find_package first
+            set(CURL_DIR CMake)
+            find_package(CURL CONFIG)
```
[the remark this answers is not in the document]
momuno: I am not sure what is best here.  
I notice your comment on the vcpkg PR "You want consistent results, so please use CONFIG consistently with find_package and find_dependency. The keyword excludes Find modules."  I know find_package(CURL) will try module first and then config.  It was working this way.

momuno:
I think this most recent change will work.  I am isolating the change needed for x64-linux to compile to UNIX.

CMakeLists.txt
```diff
@@ -501,21 +501,20 @@ if(${use_http})
...
-                if(PKG_CONFIG_FOUND)
-                    pkg_check_modules(CURL libcurl)
-                endif()
-            endif()
 
-            # if that didn't work, try CMake find_package
-            if(NOT CURL_FOUND)
-                find_package(CURL)
+            # try CMake find_package first
+            find_package(CURL)
```
ewertons, line 506: I hope this will not result in failure for some platform that 1) is not WIN32, AND 2) is not using buit-in http API. In such case we are requiring CURL to be present. But anyway, that was the behavior before.

ewertons (approved):

CMakeLists.txt
momuno: This was in case it was not found using pkg-config, to try another way, and only for cross-compilation.  the logic is still there for cross-compilation (just reversed cmake to be used first, then pkg-config)

configs/azure_c_shared_utilityConfig.cmake
```diff
@@ -9,4 +14,3 @@ set(AZURE_C_SHARED_UTILITY_INCLUDES ${AZURE_C_SHARED_UTILITY_INCLUDES} CACHE INT
 
 include("${CMAKE_CURRENT_LIST_DIR}/azure_c_shared_utilityFunctions.cmake")
 include("${CMAKE_CURRENT_LIST_DIR}/azure_iot_build_rules.cmake")
```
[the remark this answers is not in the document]
momuno: strange. when I open this in vs code it shows a blank line 17. vim only shows 16 lines though. weird.

PR #558, opened 2021-09-10
Use PAL time functions in the console logger

### LINKED ISSUE #556: consolelogger function for GNUC calles time instead of the adapter get_time for specific platform.
In `consolelogger.c` [Line 207](https://github.com/Azure/azure-c-shared-utility/blob/64ea7a9359d9f431a713658cb95bbd0c426c8e3c/src/consolelogger.c#L207) the function `consolelogger_log` that is used to Log to console (or any other endpoint according to platform) uses the function `time`.

```c
void consolelogger_log(LOG_CATEGORY log_category, const char* file, const char* func, int line, unsigned int options, const char* format, ...)
{
    time_t t;
    va_list args;
    va_start(args, format);

    t = time(NULL);
    
    // ...
```

The `time` function is not available in all scenarios, even though [there is already an API set up](https://github.com/Azure/azure-c-shared-utility/blob/64ea7a9359d9f431a713658cb95bbd0c426c8e3c/inc/azure_c_shared_utility/agenttime.h#L28) for such case in `agenttime.h` which will provide the `get_time` function on the ported module `agenttime_<custom>.c`, adapted for each implementation accordingly.

```c
/** @brief Get current calendar time.
*
*    @details This function provides the same functionality as the
*    standard C @c time() function.
*/
MOCKABLE_FUNCTION(, time_t, get_time, time_t*, currentTime);
```

ericwol-msft:
fixes #556


ericwol-msft:
#556

danewalton (approved):

ewertons (approved):

Issue #525, opened 2021-04-26
get_time_ns on iOS is 'slow'


gutley:
Hi,

get_time_ns is too slow on iOS nowadays (and probably macOS) and can account for a substantial amount of CPU usage in the worker thread just to check it, especially when IOTHUB_CLIENT_CORE_INSTANCE.do_work_freq_ms is its default value of 1

Since macOS 10.12 and iOS 10, clock_gettime is available for use with CLOCK_MONOTONIC and so get_time_ns should be updated to reflect this.

I don't have enough other platforms to do a fully testable Pull Request. In my local copy I just add this at the start of get_time_ns:

```C
int get_time_ns(struct timespec* ts)
{
    int err;

    if (__builtin_available(iOS 10.0, *))
    {
        err = clock_gettime(CLOCK_MONOTONIC, ts);
        return err;
    }

#ifdef __MACH__
....
```

Doing so makes CPU usage of get_time_ns virtually negligible

gutley:
To further demonstrate the issue, here is a 20 seconds sample of my app sitting doing nothing with a default IOTHUB_CLIENT_CORE_INSTANCE.do_work_freq_ms and the current implementation of get_time_ns.

<img width="715" alt="image" src="https://user-images.githubusercontent.com/2040690/116072433-fc151280-a686-11eb-99ca-a8e649ed9bd8.png">

As you can see, get_time_ns accounts for about 2.5 seconds of the 3.5 seconds _LL_DoWork spends just spinning its heals.

danewalton:
This might be asking a lot but more for my education and I'm curious....can you do the profiling for the implementation you're asking for? For comparison?

danewalton:
Ahh great to know yea that's significant. I'm looking into the best way to get the quicker version in but only have it compile on the systems that support it.
Examples which have this kind of change which either had issues or worked.

Redis: https://github.com/redis/redis/pull/7644/files#diff-4bbdf3e8cafcbb3340a2eaecf977a215b5f4886638cd6b031a570ca219fa571bR143-R170

Reverted on nodeJS: https://github.com/nodejs/node/issues/11104

I'll keep digging thanks for the info.

danewalton:
Hey @gutley are you able to try the PR started here: https://github.com/Azure/azure-c-shared-utility/pull/559

monotomic time looks to be failing in the implementation's ut but realtime looks to work.

gutley:
[...]

PR #559, opened 2021-09-11
Change Apple time for performance on newer versions

danewalton:
This incorporates the `clock_gettime()` function which is implemented in iOS 10.0+ and MacOS 10.12+. `CLOCK_REALTIME` is selected here instead of `CLOCK_MONOTONIC` since monotonic fails UT.

Original issue here: #525


danewalton:
@jspaith Sorry about that I should probably have closed this. Latest update in the issue is I didn't see the same performance problem the customer was originally seeing. So I'll hold off on the change.

PR #564, opened 2021-10-29
fix crash wiht invalid proxy configuration

maallahatem:
This patch fixes a crash when trying to connect iothub with invalid proxy configuration


ewertons:
That's pretty bad that this bug was out there. Thanks for the catch and even more for the proposed fix! I'm just checking on the unit tests for that, but should be good to go soon.

PR #567, opened 2021-10-30
Fix crash with invalid proxy configuration

ewertons:
This replaces the original PR https://github.com/Azure/azure-c-shared-utility/pull/564
Unit tests were added to this new PR.

### DIFF
--- a/build/.vsts-ci.yml
+++ b/build/.vsts-ci.yml
@@ -5,7 +5,7 @@ resources:
 jobs:
 - job: checksubmodule
   pool:
-    name: Hosted Ubuntu 1604
+    vmImage: 'ubuntu-20.04'
   steps:
   - script: |
       sudo apt-get update && apt-get install -y \
--- a/tests/http_proxy_io_ut/http_proxy_io_ut.c
+++ b/tests/http_proxy_io_ut/http_proxy_io_ut.c
@@ -197,15 +197,15 @@ extern "C"
 {
 #endif
 
-pfCloneOption tlsio_cyclonessl_clone_option;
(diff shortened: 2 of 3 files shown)


tests/http_proxy_io_ut/http_proxy_io_ut.c
```diff
@@ -1575,6 +1575,70 @@ TEST_FUNCTION(http_proxy_io_retrieve_options_calls_the_underlying_retrieve_optio
     http_proxy_io_get_interface_description()->concrete_io_destroy(http_io);
 }
 
+#define OPTION_UNDERLYING_IO_OPTIONS "underlying_io_options"
+
+/* GitHub PR 564 */
```
danewalton, line 1580: these labels necessary?

build/.vsts-ci.yml
```diff
@@ -5,7 +5,7 @@ resources:
 jobs:
 - job: checksubmodule
   pool:
-    name: Hosted Ubuntu 1604
+    vmImage: 'ubuntu-20.04'
```
danewalton, line 8: this is moving on from self hosted right?

danewalton (approved):

tests/http_proxy_io_ut/http_proxy_io_ut.c
ewertons: Oh, the label yes, but the define should be moved to the top. Thanks.

build/.vsts-ci.yml
ewertons: Yes, correct, we are using Azure DevOps agents for that.

PR #569, opened 2021-11-15
remove invalid xio_close

jlaliMohamed:
remove invalid xio_close when using http_compact, mbedtls and proxy

### DIFF
--- a/adapters/tlsio_mbedtls.c
+++ b/adapters/tlsio_mbedtls.c
@@ -581,8 +581,6 @@ void tlsio_mbedtls_destroy(CONCRETE_IO_HANDLE tls_io)
 
         mbedtls_uninit(tls_io_instance);
 
-        xio_close(tls_io_instance->socket_io, NULL, NULL);
-
         if (tls_io_instance->socket_io_read_bytes != NULL)
         {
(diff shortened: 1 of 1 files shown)


danewalton:
Hi @jlaliMohamed can you give more details on why you think this is invalid?

jlaliMohamed:
Hi @danewalton, The problem is produced when using http_compact, tlsio mbedtls and proxy. 
After sending a successful http request  (using mebdtls as a TLS layer and proxy) we receive a log like this:

````
[... httpapi_compact.c:774 ] Waiting for TLS connection
[... httpapi_compact.c:774 ] Waiting for TLS connection
[... httpapi_compact.c:774 ] Waiting for TLS connection
[... http_proxy_io.c:785 ] Invalid tlsio_state. Expected state is HTTP_PROXY_IO_STATE_OPEN.
````

when closing the io socket, the tls (for mbdtls) does the closing, then in the destroy tlsio_mbedtls_destroy function it calls the xio_close again.
To resume is a double call of xio_close ()

danewalton:
I have a repro looking into it to double check this is the right solution.

danewalton:
fixed ut in #572

PR #573, opened 2021-11-19
Add clang builds for embedded TLS adapters

jspaith:
Our TLS stacks that are primarily used on embedded or low-end Linux devices - wolfssl, bear, and mbed - are currently compiled and tested against gcc only.  This is generally fine as gcc is the most common embedded compiler.

However... gcc misses certain warnings condition, especially around uninitialized variables.  (See StackOverflow discussion [here](https://stackoverflow.com/questions/17705880/gcc-failing-to-warn-of-uninitialized-variable) - I was easily able to reproduce this on gcc even with -Wall and any optimization setting I tried.)

clang does not miss this class of errors.  So add a clang build to these stacks.  This build only checks compilation step and does not rerun tests gcc currently handles.


ericwol-msft (approved):

build/.vsts-ci.yml
```diff
@@ -253,6 +253,10 @@ jobs:
       name: 'sdk-c--ubuntu-18'
   displayName: 'linux_mbed'
   steps:
+  - script: |
```
danewalton, line 256: go up to the `_PREVIEW_VSTS_DOCKER_IMAGE` variable and make it `aziotbld/linux-c-ubuntu-mbed`

build/.vsts-ci.yml
```diff
@@ -287,6 +291,10 @@ jobs:
       name: 'sdk-c--ubuntu-18'
   displayName: 'wolfssl'
   steps:
+  - script: |
```
danewalton, line 294: this one `aziotbld/linux-c-ubuntu-wolfssl`

build/.vsts-ci.yml
```diff
@@ -327,6 +335,10 @@ jobs:
       name: 'sdk-c--ubuntu-18'
   displayName: 'bearssl'
   steps:
+  - script: |
```
danewalton, line 338: and this one `aziotbld/linux-c-ubuntu-bearssl`

danewalton (approved):

ericwol-msft (approved):

PR #577, opened 2021-11-24
Relax the range checks on the OpenSSL version

schopin-pro:
OpenSSL 3.0 has a high degree of API compatibility with the 1.1.1
branch, so the 1.1 code branches are also valid for higher versions.

Note that this patch does *not* address the deprecation warnings
introduced by OpenSSL 3.0.

### DIFF
--- a/adapters/tlsio_openssl.c
+++ b/adapters/tlsio_openssl.c
@@ -953,7 +953,7 @@ static int add_certificate_to_store(TLS_IO_INSTANCE* tls_io_instance, const char
         }
         else
         {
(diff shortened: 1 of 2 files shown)


schopin-pro:
This patch has been written when looking into the build failure of `azure-uamqp-python` in Ubuntu due to our transition to OpenSSL3. It will be shipped in the affected package.

CIPop:
@schopin-pro, this should now be fixed by #604.
We have tested the library builds and works on Ubuntu 22.04 (thanks @danewalton !).

PR #587, opened 2022-03-03
update console logger

danewalton:
Crash discovered here: https://github.com/Azure/azure-iot-sdk-c/issues/2248

Inserts a placeholder in case the timeString is NULL.

### DIFF
--- a/src/consolelogger.c
+++ b/src/consolelogger.c
@@ -157,12 +157,15 @@ void consolelogger_log_with_GetLastError(const char* file, const char* func, int
     timeString = get_ctime(&t);
 #endif // LOGGER_DISABLE_PAL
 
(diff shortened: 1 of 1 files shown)


momuno (approved):

src/consolelogger.c
```diff
@@ -157,12 +157,12 @@ void consolelogger_log_with_GetLastError(const char* file, const char* func, int
     timeString = get_ctime(&t);
 #endif // LOGGER_DISABLE_PAL
 
-    systemMessage = printf_alloc("Error: Time:%.24s File:%s Func:%s Line:%d %s", timeString, file, func, line, lastErrorAsString);
+    systemMessage = printf_alloc("Error: Time:%.24s File:%s Func:%s Line:%d %s", timeString == NULL ? "<NO TIME IMPL>" : timeString, file, func, line, lastErrorAsString);
```
ericwol-msft, line 160: just do this one on line 159 and not at every print?

timeString = (timeString == NULL) ? "<NO TIME IMPL>" : timeString;

ericwol-msft (approved):

PR #592, opened 2022-04-18
Fix for memory leak in http-proxy-io

momuno:
Memory leak found during investigation of logic pertaining to `xio_retrieveoptions`.
Fix could be 2 lines.  Instead, approach chosen to mimic pattern found in other files using parallel logic, such as `tlsio_mbedts.c` and `tlsio_openssl.c` 

Because `OptionHandler_AddOption` mallocs memory for the option-value being added (in this case  `underlying_io_options`), which has already been allocated once by `xio_retrieveoptions`, the memory must be destroyed if  `OptionHandler_AddOption` succeeds.

### DIFF
--- a/src/http_proxy_io.c
+++ b/src/http_proxy_io.c
@@ -1025,18 +1025,29 @@ static OPTIONHANDLER_HANDLE http_proxy_io_retrieve_options(CONCRETE_IO_HANDLE ht
             OPTIONHANDLER_HANDLE underlying_io_options;
 
             /* Codes_SRS_HTTP_PROXY_IO_01_046: [ http_proxy_io_retrieve_options shall return an OPTIONHANDLER_HANDLE obtained by calling xio_retrieveoptions on the underlying IO created in http_proxy_io_create. ]*/
-            if ((underlying_io_options = xio_retrieveoptions(http_proxy_io_instance->underlying_io)) == NULL ||
-                OptionHandler_AddOption(result, OPTION_UNDERLYING_IO_OPTIONS, underlying_io_options) != OPTIONHANDLER_OK)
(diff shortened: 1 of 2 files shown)


momuno:
Will appreciate a second set of eyes on this :) sanity check...

momuno:
Fixing the test...

ewertons (approved):

danewalton (approved):
good find :)

src/http_proxy_io.c
```diff
@@ -1025,18 +1025,29 @@ static OPTIONHANDLER_HANDLE http_proxy_io_retrieve_options(CONCRETE_IO_HANDLE ht
...
             {
-                // All is fine.
+                if (OptionHandler_AddOption(result, OPTION_UNDERLYING_IO_OPTIONS, underlying_io_options) != OPTIONHANDLER_OK)
+                {
+                    /* Codes_SRS_HTTP_PROXY_IO_01_048: [ If xio_retrieveoptions fails, http_proxy_io_retrieve_options shall return NULL. ]*/
+                    LogError("unable to save underlying_io options");
+                    OptionHandler_Destroy(result);
+                    result = NULL;
+                }
+                else
```
CIPop, line 1044: I think this should be removed as it doesn't do anything and it's confusing. (Comment is OK to stay.)

CIPop (approved):
We should remove the last empty `else` statement.

src/http_proxy_io.c
momuno: I hear you.  However, it's how it's done in (I think) every other file that has this same section of logic.  I can go either way.

CIPop: Thanks! I see it now. 
It does seem to be part of the coding standards for the repo so it's OK to follow the same pattern.

PR #595, opened 2022-05-06
Check cURL TLS platform is valid and matches CSDK TLS platform

momuno:
This is no longer a direct memory leak fix.
However, a memory leak can occur if cURL is configured to use a different TLS platform than the C SDK is using (particularly wrt mbedTLS and certificates), when using HTTP.

This fix checks that cURL is using a valid TLS platform that the CSDK supports (wolfSSL, OpenSSL, or mbedTLS), and that that platform matches what the C SDK is configured to use.

Consolidates else-if statements.


ericwol-msft (approved):

adapters/httpapi_curl.c
```diff
@@ -113,20 +113,77 @@ HTTP_HANDLE HTTPAPI_CreateConnection(const char* hostName)
...
+                    int32_t SDKSSL = CURLSSLBACKEND_NONE;
+#ifdef USE_OPENSSL
+                    SDKSSL = CURLSSLBACKEND_OPENSSL;
+                    char* SDKSSLName = "OpenSSL";
+#elif USE_WOLFSSL
+                    SDKSSL = CURLSSLBACKEND_WOLFSSL;
+                    char* SDKSSLName = "wolfSSL";
+#elif USE_MBEDTLS
+                    SDKSSL = CURLSSLBACKEND_MBEDTLS;
+                    char* SDKSSLName = "mbedTLS";
```
danewalton, line 172: we should probably add `schannel` and `bearssl` to this list since those are options in curl as well and we allow those to be configured.

image

danewalton (changes requested):

adapters/httpapi_curl.c
momuno: If either of those are used by cURL, the LogError will be: 
LogError("curl_sslbackend (%d) currently used by cURL is not supported by the C SDK. Please configure and compile cURL to use OpenSSL, wolfSSL, or mbedTLS.", info->backend);

If either of those are used by CSDK, the LogError will be: 
LogError("C SDK TLS platform is not supported to use with cURL. Please configure and compile C SDK to use OpenSSL, wolfSSL, or mbedTLS.");

I don't think its required.  We could add more info to specify what is being used by the C SDK, but that's just more to maintain in the future.  I'd rather keep the logging specifics to only the one's we do support.

momuno: Added BearSSL and tested.
Schannel won't be called by this file, since it is used on Windows.  Can't even compile with use_schannel on linux, since it is missing a windows.h file. CMake has win32 use a different file (httpapi_winhttp.c).

adapters/httpapi_curl.c
```diff
@@ -113,20 +113,87 @@ HTTP_HANDLE HTTPAPI_CreateConnection(const char* hostName)
...
+                free(httpHandleData->hostURL);
+                free(httpHandleData);
+                httpHandleData = NULL;
+            }
             else
             {
-                if ((strcpy_s(httpHandleData->hostURL, hostURL_size, "https://") == 0) &&
-                    (strcat_s(httpHandleData->hostURL, hostURL_size, hostName) == 0))
+                // Check correct TLS library has been configured for cURL.
+                const struct curl_tlssessioninfo *info = NULL;
```
danewalton, line 141: nit for this code base code style
```suggestion
                const struct curl_tlssessioninfo* info = NULL;
```

danewalton (approved):

danewalton:
It might be that the version of curl that we support doesn't yet have bearssl (based on the gate checks). If that's the case I'm good with just a comment noting that and maybe a warning that this curl version doesn't support it.

momuno:
> It might be that the version of curl that we support doesn't yet have bearssl (based on the gate checks). If that's the case I'm good with just a comment noting that and maybe a warning that this curl version doesn't support it.

Oh yikes.

PR #602, opened 2022-07-27
Adding support for OpenSSL engines when using curl

JonathanKnam:


adapters/httpapi_curl.c
```diff
@@ -294,9 +309,17 @@ static CURLcode ssl_ctx_callback(CURL *curl, void *ssl_ctx, void *userptr)
...
+                LogError("unable to load engine by ID: %s", httpHandleData->engineId);
+            }
+        }
         if (
             (httpHandleData->x509certificate != NULL) && (httpHandleData->x509privatekey != NULL) &&
-            (x509_openssl_add_credentials(ssl_ctx, httpHandleData->x509certificate, httpHandleData->x509privatekey, KEY_TYPE_DEFAULT, NULL) != 0)
+            (x509_openssl_add_credentials(ssl_ctx, httpHandleData->x509certificate, httpHandleData->x509privatekey, httpHandleData->x509privatekeytype, httpHandleData->engine) != 0)
            )
         {
             LogError("unable to x509_openssl_add_credentials");
```
CIPop, line 325: Here and everywhere below in the error handling cases, you would need to call `ENGINE_free` to avoid leaking the engine handle.

CIPop (changes requested):
I've left two comments.

For this to go through, we would also need to enable HTTP testing for both the [Provisioning](https://github.com/Azure/azure-iot-sdk-c/tree/main/provisioning_client/tests/prov_x509_client_openssl_engine_e2e) and [IoT Hub](https://github.com/Azure/azure-iot-sdk-c/tree/main/iothub_client/tests/iothubclient_openssl_engine_e2e) tests.

These are currently running for AMQP and MQTT only. 
I assume we would need to change https://github.com/Azure/azure-iot-sdk-c/blob/main/jenkins/linux_openssl_engine.sh#L17 to enable HTTP.

Once the changes are made, I can trigger a test run.

Also, in case you'd like to run these locally: 
- we use this container `docker pull aziotbld/openssl-pkcs11`. 
- running the tests is described here:  https://github.com/Azure/azure-iot-sdk-c/blob/main/build/.vsts-ci.yml#L757

adapters/httpapi_curl.c
```diff
@@ -294,9 +309,17 @@ static CURLcode ssl_ctx_callback(CURL *curl, void *ssl_ctx, void *userptr)
         HTTP_HANDLE_DATA *httpHandleData = (HTTP_HANDLE_DATA *)userptr;
 #ifdef USE_OPENSSL
         /*trying to set the x509 per device certificate*/
+        if (httpHandleData->x509privatekeytype == KEY_TYPE_ENGINE) {
+            ENGINE_load_builtin_engines();
+            httpHandleData->engine = ENGINE_by_id(httpHandleData->engineId);
+            if (httpHandleData->engine == NULL)
+            {
+                LogError("unable to load engine by ID: %s", httpHandleData->engineId);
```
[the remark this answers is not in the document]
JonathanKnam: Basically it wasn't really necessary to set the result, as it would run into the error case when the NULL was passed to x509_openssl_add_credentials below. Nevertheless I changed it to be more clear.

The main intention was, to provide a more detailed error message, otherwise this could have been skipped completely.

adapters/httpapi_curl.c
JonathanKnam: I've added ENGINE_free now.

CIPop:
@JonathanKnam this looks good to me! Thank you for the contribution!

I've started a test run to check for any regressions. If that passes, I need to enable existing OpenSSL ENGINE testing in the Azure IoT C-SDK repo and perform a manual run.

CIPop (commented):
LGTM! Thanks @JonathanKnam ! 

Not yet approved until we will manually test this with C-SDK and enable automatic testing for HTTP (tests already exist).

adapters/httpapi_curl.c
```diff
@@ -1014,10 +1068,50 @@ HTTPAPI_RESULT HTTPAPI_CloneOption(const char* optionName, const void* value, co
...
+            }
+            else
+            {
+                LogError("Unknown x509PrivatekeyType: %i", type);
+                result = HTTPAPI_ERROR;
+            }
+        }
+        else if (strcmp(OPTION_OPENSSL_ENGINE, optionName) == 0)
+        {
+            /*this is getting the engine. In this case, value is a pointer to a const char* that contains the engine as a null terminated string*/
```
CIPop, line 1102: Crashing here: the value is a char* not a char**.

CIPop (changes requested):

adapters/httpapi_curl.c
```diff
@@ -1014,10 +1068,50 @@ HTTPAPI_RESULT HTTPAPI_CloneOption(const char* optionName, const void* value, co
...
             {
-                /*return OK when the certificate has been clones successfully*/
+                /*return OK when the certificate has been cloned successfully*/
                 result = HTTPAPI_OK;
             }
         }
+#ifdef USE_OPENSSL
+        else if (strcmp(OPTION_OPENSSL_PRIVATE_KEY_TYPE, optionName) == 0)
+        {
+            const OPTION_OPENSSL_KEY_TYPE type = **(const OPTION_OPENSSL_KEY_TYPE**)value;
```
CIPop, line 1078: Crashes here as well: too many indirections.
It should be `const OPTION_OPENSSL_KEY_TYPE type = *(const OPTION_OPENSSL_KEY_TYPE*)value;`

adapters/httpapi_curl.c
CIPop: This value is passed similarly to line 1118 below: 
`if (mallocAndStrcpy_s((char**)savedValue, value) != 0)`

CIPop (changes requested):
Even with fixes pointed out in the comments, the tests are still failing as no callback is made.
Please investigate the IoT Hub Client and Provisioning Client tests. I have enabled HTTP tests in this branch: https://github.com/CIPop/azure-iot-sdk-c/tree/curl-openssl-engine

adapters/httpapi_curl.c
JonathanKnam: Thanks, I missed that. I fixed it now.

adapters/httpapi_curl.c
JonathanKnam: Fixed it.

JonathanKnam:
> Even with fixes pointed out in the comments, the tests are still failing as no callback is made. Please investigate the IoT Hub Client and Provisioning Client tests. I have enabled HTTP tests in this branch: https://github.com/CIPop/azure-iot-sdk-c/tree/curl-openssl-engine

Ok, I'll have a look after I checked, how I need to setup the environment variables (especially the private key & certificates in the expected format), so that I can run the tests.

CIPop:
> Ok, I'll have a look after I checked, how I need to setup the environment variables (especially the private key & certificates in the expected format), so that I can run the tests.

Thank you @JonathanKnam !

The test suite for OpenSSL ENGINE is this one : https://github.com/Azure/azure-iot-sdk-c/blob/main/build/.vsts-ci.yml#L754 ('OpenSSL with pkcs11 engine and SoftHSM').

The docker image is also available here:  https://hub.docker.com/r/aziotbld/openssl-pkcs11

I usually:
1.  run the image interactively:
`docker run -it <image> /bin/bash`
2. Clone the repo with submodules
3. Set the environment (ignore DPS for now)
   1. `IOTHUB_CONNECTION_STRING` - IoT Hub Owner connection string (tests will create new devices)
   5. `IOTHUB_E2E_X509_CERT_BASE64` - your certificate PEM, base64 encoded (cat my.pem | base64 -encode)
   6. `IOTHUB_E2E_X509_PRIVATE_KEY_BASE64` - your private key PEM (no password), base64 encoded
   7. `IOTHUB_E2E_X509_THUMBPRINT` - the certificate thumbprint (obtained with ` openssl x509 -in ./cert.pem -fingerprint -noout`)
   8. `IOTHUB_EVENTHUB_CONNECTION_STRING` - the event hub connection string (obtained from IoT Hub built-in endpoints)
   9. `IOTHUB_EVENTHUB_CONSUMER_GROUP` - the event hub consumer group
10. Run the same scripts as the YML:
    1. `sudo -E bash jenkins/linux_openssl_engine.sh`
    12. `cd cmake && sudo -E ../build_all/linux/run_tests.sh`

CIPop:
A recent change to allow compilation on Ubuntu 22.04 requires adding new source files using OpenSSL ENGINE functions within the CMake exceptions script:

https://github.com/Azure/azure-c-shared-utility/commit/9e4322c9d88095ed276bf7e4b31b87b82b1c4d86#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a

CIPop:
Thank you for your patience @JonathanKnam ! 
I fixed another memory leak for `engine_id` and found that the HTTP test was failing because of a test issue, not your code.

Everything appears to be working now end-to-end. I will be merging this PR then update the submodules within all Azure IoT C-SDK dependencies.

CIPop (approved):
Thanks @JonathanKnam !

PR #604, opened 2022-08-02
add openssl 3 compat and ubuntu 22.04

danewalton:


adapters/tlsio_openssl.c
```diff
@@ -1238,7 +1238,7 @@ void tlsio_openssl_deinit(void)
 {
     openssl_dynamic_locks_uninstall();
     openssl_static_locks_uninstall();
-#if  (OPENSSL_VERSION_NUMBER >= 0x00907000L) &&  (OPENSSL_VERSION_NUMBER < 0x20000000L) && (FIPS_mode_set)
+#if  (OPENSSL_VERSION_NUMBER >= 0x00907000L) && (FIPS_mode_set)
```
ewertons, line 1241: Should we just change the check to < 0x30000000L ?

ewertons: Just to make sure.

ewertons: You know, so we get a notification somehow if a new version comes out. Gives us the chance to check it out and make sure it works.

ewertons: Same comment in all the checks that were modified.

adapters/x509_openssl.c
```diff
@@ -75,7 +75,7 @@ static int load_certificate_chain(SSL_CTX* ssl_ctx, const char* certificate)
                 // certificates.
 
                 /* Codes_SRS_X509_OPENSSL_07_006: [ If successful x509_openssl_add_ecc_credentials shall to import each certificate in the cert chain. ] */
-#if (OPENSSL_VERSION_NUMBER >= 0x10100000L) && (OPENSSL_VERSION_NUMBER < 0x20000000L) || defined(LIBRESSL_VERSION_NUMBER)
+#if (OPENSSL_VERSION_NUMBER >= 0x10100000L) || defined(LIBRESSL_VERSION_NUMBER)
```
ewertons, line 78: Also, is any version of LIBRESSL ok with these API calls?

ewertons (approved):
Approved with comments

adapters/x509_openssl.c
danewalton: So far yes. I've seen this kind of behavior in other places as well.

adapters/tlsio_openssl.c
danewalton: We could but I would say we keep it removed until there comes a version that actually removes these APIs. At which point we can set the guard on that version. Otherwise it's a constantly moving goal post.

CIPop (approved):

Issue #605, opened 2023-01-11
socketio_berkeley: recv with MSG_NOSIGNAL causes errno 22 (EINVAL)


AndWass:
Version: 9e4322c9d88095ed276bf7e4b31b87b82b1c4d86 (used by [azure-iot-sdk-c LTS_07_2022_Ref02](https://github.com/Azure/azure-iot-sdk-c/tree/LTS_07_2022_Ref02))

System information:
Linux Yocto with kernel 3.18.140

Error output:
```
[error][azure-iot-sdk-c/c-utility/adapters/socketio_berkeley.c] [socketio_dowork:1068]: Socketio_Failure: Receiving data from endpoint: errno=22.
[error] IoT hub sdk: [azure-iot-sdk-c/umqtt/src/mqtt_client.c] [onOpenComplete:452]: Error: failure opening connection to endpoint
[info] IoT hub: connection status: result = IOTHUB_CLIENT_CONNECTION_UNAUTHENTICATED, reason = IOTHUB_CLIENT_CONNECTION_NO_NETWORK
```

Manually patching the `recv` call to pass 0 as flags instead of `MSG_NOSIGNAL` does not yield the error above and communication is starting to work.

danewalton:
Hi @AndWass 
I'll respond on the PR itself and we can discuss the scenario there.

PR #606, opened 2023-01-11
Do not pass MSG_NOSIGNAL as flag to `recv`

AndWass:
Related issue: https://github.com/Azure/azure-c-shared-utility/issues/605

### DIFF
--- a/adapters/socketio_berkeley.c
+++ b/adapters/socketio_berkeley.c
@@ -1049,7 +1049,7 @@ void socketio_dowork(CONCRETE_IO_HANDLE socket_io)
                 ssize_t received = 0;
                 do
                 {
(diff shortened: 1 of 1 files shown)


danewalton:
Hi @AndWass 
The flag for `MSG_NOSIGNAL` was explicitly added in this PR: https://github.com/Azure/azure-c-shared-utility/pull/479

It should only suppress the SIGPIPE error, reducing the number of errors this would actually generate. This change, as I understand, would create more error scenarios.

danewalton:
https://www.man7.org/linux/man-pages/man2/recv.2.html

Doesn't seem to be a flag for `recv()`

AndWass:
That was my reading/take on it as well.

danewalton (approved):

PR #623, opened 2023-05-01
Adding IPv6 Dual Stack Support for Linux

RLeclair:
Adding IPv6 Dual Stack support for Linux
- Enabled with cmake option.
- Moved socket creation to after DNS query to enable IPv6 socket creation.
- Added handling of IPv6 addresses for Linux async and sync dns.

### DIFF
--- a/adapters/socketio_berkeley.c
+++ b/adapters/socketio_berkeley.c
@@ -494,12 +494,7 @@ static int initiate_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
     int result;
     int flags;
     struct addrinfo* addr = NULL;
-    struct sockaddr* connect_addr = NULL;
-#ifdef IPV6_ENABLED
-    struct sockaddr_in6* connect_addr6 = NULL;
-#endif // IPV6_ENABLED
     struct sockaddr_un addrInfoUn;
-    socklen_t connect_addr_len;
 
     if(socket_io_instance->address_type == ADDRESS_TYPE_IP)
     {
@@ -516,19 +511,9 @@ static int initiate_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
             {
                 LogError("DNS resolution failed");
                 result = MU_FAILURE;
-            }
-#ifdef IPV6_ENABLED
-            else if (addr->ai_family == AF_INET6)
+            } 
+            else 
             {
-                connect_addr6 = (struct sockaddr_in6*) addr->ai_addr;
-                connect_addr_len = sizeof(struct sockaddr_in6);
-                result = 0;
-            }
-#endif // IPV6_ENABLED
-            else
-            {
-                connect_addr = addr->ai_addr;
-                connect_addr_len = sizeof(*addr->ai_addr);
                 result = 0;
             }
         }
@@ -547,14 +532,17 @@ static int initiate_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
             addrInfoUn.sun_family = AF_UNIX;
             // No need to add NULL terminator due to the above memset
             (void)memcpy(addrInfoUn.sun_path, socket_io_instance->hostname, hostname_len);
-
-            connect_addr = (struct sockaddr*)&addrInfoUn;
-            connect_addr_len = sizeof(addrInfoUn);
+            
+            addr->ai_addrlen = sizeof(addrInfoUn);
+            addr->ai_addr = (struct sockaddr*)&addrInfoUn;
+            addr->ai_family = AF_UNIX;
+            addr->ai_socktype = SOCK_STREAM;
+            addr->ai_protocol = 0;
             result = 0;
         }
     }
 
-    socket_io_instance->socket = socket (socket_io_instance->address_type == ADDRESS_TYPE_IP ? addr->ai_family : AF_UNIX, SOCK_STREAM, 0);
+    socket_io_instance->socket = socket(addr->ai_family, addr->ai_socktype, addr->ai_protocol);
 
     if (socket_io_instance->socket < SOCKET_SUCCESS)
     {
@@ -580,18 +568,7 @@ static int initiate_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
         }
         else
         {
-#ifndef IPV6_ENABLED
-            result = connect(socket_io_instance->socket, connect_addr, connect_addr_len);
-#else
-            if (addr->ai_family == AF_INET6)
-            {
-                result = connect(socket_io_instance->socket, (struct sockaddr*)connect_addr6, connect_addr_len);
-            }
-            else 
-            {
-                result = connect(socket_io_instance->socket, connect_addr, connect_addr_len);
-            }
-#endif // IPV6_ENABLED
+            result = connect(socket_io_instance->socket, addr->ai_addr, addr->ai_addrlen);
             
             if ((result != 0) && (errno != EINPROGRESS))
             {
--- a/adapters/socketio_berkeley.c
+++ b/adapters/socketio_berkeley.c
@@ -287,6 +287,7 @@ static int initiate_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
     int flags;
     struct addrinfo* addr = NULL;
     struct sockaddr* connect_addr = NULL;
+    struct sockaddr_in6* connect_addr6 = NULL;
     struct sockaddr_un addrInfoUn;
     socklen_t connect_addr_len;
 
@@ -306,11 +307,21 @@ static int initiate_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
                 LogError("DNS resolution failed");
                 result = MU_FAILURE;
             }
-            else
+            else if (addr->ai_family == AF_INET)
             {
                 connect_addr = addr->ai_addr;
                 connect_addr_len = sizeof(*addr->ai_addr);
                 result = 0;
+            } 
+            else if (addr->ai_family == AF_INET6)
+            {
+                connect_addr6 = (struct sockaddr_in6*) addr->ai_addr;
+                connect_addr_len = sizeof(struct sockaddr_in6);
+                result = 0;
+            }
+            else {
+                LogError("Socket failure...");
+                result = MU_FAILURE;
             }
         }
     }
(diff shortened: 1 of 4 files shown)


src/dns_resolver_ares.c
```diff
@@ -126,36 +140,60 @@ static void query_completed_cb(void *arg, int status, int timeouts, struct hoste
...
-        else
+#ifdef IPV6_ENABLED
+        else if (he->h_addrtype == AF_INET6)
         {
             ptr = dns->addrInfo;
-            
-            ptr->ai_addr = calloc(1, sizeof(struct sockaddr_in));
+
+            ptr->ai_addr = calloc(1, sizeof(struct sockaddr_in6));
+            addr6 = (void *)ptr->ai_addr;
```
ewertons, line 149: Why not casting to `struct sockaddr_in6*` ?

src/dns_resolver_ares.c
```diff
@@ -126,36 +140,60 @@ static void query_completed_cb(void *arg, int status, int timeouts, struct hoste
...
                 dns->is_failed = true;
                 dns->is_complete = true;
                 dns->in_progress = false;
+            } else {
+                memcpy(&addr6->sin6_addr, he->h_addr_list[0], sizeof(struct in6_addr));
+                addr6->sin6_family = he->h_addrtype;
+                addr6->sin6_port = htons((unsigned short)dns->port);
+
+                /* Codes_SRS_dns_resolver_30_033: [ If dns_resolver_is_create_complete has returned true and the lookup process has failed, dns_resolver_get_ipv4 shall return 0. ]*/
+                memcpy(dns->ip_v6, EXTRACT_IPV6(ptr), 16); // IPv6 address is 16 bytes
```
ewertons, line 164: Thanks.

src/dns_resolver_ares.c
ewertons: I see, you just followed what was done for IPv4

src/dns_resolver_sync.c
```diff
@@ -130,11 +145,20 @@ bool dns_resolver_is_lookup_complete(DNSRESOLVER_HANDLE dns_in)
                     case AF_INET:
                         /* Codes_SRS_dns_resolver_30_032: [ If dns_resolver_is_create_complete has returned true and the lookup process has succeeded, dns_resolver_get_ipv4 shall return the discovered IPv4 address. ]*/
                         dns->ip_v4 = EXTRACT_IPV4(ptr);
+#ifdef IPV6_ENABLED
```
ewertons, line 148: Why not enclosing just lines 151 to 155 instead?

ewertons (approved):

src/dns_resolver_ares.c
```diff
@@ -126,36 +140,60 @@ static void query_completed_cb(void *arg, int status, int timeouts, struct hoste
...
-            else
-            {
-                addr = (void *)ptr->ai_addr;
+        }
+#endif // IPV6_ENABLED
+        else
+        {
+            ptr = dns->addrInfo;
 
-                if (he->h_addrtype == AF_INET)
```
vaavva, line 146: Should this if statement still be in the IPv4 logic?

src/dns_resolver_sync.c
```diff
@@ -130,11 +145,20 @@ bool dns_resolver_is_lookup_complete(DNSRESOLVER_HANDLE dns_in)
...
+                    case AF_INET6:
+                        memcpy(dns->ip_v6, EXTRACT_IPV6(ptr), 16);
+                        dns->is_failed = false;
+#endif // IPV6_ENABLED
+                        break;     
                     }
                 }
                 /* Codes_SRS_dns_resolver_30_033: [ If dns_resolver_is_create_complete has returned true and the lookup process has failed, dns_resolver_get_ipv4 shall return 0. ]*/
+#ifndef IPV6_ENABLED
                 dns->is_failed = (dns->ip_v4 == 0);
```
vaavva, line 160: Does this line still need to be run for ipv4 only?

vaavva (approved):

src/dns_resolver_ares.c
```diff
@@ -61,6 +69,9 @@ DNSRESOLVER_HANDLE dns_resolver_create(const char* hostname, int port, const DNS
             result->is_failed = false;
             result->in_progress = false;
             result->ip_v4 = 0;
+#ifdef IPV6_ENABLED
```
ericwol-msft, line 72: change malloc to calloc() ?

src/dns_resolver_ares.c
```diff
@@ -126,36 +140,60 @@ static void query_completed_cb(void *arg, int status, int timeouts, struct hoste
...
+            addr6 = (void *)ptr->ai_addr;
+
             if(ptr->ai_addr == NULL)
             {
                 LogError("dns addrinfo ai_addr: allocation failed");
                 free(dns->addrInfo);
                 dns->is_failed = true;
                 dns->is_complete = true;
                 dns->in_progress = false;
+            } else {
```
ericwol-msft, line 158:
```suggestion
            } 
            else 
            {
```

src/dns_resolver_ares.c
```diff
@@ -126,36 +140,60 @@ static void query_completed_cb(void *arg, int status, int timeouts, struct hoste
...
-                    dns->in_progress = false;
-                }
+            if(ptr->ai_addr == NULL)
+            {
+                LogError("dns addrinfo ai_addr: allocation failed");
+                free(dns->addrInfo);
+                dns->is_failed = true;
+                dns->is_complete = true;
+                dns->in_progress = false;
+            } else {
```
ericwol-msft, line 186:
```suggestion
            } 
            else 
            {
```

src/dns_resolver_sync.c
```diff
@@ -57,6 +65,9 @@ DNSRESOLVER_HANDLE dns_resolver_create(const char* hostname, int port, const DNS
             result->is_complete = false;
             result->is_failed = false;
             result->ip_v4 = 0;
+#ifdef IPV6_ENABLED
+            memset(result->ip_v6, 0, 16); // Zero out the IPv6 address
```
ericwol-msft, line 69: use calloc()

ericwol-msft (approved):

adapters/socketio_berkeley.c
```diff
@@ -659,6 +500,220 @@ static int set_target_network_interface(int socket, char* mac_address)
...
+            result = MU_FAILURE;
+        }
+        else
+        {
+#ifndef IPV6_ENABLED
+            result = connect(socket_io_instance->socket, connect_addr, connect_addr_len);
+#else
+            if (addr->ai_family == AF_INET6)
+            {
+                result = connect(socket_io_instance->socket, (struct sockaddr*)connect_addr6, connect_addr_len);
```
CIPop, line 599: This should be the same connect, using the DNS record's ai_addr and ai_addrlen. Please remove the cast.

CIPop: See https://gist.github.com/CIPop/e29729d1c8d6871372f38c8cbcc43f93#file-client_ipv4_ipv6-c-L62

adapters/socketio_berkeley.c
```diff
@@ -659,6 +500,220 @@ static int set_target_network_interface(int socket, char* mac_address)
...
+        {
+            LogError("Failure: fcntl failure.");
+            result = MU_FAILURE;
+        }
+        else
+        {
+#ifndef IPV6_ENABLED
+            result = connect(socket_io_instance->socket, connect_addr, connect_addr_len);
+#else
+            if (addr->ai_family == AF_INET6)
```
CIPop, line 597: This isn't necessary for dual-mode sockets.

src/dns_resolver_ares.c
CIPop: I'm not very happy with all the casting either. A `freeaddrinfo` call is recommended to deallocated these records.
See https://gist.github.com/CIPop/e29729d1c8d6871372f38c8cbcc43f93#file-client_ipv4_ipv6-c-L75

CIPop (changes requested):
1. The first function location changed in the file: that leaves us with a large-block diff that I would like to avoid (and most of the function is identical by doing a side-by-side comparison).
2. I left a few comments about the implementation of dual-mode. Most of the Sockets API is designed to re-use the same exact code (if written at the correct layer level). Please see https://gist.github.com/CIPop/e29729d1c8d6871372f38c8cbcc43f93 for a short example using DNS `getnameinfo` then using the _same_ code to connect to either IPv4 or v6. The example is courtesy of Kerrisk / TLPI (https://www.man7.org/tlpi/index.html).

CIPop (changes requested):
1. The first function location changed in the file: that leaves us with a large-block diff that I would like to avoid (and most of the function is identical by doing a side-by-side comparison).
2. I left a few comments about the implementation of dual-mode. Most of the Sockets API is designed to re-use the same exact code (if written at the correct layer level). Please see https://gist.github.com/CIPop/e29729d1c8d6871372f38c8cbcc43f93 for a short example using DNS `getnameinfo` then using the _same_ code to connect to either IPv4 or v6. The example is courtesy of Kerrisk / TLPI (https://www.man7.org/tlpi/index.html).

adapters/socketio_berkeley.c
```diff
@@ -659,6 +500,220 @@ static int set_target_network_interface(int socket, char* mac_address)
...
+            // No need to add NULL terminator due to the above memset
+            (void)memcpy(addrInfoUn.sun_path, socket_io_instance->hostname, hostname_len);
+
+            connect_addr = (struct sockaddr*)&addrInfoUn;
+            connect_addr_len = sizeof(addrInfoUn);
+            result = 0;
+        }
+    }
+
+    socket_io_instance->socket = socket (socket_io_instance->address_type == ADDRESS_TYPE_IP ? addr->ai_family : AF_UNIX, SOCK_STREAM, 0);
```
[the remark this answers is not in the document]
RLeclair: Changed to `record->ai_protocol` as suggested

src/dns_resolver_ares.c
RLeclair: Good catch! Added it back

src/dns_resolver_sync.c
RLeclair: `dns->is_failed` is set to false below without IP_v4, I enclosed the correct lines now and added a second `dns->is_failed` check for IPv6 below

adapters/socketio_berkeley.c
```diff
@@ -659,6 +500,220 @@ static int set_target_network_interface(int socket, char* mac_address)
 }
 #endif //__APPLE__
 
+static int initiate_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
```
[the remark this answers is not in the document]
RLeclair: Here is the diff for reference: https://github.com/Azure/azure-c-shared-utility/compare/master...RLeclair:azure-c-shared-utility:raulleclair/ipv6-comp

We are now calling `set_target_network_interface` here because of the socket move so we needed to move this function down

CIPop (approved):
Thanks @RLeclair !

PR #624, opened 2023-05-01
SDL bugfixes (part 2)

vaavva:
17992325 - Defect : UsingUninitVar, Component : c-utility\src\uws_client.c
[warning]c-utility\src\singlylinkedlist.c(48,0): Warning C6001: Using uninitialized memory '*list_instance->head'.

### DIFF
--- a/src/uws_client.c
+++ b/src/uws_client.c
@@ -572,6 +572,7 @@ void uws_client_destroy(UWS_CLIENT_HANDLE uws_client)
             }
 
             free(uws_client->protocols);
+            uws_client->protocols = NULL;
         }
 
         /* Codes_SRS_UWS_CLIENT_01_019: [ uws_client_destroy shall free all resources associated with the uws instance. ]*/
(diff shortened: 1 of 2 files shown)


src/uws_client.c
```diff
@@ -568,6 +568,7 @@ void uws_client_destroy(UWS_CLIENT_HANDLE uws_client)
             for (i = 0; i < uws_client->protocol_count; i++)
             {
                 free(uws_client->protocols[i].protocol);
+                uws_client->protocols[i].protocol = NULL;
             }
 
             free(uws_client->protocols);
```
ericwol-msft, line 574: does this need a null assignment as well? Seams like they are crying about it in the other places.

ericwol-msft (approved):

src/uws_client.c
vaavva: It only seems to be complaining about the ones that are inside of arrays, but I could add it here too to be safe!

PR #613, opened 2023-03-30
Remove check for return value of strcpy

szsam:
The return value for strcpy, strncpy, or related string copy functions have no reserved return value to indicate an error.

PR #625, opened 2023-05-03
Remove check for return value of strcpy

ewertons:
The return value for strcpy, strncpy, or related string copy functions have no reserved return value to indicate an error.

This is a port of https://github.com/Azure/azure-c-shared-utility/pull/613

### DIFF
--- a/adapters/socketio_berkeley.c
+++ b/adapters/socketio_berkeley.c
@@ -130,11 +130,9 @@ static void* socketio_CloneOption(const char* name, const void* value)
                 {
                     LogError("Failed cloning option %s (malloc failed)", name);
                 }
-                else if (strcpy((char*)result, (char*)value) == NULL)
+                else
                 {
-                    LogError("Failed cloning option %s (strcpy failed)", name);
-                    free(result);
-                    result = NULL;
+                    strcpy((char *)result, (char *)value);
                 }
             }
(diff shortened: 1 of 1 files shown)


adapters/socketio_berkeley.c
```diff
@@ -485,14 +483,10 @@ static NETWORK_INTERFACE_DESCRIPTION* create_network_interface_description(struc
         destroy_network_interface_descriptions(result);
         result = NULL;
     }
-    else if (strcpy(result->name, ifr->ifr_name) == NULL)
-    {
-        LogError("failed setting interface description name (strcpy failed)");
```
vaavva, line 490: Is there any way we should still be checking for these error states (like result->name != NULL after the strcpy), or would that not be checking anything either?

vaavva (approved):

adapters/socketio_berkeley.c
ewertons: That would not be checking anything. 
According to the C99 spec all `strcpy` does it to return the pointer of the first argument. There are no failures reported by this function.

ericwol-msft: result->name is the same before and after the call. The only protections is to use strncpy just incase the math on line 482 is wrong.

PR #626, opened 2023-05-03
UUID_from_string: fix UUID_T overflow

ewertons:
This is a port of https://github.com/Azure/azure-c-shared-utility/pull/314

### DIFF
--- a/src/uuid.c
+++ b/src/uuid.c
@@ -12,7 +12,13 @@
 #define UUID_STRING_LENGTH          36
 #define UUID_STRING_SIZE            (UUID_STRING_LENGTH + 1)
 #define __SUCCESS__                 0
-#define UUID_FORMAT_STRING          "%02x%02x%02x%02x-%02x%02x-%02x%02x-%02x%02x-%02x%02x%02x%02x%02x%02x"
+#define HEXA_DIGIT_VAL(c) \
+    ((((c) >= '0') && ((c) <= '9')) \
+        ? ((c) - '0') \
+        : (((c) >= 'a') && ((c) <= 'f')) \
+            ? ((c) - 'a' + 10) \
+            : (((c) >= 'A') && ((c) <= 'F')) ? ((c) - 'A' + 10) : -1)
+
 
 int UUID_from_string(const char* uuid_string, UUID_T* uuid)
 {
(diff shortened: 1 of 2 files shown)


src/uuid.c
```diff
@@ -12,7 +12,15 @@
 #define UUID_STRING_LENGTH          36
 #define UUID_STRING_SIZE            (UUID_STRING_LENGTH + 1)
 #define __SUCCESS__                 0
-#define UUID_FORMAT_STRING          "%02x%02x%02x%02x-%02x%02x-%02x%02x-%02x%02x-%02x%02x%02x%02x%02x%02x"
+#define HEXA_DIGIT_VAL(c) \
+    ((((c) >= '0') && ((c) <= '9')) \
+        ? ((c) - '0') \
+        : (((c) >= 'a') && ((c) <= 'f')) \
+            ? ((c) - 'a' + 10) \
+            : (((c) >= 'A') && ((c) <= 'F')) ? ((c) - 'A' + 10) : -1)
```
vaavva, line 20:
```suggestion
#define HEXA_DIGIT_VAL(c)                                             \
    do {                                                              \
        ((((c) >= '0') && ((c) <= '9'))                               \
          ? ((c) - '0')                                               \
          : (((c) >= 'a') && ((c) <= 'f'))                            \
            ? ((c) - 'a' + 10)                                        \
            : (((c) >= 'A') && ((c) <= 'F')) ? ((c) - 'A' + 10) : -1) \
    } while(0)
```

src/uuid.c
```diff
@@ -70,6 +87,7 @@ int UUID_from_string(const char* uuid_string, UUID_T* uuid)
                     }
                     else
                     {
+                        uuid_bytes[j] = (char)(higherOrderDigit * 16 + lowerOrderDigit);
```
vaavva, line 90: nit: can this be bit shifted instead of multiplied? (since it looks like that's the goal)

tests/uuid_ut/uuid_ut.c
```diff
@@ -321,6 +321,43 @@ TEST_FUNCTION(UUID_from_string_succeed)
     }
 }
 
+TEST_FUNCTION(UUID_from_string_invalid_fails)
+{
+    //Arrange
+    int result;
+    UUID_T uuid;
+    const char* invalid_uuids[] = {
```
vaavva, line 329: should this be defined at the top of the file?

vaavva (approved):

src/uuid.c
ewertons: The do while zero is useful if there were several functions. In this case we only have a long conditional statement, and it would be ok not to be in the do while zero form.

src/uuid.c
ewertons: That's a good idea. Most compilers will be smart enough to optimize this though. But let's do it.

tests/uuid_ut/uuid_ut.c
ewertons: Ok.

ewertons: Done.

src/uuid.c
vaavva: Suggestion to just align the `\`s then for readability!

src/uuid.c
ewertons: Done.

Issue #622, opened 2023-04-27
hard requirement on CURL 


Kevin0626:
Hi，

we have software using a previous version of c-utility and recently we want to update it to the latest version. Previously we turned off the option "use_http" to reduce the reliance on CURL and other components. 

After upgrading to the latest version, we found the option is no longer available. The following file has made it a must-have. .
https://github.com/Azure/azure-c-shared-utility/blob/master/configs/azure_c_shared_utilityConfig.cmake
```
if(UNIX)
    include(CMakeFindDependencyMacro)
    find_dependency(CURL)
endif()
```

@momuno, we found this change was introduced in your commitment below. We just wonder if it is possible to restore the "use_http" option. Our software runs on an embedded ARM system and currently has no need to use CURL. Thank you very much!

https://github.com/Azure/azure-c-shared-utility/commit/73a2039cc0de5ececf5755ebe731a6b5b98f85da

ewertons:
Issue fixed. If you would like to follow up please reopen it.
Thank you for your contribution.
-Azure SDKs Team.

PR #629, opened 2023-05-16
Fix library when using use_http=OFF (gh#622)

ewertons:
This fix addresses https://github.com/Azure/azure-c-shared-utility/issues/622

Tested on Ubuntu 20.04 using -Duse_http=ON and OFF, with and without curl (and libcurl-dev) installed and uninstalled.
Also tested with azure-iot-sdk-c locally with the use_http=ON/OFF.

### DIFF
--- a/adapters/platform_linux.c
+++ b/adapters/platform_linux.c
@@ -4,7 +4,9 @@
 #include "azure_c_shared_utility/platform.h"
 #include "azure_c_shared_utility/xio.h"
 #include "azure_c_shared_utility/xlogging.h"
+#ifdef USE_HTTP
 #include "azure_c_shared_utility/httpapiex.h"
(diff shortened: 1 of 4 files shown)


adapters/platform_linux.c
```diff
@@ -31,11 +31,13 @@ int platform_init(void)
 {
     int result = 0;
 #ifndef DONT_USE_UPLOADTOBLOB
+#if !defined(HTTP_DISABLED)
```
vaavva, line 34: Should the include for `#include "azure_c_shared_utility/httpapiex.h"` be conditional as well?

ewertons: Fixed!

vaavva (approved):

ericwol-msft (approved):

PR #635, opened 2023-07-21
Enable libaziotsharedutil so versioning

besquivel-pivotal:
Use the C_SHARED_VERSION on the aziotsharedutil_dll.so library for versioning. Have cmake build the symlinks required for proper so library versioning.

**Testing**
- Successful build via the azure-iot-sdk-c.
- Verified that the so versioning uses the version.txt string.
- Verified that the so library and versioning symlinks are created
/usr/lib/libaziotsharedutil_dll.so
/usr/lib/libaziotsharedutil_dll.so.1
/usr/lib/libaziotsharedutil_dll.so.1.1.12

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,10 @@ endif()
 project(azure_c_shared_utility)
 
 FILE(READ ${CMAKE_CURRENT_LIST_DIR}/version.txt C_SHARED_VERSION)
-string(REGEX MATCH "^([0-9]+)[\\.]([0-9]+)[\\.]([0-9]+)$" _ ${C_SHARED_VERSION})
+string(STRIP "${C_SHARED_VERSION}" C_SHARED_VERSION)
+if(NOT C_SHARED_VERSION MATCHES "^([0-9]+)\\.([0-9]+)\\.([0-9]+)$")
+    message(FATAL_ERROR "Malformed version in version.txt: '${C_SHARED_VERSION}'")
+endif()
 set(C_SHARED_VERSION_MAJOR ${CMAKE_MATCH_1})
 
 #bring in dependencies
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,8 @@ endif()
 project(azure_c_shared_utility)
 
 FILE(READ ${CMAKE_CURRENT_LIST_DIR}/version.txt C_SHARED_VERSION)
+string(REGEX MATCH "^([0-9]+)[\\.]([0-9]+)[\\.]([0-9]+)$" _ ${C_SHARED_VERSION})
+set(C_SHARED_VERSION_MAJOR ${CMAKE_MATCH_1})
 
 #bring in dependencies
 #do not add or build any tests of the dependencies
@@ -598,7 +600,11 @@ if(${build_as_dynamic})
         ${source_h_files}
         ${def_files}
     )
-    set_target_properties(aziotsharedutil_dll PROPERTIES OUTPUT_NAME "aziotsharedutil_dll")
+    set_target_properties(aziotsharedutil_dll PROPERTIES
+        OUTPUT_NAME "aziotsharedutil_dll"
+        VERSION ${C_SHARED_VERSION}
+        SOVERSION ${C_SHARED_VERSION_MAJOR}
+    )
 endif()
 
 set(aziotsharedutil_target_libs)


ewertons:
Reviewed against master `654c2ec`. **Recommendation: take it, after a rebase and two corrections.** The direction is right and matches what `azure-iot-sdk-c` already does for `iothub_client_dll` / `prov_device_client_dll` (`VERSION ${IOT_SDK_VERSION}` / `SOVERSION ${IOT_SDK_VERSION_MAJOR}`).

### Findings

1. **Premise needs correcting.** Master already emits a SONAME for `aziotsharedutil_dll`, just an unversioned one: `readelf -d libaziotsharedutil_dll.so` → `SONAME [libaziotsharedutil_dll.so]`. What is missing is the version component and the symlink chain, which is what blocks `dh_makeshlibs`/rpm dependency generation and the runtime-vs-dev package split.

2. **The version parsing is fragile.** `FILE(READ)` is verbatim, and `version.txt` currently has no trailing newline — the only reason the anchored regex matches. Add one and it fails silently, `CMAKE_MATCH_1` is empty, `SOVERSION` expands to nothing, and configure dies with `set_target_properties called with incorrect number of arguments` (reproduced). The diff below adds `string(STRIP)` plus explicit validation, so bad input fails with a readable message. `STRIP` also makes the existing `write_basic_package_version_file(VERSION ${C_SHARED_VERSION} ...)` newline-proof, which it is not today.

3. **Minor:** `string(REGEX MATCH ... _ ...)` clobbers a top-level variable named `_`. `if(... MATCHES ...)` populates `CMAKE_MATCH_<n>` identically without it.

4. **`SOVERSION` = major is the right choice.** `version.txt` has been pinned at `1.1.12` since 2018, so tracking the major is operationally identical to a deliberate `SOVERSION=1` pin: the SONAME cannot drift, only an explicit major bump moves it.

### Recommended diff (rebased onto `654c2ec`)

```diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,11 @@ endif()
 project(azure_c_shared_utility)
 
 FILE(READ ${CMAKE_CURRENT_LIST_DIR}/version.txt C_SHARED_VERSION)
+string(STRIP "${C_SHARED_VERSION}" C_SHARED_VERSION)
+if(NOT C_SHARED_VERSION MATCHES "^([0-9]+)\.([0-9]+)\.([0-9]+)$")
+    message(FATAL_ERROR "Malformed version in version.txt: '${C_SHARED_VERSION}'")
+endif()
+set(C_SHARED_VERSION_MAJOR ${CMAKE_MATCH_1})
 
 #bring in dependencies
 #do not add or build any tests of the dependencies
@@ -598,7 +603,11 @@ if(${build_as_dynamic})
         ${source_h_files}
         ${def_files}
     )
-    set_target_properties(aziotsharedutil_dll PROPERTIES OUTPUT_NAME "aziotsharedutil_dll")
+    set_target_properties(aziotsharedutil_dll PROPERTIES
+        OUTPUT_NAME "aziotsharedutil_dll"
+        VERSION ${C_SHARED_VERSION}
+        SOVERSION ${C_SHARED_VERSION_MAJOR}
+    )
 endif()
 
 set(aziotsharedutil_target_libs)
```

### Verification

Linux x86_64, gcc 12.2.0, GNU ld 2.40, cmake 3.25.1, OpenSSL 3.0.20; before and after, both configured with `-Duse_openssl=ON -Drun_unittests=ON -Dskip_samples=ON -Dbuild_as_dynamic=ON -DCMAKE_BUILD_TYPE=Debug`.

| | before | after |
|---|---|---|
| build output | one regular file `libaziotsharedutil_dll.so` | `.so` → `.so.1` → `.so.1.1.12` |
| SONAME | `libaziotsharedutil_dll.so` | `libaziotsharedutil_dll.so.1` |
| `cmake --install` | one file | versioned file + both symlinks, no `install()` change needed |
| exported targets | `IMPORTED_LOCATION .../libaziotsharedutil_dll.so`, `IMPORTED_SONAME libaziotsharedutil_dll.so` | `.../libaziotsharedutil_dll.so.1.1.12`, `libaziotsharedutil_dll.so.1` |
| `ctest` | 49/49 passed | 49/49 passed |

Also confirmed: the default (static) configuration is unaffected and still configures cleanly, and a sample consumer links and runs against the versioned install with `DT_NEEDED libaziotsharedutil_dll.so.1`.

### For whoever merges

- One-time break: binaries already linked against the unversioned `.so` keep resolving while the unversioned development symlink is on the search path, but fail once a package ships only `libaziotsharedutil_dll.so.1*` at runtime. Consumers of shared builds must be relinked.
- Packaging recipes that glob `lib/libaziotsharedutil_dll.so` now see three paths and should split `.so` into `-dev` and `.so.1*` into the runtime package.
- No effect on Windows DLL/import-library names (`VERSION` only stamps the image version there). macOS dylib naming and `install_name` do change; I could not test that platform.
- Separately: `version.txt` has not been updated since 2018, so `.so.1.1.12` will be a stable path across very different snapshots. Independent of this change, but worth deciding whether to maintain it per release or document it as deliberately frozen.

ewertons:
Updated the branch with the change recommended above (previous head was `35f12b37`, preserved in this thread for reference).

- `Enable so versioning` — rebased onto current master `5b14916`, unchanged content, original authorship kept.
- `Harden version.txt parsing for so versioning` — replaces the anchored `string(REGEX MATCH ... _ ...)` with `string(STRIP)` plus an explicit `if(... MATCHES ...)` guard and a `FATAL_ERROR`.

Verified locally on Linux x86_64 (gcc 12.2.0, GNU ld 2.40, cmake 3.25.1, OpenSSL 3.0.20), configured with `-Duse_openssl=ON -Drun_unittests=ON -Dskip_samples=ON -Dbuild_as_dynamic=ON -DCMAKE_BUILD_TYPE=Debug`:

- build output `libaziotsharedutil_dll.so` → `.so.1` → `.so.1.1.12`; `readelf -d` → `SONAME libaziotsharedutil_dll.so.1`
- `cmake --install` produces the versioned file plus both symlinks, with no change to the `install()` rules
- exported targets file: `IMPORTED_LOCATION .../libaziotsharedutil_dll.so.1.1.12`, `IMPORTED_SONAME libaziotsharedutil_dll.so.1`
- `ctest` 49/49 passed
- default (static) configuration unaffected, configures cleanly
- `version.txt` with a trailing newline now configures successfully; a malformed value fails with `Malformed version in version.txt: 'not-a-version'` instead of the confusing `set_target_properties` arity error

The downstream notes in my previous comment still apply: consumers linked against the unversioned `libaziotsharedutil_dll.so` need relinking, and packaging recipes should split `.so` into `-dev` and `.so.1*` into the runtime package.

ewertons:
CI validation of `cc32c0ff`: **17/17 jobs pass**, including Windows x64, Windows Dynamic, XCode Native, OSX, ARM64 cross-compile, WolfSSL, mbedTLS, mbedTLS 4.x, BearSSL and Installed Deps.

The pipeline does not trigger on pull requests from forks, so the checks were produced by building this exact commit on an in-repo branch; they attach to the commit and so appear here. `/azp run` does not help — the bot replies "No pipelines are associated with this pull request".

Local verification (Linux x86-64, gcc 12.2.0, cmake 3.25.1, OpenSSL 3.0.20, `-Dbuild_as_dynamic=ON -Drun_unittests=ON`): ctest 49/49, symlink chain `.so` -> `.so.1` -> `.so.1.1.12`, SONAME `libaziotsharedutil_dll.so.1`.

maximsemenov80 (approved):

ewertons:
CI validation of `3b450db`: **17/17 jobs pass**. Ran via a temporary branch in this repo because fork PRs are not built automatically; the results are attached to this PR's head commit.

Issue #636, opened 2023-07-31
Segmentation fault caused by initiate_socket_connection in adapters/socketio_berkeley.c when DNS resolution failed


advantech-edgelink:
```c
static int initiate_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
{
    int result;
    int flags;
    struct addrinfo* addr = NULL;
    struct sockaddr_un addrInfoUn;

    if(socket_io_instance->address_type == ADDRESS_TYPE_IP)
    {
        ...
    }
    else
    {
        ...
    }

    if (socket_io_instance->address_type == ADDRESS_TYPE_IP)
    {
        socket_io_instance->socket = socket(addr->ai_family, addr->ai_socktype, addr->ai_protocol);
    }
    else
    {
        socket_io_instance->socket = socket(AF_UNIX, SOCK_STREAM, 0);
    }
    ...
}
```

The `addr` may be NULL when executing `socket_io_instance->socket = socket(addr->ai_family, addr->ai_socktype, addr->ai_protocol);`

ewertons:
Thanks for reporting, @advantech-edgelink ! A PR with the fix will come briefly.

RLeclair:
Hi @advantech-edgelink, we've merged the PR with the fix, thank you again for reporting! Closing the issue now.

PR #637, opened 2023-07-31
Fix for Segfault in initiate_socket_connection (socketio_berkeley) when DNS resolution fails (gh#636)

ewertons:
Addresses https://github.com/Azure/azure-c-shared-utility/issues/636

Note: there are no active unit tests for socketio_berkeley.

### DIFF
--- a/adapters/socketio_berkeley.c
+++ b/adapters/socketio_berkeley.c
@@ -536,31 +536,31 @@ static int initiate_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
         }
     }
 
-    if (socket_io_instance->address_type == ADDRESS_TYPE_IP)
-    {
-        socket_io_instance->socket = socket(addr->ai_family, addr->ai_socktype, addr->ai_protocol);
-    }
-    else
-    {
-        socket_io_instance->socket = socket(AF_UNIX, SOCK_STREAM, 0);
-    }
-
-    if (socket_io_instance->socket < SOCKET_SUCCESS)
-    {
-        LogError("Failure: socket create failure %d.", socket_io_instance->socket);
-        result = MU_FAILURE;
-    }
-#ifndef __APPLE__
-    else if (socket_io_instance->target_mac_address != NULL &&
-                set_target_network_interface(socket_io_instance->socket, socket_io_instance->target_mac_address) != 0)
-    {
-        LogError("Failure: failed selecting target network interface (MACADDR=%s).", socket_io_instance->target_mac_address);
-        result = MU_FAILURE;
-    }
-#endif //__APPLE__
-
     if(result == 0)
     {
+        if (socket_io_instance->address_type == ADDRESS_TYPE_IP)
+        {
+            socket_io_instance->socket = socket(addr->ai_family, addr->ai_socktype, addr->ai_protocol);
+        }
+        else
+        {
+            socket_io_instance->socket = socket(AF_UNIX, SOCK_STREAM, 0);
+        }
+
+        if (socket_io_instance->socket < SOCKET_SUCCESS)
+        {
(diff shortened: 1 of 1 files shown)


adapters/socketio_berkeley.c
```diff
@@ -536,31 +536,31 @@ static int initiate_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
         }
     }
 
-    if (socket_io_instance->address_type == ADDRESS_TYPE_IP)
-    {
-        socket_io_instance->socket = socket(addr->ai_family, addr->ai_socktype, addr->ai_protocol);
-    }
-    else
+    if(result == 0)
```
CIPop, line 539: It would be great if we could add a UT as well.

adapters/socketio_berkeley.c
```diff
@@ -536,31 +536,31 @@ static int initiate_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
...
+        #ifndef __APPLE__
+        else if (socket_io_instance->target_mac_address != NULL &&
+                    set_target_network_interface(socket_io_instance->socket, socket_io_instance->target_mac_address) != 0)
+        {
+            LogError("Failure: failed selecting target network interface (MACADDR=%s).", socket_io_instance->target_mac_address);
+            result = MU_FAILURE;
+        }
+        #endif //__APPLE__
 
-    if(result == 0)
```
CIPop, line 562: `result` is modified above (e.g. line 553, 560, etc) - I think this check is still needed, in addition to the one above.

CIPop (changes requested):
I think we're still missing a check.
Ideally, we would need a UT to confirm we're able to react to errors at all levels.

adapters/socketio_berkeley.c
ewertons: Correct. It's missing an `else` on line 563. I just pushed the fix.

adapters/socketio_berkeley.c
ewertons: The UT for this module are disabled. Filed an issue to track that.
https://github.com/Azure/azure-c-shared-utility/issues/638

adapters/socketio_berkeley.c
```diff
@@ -536,32 +536,31 @@ static int initiate_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
...
+                    set_target_network_interface(socket_io_instance->socket, socket_io_instance->target_mac_address) != 0)
+        {
+            LogError("Failure: failed selecting target network interface (MACADDR=%s).", socket_io_instance->target_mac_address);
+            result = MU_FAILURE;
+        }
+        #endif //__APPLE__
+        else if ((-1 == (flags = fcntl(socket_io_instance->socket, F_GETFL, 0))) ||
             (fcntl(socket_io_instance->socket, F_SETFL, flags | O_NONBLOCK) == -1))
         {
             LogError("Failure: fcntl failure.");
```
CIPop, line 566: Code in line 580 is a bit unclear as well: it would help if we add a comment:


    if ((result != 0) && (errno != EINPROGRESS))
            {
                LogError("Failure: connect failure %d.", errno);
                result = MU_FAILURE;
            }
            else  // result == 0 || errno == EINPROGRESS
            {
                // Async connect will return -1.
                result = 0;                   // <----------- Note that result could already be 0 here.
                if (socket_io_instance->on_io_open_complete != NULL)
                {
                    socket_io_instance->on_io_open_complete(socket_io_instance->on_io_open_complete_context, IO_OPEN_OK /*: IO_OPEN_ERROR*/);
                }
            }

CIPop (approved):
Thanks @ewertons !

@RLeclair will manually verify we have not missed anything by simulating errors in a debugger (since adding UTs would be a larger project).

CIPop (approved):

PR #644, opened 2023-10-09
ssl: formalise LibreSSL detection checks

chrisnovakovic:
The `tlsio_openssl` adapter checks whether libssl is provided by LibreSSL before performing certain operations by checking whether `OPENSSL_VERSION_NUMBER` is or exceeds `0x20000000`. This check causes compile-time failures under OpenSSL 3, which identifies itself with an `OPENSSL_VERSION_NUMBER` of `0x30000000` or above and has an API that deviates from LibreSSL's.

When checking for LibreSSL, instead explicitly check whether `LIBRESSL_VERSION_NUMBER` is defined.


CIPop (approved):

ericwol-msft:
#651

PR #646, opened 2023-10-22
Add support for decoding the + char as space

fmuntean:


CIPop (changes requested):
URL Encoding and Decoding is specified by https://www.rfc-editor.org/rfc/rfc3986

The proper way to encode a space is using percent encoding. For an example please see section 2.1:

> A percent-encoding mechanism is used to represent a data octet in a
   component when that octet's corresponding character is outside the
   allowed set or is being used as a delimiter of, or within, the
   component.  A percent-encoded octet is encoded as a character
   triplet, consisting of the percent character "%" followed by the two
   hexadecimal digits representing that octet's numeric value.  For
   example, "%20" is the percent-encoding for the binary octet
   "00100000" (ABNF: %x20), which in US-ASCII corresponds to the space
   character (SP).  [Section 2.4](https://www.rfc-editor.org/rfc/rfc3986#section-2.4) describes when percent-encoding and
   decoding is applied.

You can also test this in any compliant browser:
`https://test/hello world` will be encoded prior to the HTTP request into `https://test/hello%20world`

If the space is encoded as + (or otherwise)  normalization (needed for comparison) would be wrong.

PR #663, opened 2024-05-15
Update first_pending_io, schannel fix

RLeclair:


ewertons (approved):
Too bad we do not have unit tests at all for tlsio_schannel.c.

ewertons (changes requested):
There are other places in this file that may be affected by similar issue. 
Line 744 has a similar issue.

ewertons (approved):

PR #669, opened 2026-04-24
Add csr_gen platform abstraction for CSR generation

maximsemenov80:
Introduce csr_gen.h with csr_gen_ec_p256() and two adapter implementations:
  - adapters/csr_gen_openssl.c  (Linux, built with use_openssl)
  - adapters/csr_gen_schannel.c (Windows, built with use_schannel, using BCrypt + CryptEncodeObjectEx)

Both produce identical output formats:
  * csr_base64      = Base64(DER(PKCS#10 CertificationRequest))
  * private_key_pem = PKCS#8 PEM

This lets test/e2e code generate ECDSA P-256 CSRs without calling native SSL APIs directly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -288,6 +288,7 @@ if(${use_schannel})
     set(source_c_files ${source_c_files}
         ./adapters/tlsio_schannel.c
         ./adapters/x509_schannel.c
+        ./adapters/csr_gen_schannel.c
     )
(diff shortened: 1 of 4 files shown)


ewertons (approved):
This is ok to go in. Later we will need to refactor a little bit to become more separate from the concept of a specific feature in C SDK.

PR #671, opened 2026-04-27
Update macro-utils lib

ewertons:


ewertons:
Closing unmerged: superseded and now a no-op.

- `41c5136` (merged 2026-04-27, ~2h after this PR opened) removed `deps/azure-macro-utils-c` entirely and added `deps/macro-utils-c` pinned at `a46f219` — the exact SHA this PR bumps to. Rebasing this PR onto `master` is a modify/delete conflict whose only correct resolution is accepting the deletion, which yields an empty commit.
- The bump was also a no-op on its own base. `a46f219` renames the CMake target to `macro_utils_c` and no longer exports `MACRO_UTILS_INC_FOLDER`, so the `if (NOT TARGET azure_macro_utils_c)` guard in `dependencies.cmake` never tripped and the build kept using the old headers from `deps/umock-c/deps/azure-macro-utils-c`. Building base vs. this PR produced identical objects apart from one unreferenced `libmacro_utils_c.a`; identical warnings (none) and 49/49 unit tests passing both ways.
- It also bumped only 1 of the 6 recursive macro-utils pins, leaving two incompatible header layouts in one tree (the Check Submodules failure).

`master` now pins `macro-utils-c` at `045636a`, which is ahead of `a46f219` and includes the `-Wsign-compare` and C99 `MU_STATIC_ASSERT` fixes, so re-targeting this PR would be a downgrade.

PR #679, opened 2026-07-07
Add Negotiate/GSSAPI proxy authentication for HTTPS and MQTT-over-WSS

mlilien:
## Summary

Adds support for authenticating to HTTP proxies that require **Negotiate (Kerberos/SPNEGO)** rather than Basic auth. This covers both proxy code paths in the library:

- **HTTPS / DPS path (`adapters/httpapi_curl.c`)** — sets `CURLOPT_PROXYAUTH` to `CURLAUTH_NEGOTIATE | CURLAUTH_BASIC`, so libcurl satisfies whichever scheme the proxy advertises in its 407 response, with Basic kept as a fallback. `CURLE_NOT_BUILT_IN` is ignored so builds of curl without GSSAPI still work.
- **MQTT-over-WSS path (`src/http_proxy_io.c`)** — the hand-rolled HTTP `CONNECT` tunnel now performs a multi-round-trip SPNEGO handshake. On a `407` carrying `Proxy-Authenticate: Negotiate`, it runs `gss_init_sec_context`, base64-encodes the output token, and re-issues `CONNECT` with `Proxy-Authorization: Negotiate <token>`, draining the 407 body before reusing the connection.

All new behavior is **gated behind a new CMake option `use_gssapi` (default OFF)**, which links `gssapi_krb5` and defines `AZURE_C_SHARED_UTILITY_USE_GSSAPI`. With the option off, the build is byte-for-byte unchanged.

## Details

- SPNEGO mechanism (OID `1.3.6.1.5.5.2`, per RFC 4559) is requested explicitly rather than raw krb5, using static OID descriptors portable across MIT and Heimdal.
- `Proxy-Authenticate` is parsed as an RFC 7235 comma-separated challenge list, honoring quoted strings, so `Negotiate` is recognized at any position and across multiple header lines (e.g. `Negotiate, Basic realm=...` and `Basic realm=..., Negotiate`).
- The negotiation aborts cleanly (→ `IO_OPEN_ERROR`) when the 407 signals `Connection: close` / `Proxy-Connection: close`, when the body length is ambiguous (`Transfer-Encoding`, missing/malformed/overflowing `Content-Length`), or when GSSAPI fails / yields an empty token.
- `http_proxy_io_open` resets all per-session negotiate state (GSS context, negotiate-complete flag, body-drain counter) and frees any stale receive buffer, so reopened instances don't hang or refuse to re-authenticate.
- `GSS_C_MUTUAL_FLAG` is not requested, since the server's final token is never processed and mutual auth could not actually be verified.

## Tests & docs

- Behavior specified as `SRS_HTTP_PROXY_IO_01_096..105` in `devdoc/http_proxy_io_requirements.md`.
- 18 new unit tests in `tests/http_proxy_io_ut` covering the negotiate handshake, body drain, challenge-list parsing, abort paths, and state reset on reopen (GSSAPI mocked).

## Compatibility / risk

- Default build (`use_gssapi=OFF`) is unaffected.
- The curl `CURLOPT_PROXYAUTH` change is active whenever a proxy is configured, but only broadens the accepted auth schemes; Basic remains the fallback.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,6 +84,7 @@ option(use_builtin_httpapi "set use_builtin_httpapi to ON to use the built-in ht
 option(use_cppunittest "set use_cppunittest to ON to build CppUnitTest tests on Windows (default is OFF)" OFF)
 option(suppress_header_searches "do not try to find headers - used when compiler check will fail" OFF)
 option(use_custom_heap "use externally defined heap functions instead of the malloc family" OFF)
+option(use_gssapi "set use_gssapi to ON to enable Kerberos/SPNEGO proxy authentication in http_proxy_io (default is OFF)" OFF)
 option(no_openssl_engine "Disables the use of ENGINEs in OpenSSL" OFF)
 option(enable_ipv6 "set enable_ipv6 to ON to enable dual-ip-stack (IPv4 and IPv6) support (default is OFF for IPv4 only)" OFF)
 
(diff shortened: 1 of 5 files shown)


mlilien:
@microsoft-github-policy-service agree company="Conplement AG"

PR #683, opened 2026-08-26
fix: make socketio retries recoverable

### LINKED ISSUE #682: socketio_berkeley and socketio_win32 cannot recover from transient DNS or connection failures
## Describe the bug

Transient DNS or connection failures can leave the Berkeley and Win32 socket adapters unable to retry. This was first observed through the Azure Event Hubs C++ SDK on Linux with the vcpkg `azure-c-shared-utility` port:

```text
File: adapters/socketio_berkeley.c:534 Func: initiate_socket_connection: DNS resolution failed. Hostname:cpp-connscale-ns.servicebus.windows.net
File: adapters/socketio_berkeley.c:863 Func: socketio_open: Failure: socket state is not closed.
```

The two Berkeley messages are causally linked:

1. The default synchronous resolver calls `getaddrinfo()` and marks the lookup complete even when it fails.
2. `lookup_address()` changes `io_state` from `IO_STATE_CLOSED` to `IO_STATE_OPEN` whenever the lookup is complete, without checking whether it succeeded.
3. `initiate_socket_connection()` then gets a null address, reports the DNS failure, and returns an error.
4. The failed open path does not restore `io_state` to `IO_STATE_CLOSED`.
5. A retry on the same XIO stack reaches `socketio_open()` with `io_state == IO_STATE_OPEN` and fails with `socket state is not closed` before attempting DNS again.

Relevant Berkeley code:

- [`lookup_address()` sets `IO_STATE_OPEN`](https://github.com/Azure/azure-c-shared-utility/blob/772a4f8bc338140b4a0f404cf9c344283c5c937f/adapters/socketio_berkeley.c#L268-L281)
- [`initiate_socket_connection()` detects the failed result](https://github.com/Azure/azure-c-shared-utility/blob/772a4f8bc338140b4a0f404cf9c344283c5c937f/adapters/socketio_berkeley.c#L514-L540)
- [`socketio_open()` rejects the retry](https://github.com/Azure/azure-c-shared-utility/blob/772a4f8bc338140b4a0f404cf9c344283c5c937f/adapters/socketio_berkeley.c#L849-L910)

There is also a cross-platform resolver recovery problem. The synchronous resolver caches a failed lookup as complete and failed for the lifetime of the socket I/O object. `socketio_close()` changes the socket state but does not reset or recreate the resolver. Therefore, explicitly closing before retrying removes the state warning but the same object still returns the cached DNS failure without another `getaddrinfo()` call.

- [`dns_resolver_sync` caches the failed lookup](https://github.com/Azure/azure-c-shared-utility/blob/772a4f8bc338140b4a0f404cf9c344283c5c937f/src/dns_resolver_sync.c#L131-L177)
- [`dns_resolver_get_addrInfo()` keeps returning null](https://github.com/Azure/azure-c-shared-utility/blob/772a4f8bc338140b4a0f404cf9c344283c5c937f/src/dns_resolver_sync.c#L245-L278)

### Windows impact

Windows uses `socketio_win32.c` and always selects `dns_resolver_sync.c`. A transient DNS failure therefore remains cached on Windows too. The immediate state behavior differs from Berkeley: `lookup_address()` checks `dns_resolver_get_ipv4()` before setting `IO_STATE_OPEN`, so an initial DNS failure leaves the state closed and does not by itself produce the second `socket state is not closed` message. A retry on the same handle still cannot recover because it receives the cached DNS failure instead of performing a fresh lookup.

Win32 also has a related failed-open state bug. After DNS succeeds, `lookup_address()` sets `io_state` to `IO_STATE_OPEN`. If `connect()` or `ioctlsocket()` then fails, `socketio_open()` closes and invalidates the socket descriptor but does not restore `io_state` to `IO_STATE_CLOSED`. A retry on the same handle consequently fails with `socket state is not closed`.

- [Windows selects `socketio_win32.c` and `dns_resolver_sync.c`](https://github.com/Azure/azure-c-shared-utility/blob/772a4f8bc338140b4a0f404cf9c344283c5c937f/configs/azure_c_shared_utilityFunctions.cmake#L710-L720)
- [`lookup_address()` handles DNS failure before setting `IO_STATE_OPEN`](https://github.com/Azure/azure-c-shared-utility/blob/772a4f8bc338140b4a0f404cf9c344283c5c937f/adapters/socketio_win32.c#L156-L183)
- [`connect_socket()` reports `connect()` and `ioctlsocket()` failures](https://github.com/Azure/azure-c-shared-utility/blob/772a4f8bc338140b4a0f404cf9c344283c5c937f/adapters/socketio_win32.c#L186-L206)
- [`socketio_open()` cleans up the descriptor without restoring the closed state](https://github.com/Azure/azure-c-shared-utility/blob/772a4f8bc338140b4a0f404cf9c344283c5c937f/adapters/socketio_win32.c#L466-L485)

The vcpkg source directory suffix `283c5c937f` corresponds to commit `772a4f8bc338140b4a0f404cf9c344283c5c937f`. The vcpkg port uses the default `use_c_ares=OFF` configuration, so the observed Linux report and the Windows behavior described above use the synchronous resolver path.

## Expected behavior

A transient DNS or connection failure should not permanently poison a socket I/O object.

After a failed open, a subsequent connection attempt should:

- start from a valid closed/not-open state;
- perform a fresh DNS lookup; and
- be able to connect if DNS or connectivity has recovered.

This matters for layered XIO users. During a failed TLS open, the TLS adapter returns to its not-open state. When an upper layer retries the same transport, it calls the underlying socket open again and encounters the stale socket state or cached resolver result.

### Berkeley and Win32 DNS recovery

Use a mocked resolver where the first lookup fails with a transient result such as `EAI_AGAIN` and the second lookup succeeds:

1. Create one socket I/O handle.
2. Call `socketio_open()` and fail the first DNS lookup.
3. Retry according to the supported open/close contract.
4. Verify that the second attempt performs another lookup and can proceed to `socket()` and `connect()`.

### Win32 connection recovery

1. Create one Win32 socket I/O handle and complete DNS successfully.
2. Make `connect()` or `ioctlsocket()` fail during the first `socketio_open()` call.
3. Retry according to the supported open/close contract.
4. Verify that the adapter starts from `IO_STATE_CLOSED`, creates a new socket, and attempts the connection again.

The Berkeley socket unit tests are currently disabled; see #638. This likely explains why the failed-open retry behavior is not covered.

A complete fix likely needs both parts:

- restore a retryable state after every failed open path in both socket adapters; and
- reset or recreate the DNS resolver for a new connection attempt.

For diagnostics, the synchronous resolver should also log `gai_strerror(getAddrResult)` and, for `EAI_SYSTEM`, `errno`, rather than only a numeric result.

## History

PR #437 addressed the same general DNS/socket-state area by propagating the return value from `initiate_socket_connection()`. It did not restore the socket state or reset the cached resolver after failure.

j7nw4r:
## Summary

Transient Berkeley and Win32 socketio failures leave the adapter with a stale socket or a resolver that caches the failed lookup, so a later open on the same handle cannot recover.

## Motivation

The synchronous resolver marks a failed lookup as complete and never retries it. The failed open paths do not restore `IO_STATE_CLOSED`. A retry on the same XIO stack then fails before it tries DNS again.

## Changes

- Destroys the resolver in `socketio_cleanup` and creates it again on the next open, in both adapters.
- Restores `IO_STATE_CLOSED` on every failed open path, with one cleanup layer per adapter.
- Keeps `socketio_close` a no-op in `IO_STATE_CLOSED`, so an unopened accepted socket keeps its descriptor.
- Reports `IO_OPEN_OK` from the Berkeley dowork path only after the connect wait succeeds.
- Fails the Berkeley `lookup_address` on a completed lookup with no address, and rejects descriptors at or above `FD_SETSIZE`.
- Logs `gai_strerror` text, and `errno` for `EAI_SYSTEM`, in the synchronous resolver.
- Re-enables `socketio_berkeley_ut` and adds retry coverage for both adapters and for the resolver diagnostics.

## Test plan

- `build_all/linux/build.sh --run-unittests` on macOS: 45 of 45 suites pass, including 10 of 10 Berkeley retry tests.
- `dns_resolver_ut` in a `LINUX=1` tree on macOS: 16 of 16 tests pass.
- `git diff --check` passes.
- The Linux-only interface test and the `no_logging` build were checked by trace only; the Linux CI jobs run them.
- The Win32 suite was traced by hand against the adapter and needs the Windows CI job.

Fixes #682

### DIFF
--- a/adapters/socketio_berkeley.c
+++ b/adapters/socketio_berkeley.c
@@ -20,6 +20,7 @@
 #endif
 
 #include <signal.h>
+#include <stdbool.h>
 #include <stdlib.h>
 #include <stddef.h>
 #include <stdio.h>
@@ -29,6 +30,7 @@
 #include <sys/types.h>
 #include <sys/socket.h>
 #include <sys/select.h>
+#include <sys/time.h>
 #ifdef TIZENRT
 #include <net/lwip/tcp.h>
 #else
@@ -154,8 +156,6 @@ static void* socketio_CloneOption(const char* name, const void* value)
     }
     return result;
 }
-
-
 
 /*this function destroys an option previously created*/
 static void socketio_DestroyOption(const char* name, const void* value)
@@ -703,6 +703,38 @@ static int lookup_address_and_initiate_socket_connection(SOCKET_IO_INSTANCE* soc
     return result;
 }
 
+// Computes the time left until deadline. Returns false, leaving remaining
+// zeroed, once the deadline has passed.
+static bool get_time_remaining(const struct timeval* deadline, struct timeval* remaining)
+{
+    bool result;
+    struct timeval now;
+
+    (void)gettimeofday(&now, NULL);
+
+    remaining->tv_sec = deadline->tv_sec - now.tv_sec;
+    remaining->tv_usec = deadline->tv_usec - now.tv_usec;
+
+    if (remaining->tv_usec < 0)
+    {
+        remaining->tv_usec += 1000000;
+        remaining->tv_sec--;
+    }
+
+    if (remaining->tv_sec < 0 || (remaining->tv_sec == 0 && remaining->tv_usec == 0))
+    {
+        remaining->tv_sec = 0;
+        remaining->tv_usec = 0;
+        result = false;
+    }
+    else
+    {
+        result = true;
+    }
+
+    return result;
+}
+
 static int wait_for_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
 {
     int result;
@@ -713,6 +745,8 @@ static int wait_for_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
     fd_set fdset;
     struct timeval tv;
 
+    // FD_SET indexes the descriptor set by descriptor value, so a descriptor at
+    // or above FD_SETSIZE writes past the end of the set.
     if (socket_io_instance->socket >= FD_SETSIZE)
     {
         LogError("Failure: socket %d is not below FD_SETSIZE.", socket_io_instance->socket);
@@ -720,13 +754,26 @@ static int wait_for_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
     }
     else
     {
-        FD_ZERO(&fdset);
-        FD_SET(socket_io_instance->socket, &fdset);
-        tv.tv_sec = CONNECT_TIMEOUT;
-        tv.tv_usec = 0;
+        struct timeval deadline;
 
+        (void)gettimeofday(&deadline, NULL);
+        deadline.tv_sec += CONNECT_TIMEOUT;
+
+        // select() may report EINTR before the connection is decided. Retrying
+        // against a fixed deadline bounds the total wait: platforms that leave
+        // the timeout untouched would otherwise restart it on every signal.
         do
         {
+            if (!get_time_remaining(&deadline, &tv))
+            {
+                // Deadline reached; report it the same way select() reports a timeout.
+                retval = 0;
+                break;
+            }
+
+            FD_ZERO(&fdset);
+            FD_SET(socket_io_instance->socket, &fdset);
+
             retval = select(socket_io_instance->socket + 1, NULL, &fdset, NULL, &tv);
 
             if (retval < 0)
--- a/adapters/socketio_berkeley.c
+++ b/adapters/socketio_berkeley.c
@@ -20,7 +20,6 @@
 #endif
 
 #include <signal.h>
-#include <stdbool.h>
 #include <stdlib.h>
 #include <stddef.h>
 #include <stdio.h>
@@ -101,7 +100,6 @@ typedef struct SOCKET_IO_INSTANCE_TAG
     SINGLYLINKEDLIST_HANDLE pending_io_list;
     unsigned char recv_bytes[XIO_RECEIVE_BUFFER_SIZE];
     DNSRESOLVER_HANDLE dns_resolver;
-    bool dns_resolver_stale;
 } SOCKET_IO_INSTANCE;
 
 typedef struct NETWORK_INTERFACE_DESCRIPTION_TAG
@@ -232,9 +230,11 @@ static void socketio_cleanup(SOCKET_IO_INSTANCE* socket_io_instance)
     socket_io_instance->socket = INVALID_SOCKET;
     socket_io_instance->io_state = IO_STATE_CLOSED;
 
-    if (socket_io_instance->address_type == ADDRESS_TYPE_IP && socket_io_instance->hostname != NULL)
+    if (socket_io_instance->address_type == ADDRESS_TYPE_IP && socket_io_instance->hostname != NULL &&
+        socket_io_instance->dns_resolver != NULL)
     {
-        socket_io_instance->dns_resolver_stale = true;
+        dns_resolver_destroy(socket_io_instance->dns_resolver);
+        socket_io_instance->dns_resolver = NULL;
     }
 }
 
@@ -243,23 +243,13 @@ static int refresh_dns_resolver(SOCKET_IO_INSTANCE* socket_io_instance)
     int result = 0;
 
     if (socket_io_instance->address_type == ADDRESS_TYPE_IP && socket_io_instance->hostname != NULL &&
-        (socket_io_instance->dns_resolver == NULL || socket_io_instance->dns_resolver_stale))
+        socket_io_instance->dns_resolver == NULL)
     {
-        if (socket_io_instance->dns_resolver != NULL)
-        {
-            dns_resolver_destroy(socket_io_instance->dns_resolver);
-            socket_io_instance->dns_resolver = NULL;
-        }
-
         socket_io_instance->dns_resolver = dns_resolver_create(socket_io_instance->hostname, socket_io_instance->port, NULL);
         if (socket_io_instance->dns_resolver == NULL)
         {
             LogError("Failure: unable to create DNS resolver.");
             result = MU_FAILURE;
-        }
-        else
-        {
-            socket_io_instance->dns_resolver_stale = false;
         }
     }
 
@@ -976,7 +966,7 @@ int socketio_close(CONCRETE_IO_HANDLE socket_io, ON_IO_CLOSE_COMPLETE on_io_clos
     else
     {
         SOCKET_IO_INSTANCE* socket_io_instance = (SOCKET_IO_INSTANCE*)socket_io;
-        if (socket_io_instance->io_state != IO_STATE_CLOSING)
+        if ((socket_io_instance->io_state != IO_STATE_CLOSED) && (socket_io_instance->io_state != IO_STATE_CLOSING))
         {
             // Only shut down an active socket before cleanup.
             if (socket_io_instance->socket != INVALID_SOCKET)
--- a/adapters/socketio_berkeley.c
+++ b/adapters/socketio_berkeley.c
@@ -238,6 +238,15 @@ static void socketio_cleanup(SOCKET_IO_INSTANCE* socket_io_instance)
     }
 }
 
+static void indicate_open_failure(SOCKET_IO_INSTANCE* socket_io_instance)
+{
+    socketio_cleanup(socket_io_instance);
+    if (socket_io_instance->on_io_error != NULL)
+    {
+        socket_io_instance->on_io_error(socket_io_instance->on_io_error_context);
+    }
+}
+
 static int refresh_dns_resolver(SOCKET_IO_INSTANCE* socket_io_instance)
 {
     int result = 0;
@@ -317,6 +326,11 @@ static int lookup_address(SOCKET_IO_INSTANCE* socket_io_instance)
         else if (!dns_resolver_is_lookup_complete(socket_io_instance->dns_resolver))
         {
             socket_io_instance->io_state = IO_STATE_OPENING;
+        }
+        else if (dns_resolver_get_addrInfo(socket_io_instance->dns_resolver) == NULL)
+        {
+            LogError("DNS resolution failed. Hostname:%s", socket_io_instance->hostname);
+            result = MU_FAILURE;
         }
         else
         {
@@ -659,18 +673,8 @@ static int initiate_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
             {
                 // Async connect will return -1.
                 result = 0;
-                if (socket_io_instance->on_io_open_complete != NULL)
-                {
-                    socket_io_instance->on_io_open_complete(socket_io_instance->on_io_open_complete_context, IO_OPEN_OK /*: IO_OPEN_ERROR*/);
-                }
             }
         }
-
-    }
-
-    if (result != 0)
-    {
-        socketio_cleanup(socket_io_instance);
     }
 
     return result;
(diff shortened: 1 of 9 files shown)


ewertons:
Verified on Linux (glibc, x86-64, gcc 12), default `use_c_ares=OFF`.

**Issue #682 reproduces on master.** Linking the built library into a small driver, a second `xio_open` on the same handle gives exactly the reported pair:
```
socketio_berkeley.c:534 initiate_socket_connection: DNS resolution failed. Hostname:...
socketio_berkeley.c:863 socketio_open: Failure: socket state is not closed.
```
**This branch fixes both parts.** The second open performs a fresh `getaddrinfo`, and a connect that failed with ECONNREFUSED succeeds on retry once a listener is up (master returns MU_FAILURE both times).

Two things block the Linux CI job:

**1. `tests/socketio_berkeley_ut/socketio_berkeley_ut.c` does not compile on Linux.**
```
socketio_berkeley_ut.c:309: error: invalid use of undefined type 'struct ifconf'
socketio_berkeley_ut.c:310: error: invalid application of 'sizeof' to incomplete type 'struct ifreq'
```
The build adds `-D_POSIX_C_SOURCE=200112L -D_XOPEN_SOURCE=500`, which hides `ifreq`/`ifconf` behind `__USE_MISC` in glibc's `<net/if.h>`. `adapters/socketio_berkeley.c` already works around this by defining `_DEFAULT_SOURCE`/`_BSD_SOURCE` before including `<net/if.h>`; the test does not. Adding the same before the first include compiles it.

**2. `socketio_open_interface_setup_failure_is_retryable` then fails on Linux** — second open returns MU_FAILURE with `Did not find a network interface matching MAC ADDRESS`. This is not a test bug; it exposes a real defect in `get_network_interface_descriptions`. On Linux `ifr_hwaddr` and `ifr_addr` are the same `ifr_ifru` union member (both offset 16), so `ioctl(SIOCGIFADDR)` overwrites the hardware address read by `ioctl(SIOCGIFHWADDR)` into the same `struct ifreq`. Measured against a real interface:
```
eth0 after SIOCGIFHWADDR: F2:2B:D7:E2:57:86
eth0 after SIOCGIFADDR  : 00:00:AC:13:00:0C   <- value create_network_interface_description formats
```
So `OPTION_NET_INT_MAC_ADDRESS` never matches on Linux. macOS `struct ifreq` does not alias that way, which is why the suite passes there. Copying the hardware address out before the SIOCGIFADDR call and using the saved copy when formatting makes the test pass.

With those two changes: **49/49 suites pass, 11/11 Berkeley tests.**

Smaller points:

3. `dns_resolver_destroy` calls `freeaddrinfo` only when `is_complete && !is_failed`. When `getaddrinfo` succeeds but returns no address of the enabled family, `is_failed` is set while `addrInfo` is non-NULL and the list leaks. That was once per socket; recreating the resolver on every open makes it once per retry. ASan over 50 retries on each of the two paths I could drive here (resolution failure, connection refused) showed no leak, since both leave `addrInfo` NULL.
4. The Berkeley `socketio_dowork` failure paths call `on_io_error` but never `on_io_open_complete(IO_OPEN_ERROR)`, so a layered caller waiting on an async open only learns through the error callback. Pre-existing, but worth settling now that the state is left retryable.
5. `dns_resolver_sync.c` now calls `gai_strerror` unconditionally on non-Windows. Worth confirming it exists on the non-glibc targets this file supports (the header comments mention lwIP).
6. The Win32 adapter and `socketio_win32_ut` were not exercised — no Windows environment available here.

ewertons:
Opened #684 with the two product bugs from my earlier comment, so this PR does not have to carry them: the `struct ifreq` union aliasing that breaks `OPTION_NET_INT_MAC_ADDRESS` on Linux, and the `addrinfo` leak in `dns_resolver_destroy`. It deliberately leaves `socketio_berkeley_ut` and `tests/CMakeLists.txt` alone, which are yours.

Once #684 is in, this branch needs exactly two things. I merged the two locally to confirm:

1. One conflict, in `dns_resolver_ut.c` `TEST_FUNCTION_INITIALIZE` — both sides add a reset line. Keep both:
```c
umock_c_reset_all_calls();
g_freeaddrinfo_call_count = 0;
#ifndef NO_LOGGING
g_log_message[0] = '\0';
#endif
```
2. The feature-test macros for `socketio_berkeley_ut.c`, so `struct ifreq`/`ifconf` are visible under the build's `-D_POSIX_C_SOURCE=200112L -D_XOPEN_SOURCE=500`:
```c
#ifndef __APPLE__
#define _DEFAULT_SOURCE
#define _BSD_SOURCE
#endif
```
before the first include, mirroring what `socketio_berkeley.c` already does.

`adapters/socketio_berkeley.c` and `src/dns_resolver_sync.c` merge cleanly.

With those applied, the merged tree builds and passes 49/49 suites on Linux, including 11/11 in the re-enabled `socketio_berkeley_ut` — `socketio_open_interface_setup_failure_is_retryable` passes once the MAC bug is fixed.

Happy to push both onto this branch after #684 merges if you prefer.

ewertons:
Pushed the two updates to this branch now that #684 is merged.

- `53c20bfb` merges master. One conflict, in `dns_resolver_ut.c` `TEST_FUNCTION_INITIALIZE`; kept both per-test resets (`g_freeaddrinfo_call_count` from master, `g_log_message` from here). `socketio_berkeley.c` and `dns_resolver_sync.c` merged cleanly.
- `45cb32a0` adds `_DEFAULT_SOURCE`/`_BSD_SOURCE` to `socketio_berkeley_ut.c` so `struct ifreq`/`ifconf` are visible under the build's `-D_POSIX_C_SOURCE=200112L -D_XOPEN_SOURCE=500`, mirroring `socketio_berkeley.c`.

Your commits are untouched.

Verified on Linux x86-64 (glibc, gcc 12):
- 49/49 suites pass, including the re-enabled `socketio_berkeley_ut` at 11/11 and `dns_resolver_ut` at 19/19.
- `socketio_open_interface_setup_failure_is_retryable` now passes; the MAC matching bug behind it was the `struct ifreq` union aliasing fixed in #684.
- Library builds clean with `-Dno_logging=ON`.
- `git diff --check` clean.

Not covered locally: the Win32 adapter and `socketio_win32_ut`, which need the Windows CI jobs.

Two open points from my earlier review, both pre-existing and neither blocking:
1. Berkeley `socketio_dowork` failure paths call `on_io_error` but never `on_io_open_complete(IO_OPEN_ERROR)`, so a layered caller awaiting an async open only learns via the error callback.
2. `dns_resolver_sync.c` calls `gai_strerror` unconditionally on non-Windows; worth confirming it exists on the non-glibc targets this file supports.

ewertons:
CI validation of this branch at `45cb32a0` is green: **17/17 jobs pass**, including Windows x64, Windows Dynamic, OSX, XCode Native, Linux with C-Ares, Linux Ubuntu with Options, ARM64 cross-compile, BearSSL, WolfSSL, mbedTLS and mbedTLS 4.x.

The pipeline does not build pull requests from forks, so the run was done on a branch carrying this exact tree. That covers the Win32 adapter and `socketio_win32_ut`, which were the remaining gap.

This PR is `MERGEABLE` and validated from my side.

adapters/socketio_berkeley.c
```diff
@@ -155,6 +155,8 @@ static void* socketio_CloneOption(const char* name, const void* value)
     return result;
 }
 
+
```
ewertons: No unnecessary extra lines.

adapters/socketio_berkeley.c
```diff
@@ -669,55 +713,54 @@ static int wait_for_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
...
-        retval = select(socket_io_instance->socket + 1, NULL, &fdset, NULL, &tv);
-
-        if (retval < 0)
-        {
-            select_errno = errno;
-        }
-    } while (retval < 0 && select_errno == EINTR);
-
-    if (retval != 1)
+    if (socket_io_instance->socket >= FD_SETSIZE)
```
ewertons, line 750: Add a 2 line comment above stating why the check

adapters/socketio_berkeley.c
```diff
@@ -669,55 +713,54 @@ static int wait_for_socket_connection(SOCKET_IO_INSTANCE* socket_io_instance)
...
-            result = MU_FAILURE;
-        }
-        else if (so_error != 0)
+            retval = select(socket_io_instance->socket + 1, NULL, &fdset, NULL, &tv);
+
+            if (retval < 0)
+            {
+                select_errno = errno;
+            }
+        } while (retval < 0 && select_errno == EINTR);
```
ewertons, line 783: Can this result in an infinite loop?

adapters/socketio_berkeley.c
ewertons: Added in 36e96572:

```c
// FD_SET indexes the descriptor set by descriptor value, so a descriptor at
// or above FD_SETSIZE writes past the end of the set.
```

adapters/socketio_berkeley.c
ewertons: Removed in 36e96572. Master has one blank line there; this had three.

adapters/socketio_berkeley.c
ewertons: Yes, on platforms where `select()` does not update the timeout. Fixed in 36e96572.

The loop is pre-existing and unchanged by this PR; it was only re-indented into the new `FD_SETSIZE` branch.

Behaviour differs by platform because the timeout is only updated in place on some:
- Linux updates it. Measured: a 3s timeout came back as 2.900043 after a 100ms interruption, and a loop under a 100Hz signal ended after 30 iterations at the original deadline. Bounded.
- macOS and the BSDs leave it untouched, so every EINTR restarts the full `CONNECT_TIMEOUT` window. A signal rate faster than the timeout keeps the loop running indefinitely. This adapter builds for those targets.

Demonstrated with a fake `select()` that interrupts every wait:
- before: still running when killed at 30s;
- after: returns at 10.004s, which is `CONNECT_TIMEOUT`.

The fix takes the deadline once before the loop and derives each timeout from it, so the bound holds regardless of whether the platform updates the timeout. The descriptor set is also re-armed each pass rather than relying on it surviving an interrupted call.

Two tests added: interrupted waits are retried and still connect, and an uninterrupted EINTR storm stops at the deadline. The second hangs without the change.

ewertons:
CI validation of `36e96572` is green: **17/17 jobs pass**, including Windows x64, Windows Dynamic, OSX, XCode Native, Linux with C-Ares, Linux Ubuntu with Options, ARM64 cross-compile, BearSSL, WolfSSL, mbedTLS and mbedTLS 4.x.

Local, Linux x86-64 (glibc, gcc 12): 49/49 suites, `socketio_berkeley_ut` 13/13, `dns_resolver_ut` 19/19.

ewertons:
Force-pushed `36e96572` -> `866d7000` to remove build artifacts I committed by mistake.

My last commit used `git add -A` from a working tree holding a build directory named `cm/`, which added 1575 generated files (1583 files / +196776 lines on the PR). `.gitignore` covers build output as `/cmake*`; `cm/` fell outside that pattern, so nothing caught it.

`866d7000` is the same commit with only `adapters/socketio_berkeley.c` and `tests/socketio_berkeley_ut/socketio_berkeley_ut.c` in it. `git diff 36e96572 866d7000 -- . ':!cm'` is empty, so no source changed. The PR is back to 8 files, +969/-598.

Your commits were not touched.

Re-verified on the cleaned tree: 49/49 suites, `socketio_berkeley_ut` 13/13. CI re-running; I will report the result.

Sorry for the noise.

ewertons:
CI validation of `866d7000` (artifacts removed): **17/17 jobs pass**, including Windows x64, Windows Dynamic, OSX, XCode Native, Linux with C-Ares, Linux Ubuntu with Options, ARM64 cross-compile, BearSSL, WolfSSL, mbedTLS and mbedTLS 4.x.

ewertons (approved):

PR #685, opened 2026-08-26
DO NOT MERGE: CI validation for #683

ewertons:
Do not merge. Validation only.

The pipeline does not build pull requests from forks ("No pipelines are associated with this pull request" on `/azp run`), so #683 cannot get a CI signal on its own. This branch carries the exact tree of #683 head `45cb32a0` so the Windows, macOS and option-matrix jobs can run against it.

Merge #683, not this. I will close this once the run finishes and report the result there.

### DIFF
--- a/adapters/socketio_berkeley.c
+++ b/adapters/socketio_berkeley.c
@@ -20,7 +20,6 @@
 #endif
 
 #include <signal.h>
-#include <stdbool.h>
 #include <stdlib.h>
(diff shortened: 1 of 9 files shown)


ewertons:
Validation complete: 17/17 jobs pass on the tree of #683 head 45cb32a0, including Windows x64 and Windows Dynamic. Closing; merge #683.

PR #689, opened 2026-08-28
ci: run the Windows tests from the directory they are built in

ewertons:
The Windows build scripts configure into `cmake\shared-util_<platform>` (`build_all/windows/build.cmd`), but the Run Tests step did `cd cmake` and ran ctest there. That directory has no `CTestTestfile.cmake`, so ctest printed `No tests were found!!!` and **exited 0** — the step passed without running anything.

Net effect: the Windows jobs have been building the test binaries and never executing them. This predates the recent pipeline work (the step has looked like this since 71d6f52).

Changes, both Windows jobs:
- `cd cmake\shared-util_x64` / `cd cmake\shared-util_Win32` so ctest runs where the tests are.
- `--no-tests=error`, so an empty test list fails the step instead of passing silently.

Verified locally (ctest 3.25.1): with no tests present, plain `ctest` exits `0`; `ctest --no-tests=error` exits `8`. The Windows run itself could not be executed locally — no Windows host available — so this needs the pipeline to confirm.

Expect this to surface pre-existing Windows test failures, since these suites have not run in CI for a long time. I will follow up on whatever it reports.

Note the x86 job is currently `condition: false`, so only the x64 job changes behaviour now; the x86 step is corrected the same way for when it is re-enabled.

### DIFF
--- a/build/.vsts-ci.yml
+++ b/build/.vsts-ci.yml
@@ -50,7 +50,7 @@ jobs:
       call jenkins\windows_c.cmd
     displayName: 'Build'
   - script: |
-      call jenkins\windows_c_VsDevCmd.cmd x86 && cd cmake && ctest -T test --no-compress-output -C Debug -V -j 16 --schedule-random
+      call jenkins\windows_c_VsDevCmd.cmd x86 && cd cmake\shared-util_Win32 && ctest -T test --no-compress-output -C Debug -V -j 16 --schedule-random --no-tests=error
     displayName: 'Run Tests'
   - task: PublishTestResults@2
     displayName: 'Publish Windows x86 Results'
@@ -81,7 +81,7 @@ jobs:
   - task: CodeQL3000Finalize@0
     condition: always()
   - script: |
-      call jenkins\windows_c_VsDevCmd.cmd x64 && cd cmake && ctest -T test --no-compress-output -C Debug -V -j 16 --schedule-random
+      call jenkins\windows_c_VsDevCmd.cmd x64 && cd cmake\shared-util_x64 && ctest -T test --no-compress-output -C Debug -V -j 16 --schedule-random --no-tests=error
     displayName: 'Run Tests'
   - task: PublishTestResults@2
     displayName: 'Publish Windows x64 Results'
--- a/tests/socketio_win32_ut/socketio_win32_ut.c
+++ b/tests/socketio_win32_ut/socketio_win32_ut.c
@@ -260,7 +260,15 @@ char* umocktypes_stringify_const_ADDRINFOA_ptr(const ADDRINFOA** value)
     char temp_buffer[256];
     int length;
 
-    length = sprintf(temp_buffer, "{ ai_flags = %d, ai_family = %d, ai_socktype = %d, ai_protocol = %d, ai_addrlen = %u, ai_canonname = %s", (*value)->ai_flags, (*value)->ai_family, (*value)->ai_socktype, (*value)->ai_protocol, (unsigned int)((*value)->ai_addrlen), (*value)->ai_canonname);
+    if (*value == NULL)
+    {
+        length = sprintf(temp_buffer, "NULL");
+    }
+    else
+    {
+        const char* canonname = ((*value)->ai_canonname == NULL) ? "NULL" : (*value)->ai_canonname;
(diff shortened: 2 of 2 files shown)


ewertons (commented):
Reviewed/Approved

ewertons:
Result of enabling the Windows tests: **52 tests ran, 51 passed, 1 failed** (previously 0 ran).

- `socketio_int` passes on Windows (3/3), as do `x509_schannel_int` and `string_utils_int`.
- One failure: `socketio_win32_ut` **crashes (SEGFAULT)**.

Isolated by running that suite on its own: the crash occurs in `socketio_send_returns_1_succeeds`. The suite reaches it after ~34 other tests, all of which pass.

This is pre-existing and independent of this change — the suite has not executed in CI since the test step started using the wrong directory, and it was mass-edited in the meantime (umock `IGNORED_ARG` migration, ctest/umock compatibility fixes, the CMake rewrite) with no way to notice a runtime break.

Reading the sources did not identify the faulting operation. Pinning down the faulting frame needs a Windows debugger; that is not available to me, so I cannot go further without either a Windows environment or several slow pipeline iterations.

Options, happy to take direction:
1. Hold this PR until `socketio_win32_ut` is fixed.
2. Land this PR with `socketio_win32_ut` explicitly excluded and a comment pointing at the crash, so the other 51 tests start running now instead of none, and track the crash separately.

The x86 job is `condition: false`, so only x64 changes behaviour.

ewertons:
## Root cause found and fixed — `socketio_win32_ut` passes, no exclusion needed

Pushed in `ec2912fc`. Reproduced and verified on a Windows host (VS 2022, x64 Debug), so no CI probing was needed.

### What was actually crashing

Not `socketio_send_returns_1_succeeds` — that test passes. The process faulted in the *next* test, `socketio_close_Succeeds`, while **registering** an expectation:

```c
EXPECTED_CALL(freeaddrinfo(IGNORED_ARG));
```

`EXPECTED_CALL` copies each argument *before* it marks it ignored, and `IGNORED_ARG` is `0`. This suite registers a custom `const ADDRINFOA*` type whose copier dereferences the pointee unconditionally, so it dereferenced null:

```
Exception Code: C0000005
    at umocktypes_copy_const_ADDRINFOA_ptr()  socketio_win32_ut.c:303
    at umocktypes_copy()                      umock-c/src/umocktypes.c:475
    at umock_c_expected_freeaddrinfo()        socketio_win32_ut.c:160
    at socketio_win32_unittests::socketio_close_Succeeds()  socketio_win32_ut.c:859
```

(Isolated with `vstest.console.exe` against the CppUnitTest DLL target, which yields a symbolized frame that the ctest executable does not.)

### Why it looked like a `socketio_send` crash

The serial run reported `socketio_send_returns_1_succeeds` as the last test to *start* logging before the process died. It had already completed successfully; the fault happened during the arrange phase of the test that followed it.

### The fix

Treat the custom type as nullable across all three callbacks:

- **copy** — NULL in, NULL out, instead of dereferencing
- **stringify** — render a NULL pointer, and a NULL `ai_canonname`, as `NULL`
- **are_equal** — compare nullable pointers and names without passing either to `strcmp`

Two lines in the existing comparator were also unreachable-by-luck: the `ai_canonname` guard fell through to `strcmp(NULL, NULL)` whenever both were null. That is corrected in the same change.

### Latent, not new

The three `freeaddrinfo(IGNORED_ARG)` expectations arrived with the resolver-teardown change, but the suite has not executed since the Windows test path broke — so the crash has been sitting in the tree unexecuted. Enabling the tests is what surfaced it, which is precisely the point of this PR.

### Verification

Exact pipeline command, run in `cmake\shared-util_x64`:

```
ctest -T test --no-compress-output -C Debug -V -j 16 --schedule-random --no-tests=error
```

```
100% tests passed, 0 tests failed out of 52
Total Test time (real) =  31.54 sec
```

**52/52 pass**, including `socketio_win32_ut` (0.87 s on its own). `socketio_int` passes on Windows as well.

### On the three options

Options 2 (exclude the suite) and 3 (hold the PR) are no longer needed. The branch now enables the Windows tests *and* they all pass, so this can go in as-is.

maximsemenov80 (approved):

